PR #10, opened 2025-09-02
[KEP-8] Introduce template reference to reduce YAML duplication

### LINKED ISSUE #8: [FEATURES] RBG YAML Deduplication & Simplification for RBG
**Problem Statement**  
When deploying large-scale inference services, we repeatedly hit the same pain points:

1. **Config explosion** – every role (prefill / decode) needs its own 200-line YAML.  
2. **Copy-paste errors** – changing an image tag or env value across dev / staging / prod is manual and error-prone.  
3. **Cross-team friction** – platform engineers want to roll out new base images, probes, or affinity rules without forcing every business team to re-apply YAML.  
4. **Leader vs Worker drift** – the same template must express different startup flags, ports, or resource requests for leader and worker pods, which today requires two nearly-identical manifests.

cheyang:
To fix #8

### DIFF
--- a/keps/8-introduce-template-reference-to-reduce-yaml-duplication/README.md
+++ b/keps/8-introduce-template-reference-to-reduce-yaml-duplication/README.md
@@ -145,7 +145,7 @@ proposal will be implemented, this is the place to discuss them.
 ### API Overview
 
 ```yaml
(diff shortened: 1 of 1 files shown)


veophi:
How to define `status.ObservedGeneration` of RBG if only its `PodTemplate` object was modified?

PR #34, opened 2025-09-22
feat: Supports using controllerrevision hash to update role

### LINKED ISSUE #21: [Bug]  How to change the unit scale of P or D at the RGB layer, given that the P/D scale of LWS becomes unmodifiable once it is determined?
### Checklist

### Describe the bug

When certain fields of LWS are not allowed to be modified, it is unreasonable that no error is reported after modifying these immutable fields under the RGB resource. In this case, how should I manage my detached resources? Is deleting and recreating them the only option? This seems to come with a relatively high cost.

### Reproduction

 Modify the p or d unit scale

### Environment

PD disaggregation

bcfre:
### Ⅰ. Motivation

1. Store the state of RBG Roles in ControllerRevisions to prepare for future rollback support.
2. Track the revision hash for each Role deployment on the RBG via labels, and compare it against the corresponding ControllerRevision to ensure Role changes on the RBG are not overlooked.

### Ⅱ. Modifications

Added two types of labels: RBG and specific role.
Added logic for creating and cleaning up ControllerRevisions.

fixes #21 

Add ut and update e2e

- example for rbg
```go
kubectl apply -f examples/basics/rbg-base.yaml 

kubectl get rbg,lws,sts,deploy,po,controllerrevision -l rolebasedgroup.workloads.x-k8s.io/name=nginx-cluster
# output
NAME                                    READY   AGE
statefulset.apps/nginx-cluster-leader   1/1     3m53s

NAME                                   READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/nginx-cluster-worker   3/3     3            3           3m53s

NAME                                       READY   STATUS    RESTARTS   AGE
pod/nginx-cluster-leader-0                 1/1     Running   0          3m54s
pod/nginx-cluster-worker-97b95d9cd-52frk   1/1     Running   0          3m52s
pod/nginx-cluster-worker-97b95d9cd-6h9rs   1/1     Running   0          3m52s
pod/nginx-cluster-worker-97b95d9cd-vjjfh   1/1     Running   0          3m52s

NAME                                                      CONTROLLER                                        REVISION   AGE
controllerrevision.apps/nginx-cluster-654fb598b9-1        rolebasedgroup.workloads.x-k8s.io/nginx-cluster   1          3m58s
controllerrevision.apps/nginx-cluster-leader-5b77c6c65f   statefulset.apps/nginx-cluster-leader             1          3m54s
```
- view the labels of a ControllerRevision
```
kubectl get  controllerrevision.apps/nginx-cluster-654fb598b9-1 --show-labels
NAME                         CONTROLLER                                        REVISION   AGE     LABELS
nginx-cluster-654fb598b9-1   rolebasedgroup.workloads.x-k8s.io/nginx-cluster   1          6m43s   rolebasedgroup.workloads.x-k8s.io/controller-revision-hash=654fb598b9,rolebasedgroup.workloads.x-k8s.io/name=nginx-cluster,rolebasedgroup.workloads.x-k8s.io/role-revision-hash-leader=bc666cd45,rolebasedgroup.workloads.x-k8s.io/role-revision-hash-worker=6c98b798bd
```
## Checklist


pkg/utils/revision_utils.go
```diff
@@ -51,3 +65,227 @@ func GetHighestRevision(revisions []*appsv1.ControllerRevision) *appsv1.Controll
...
+	}
+	return restoredRbg, nil
+}
+
+func CleanExpiredRevision(
+	ctx context.Context, client client.Client,
+	rbg *workloadsv1alpha1.RoleBasedGroup,
+	revisions []*appsv1.ControllerRevision) ([]*appsv1.ControllerRevision, error) {
+	// todo: Use the default value temporarily, and add new attribute fields in RBG later
+	exceedNum := len(revisions) - 10
```
cheyang: I suggest the default value is 5, and please use constant  `DefaultRevisionHistoryLimit ` to define it.

api/workloads/v1alpha1/constant.go
```diff
@@ -29,6 +29,18 @@ const (
...
+	// RoleBasedGroup Roles's template. Place it in the rbg controllerrevision label.
+	RevisionKey = RBGPrefix + "controller-revision-hash"
+
+	// RoleRevisionKeyFmt is the labels key used to store the revision hash of
+	// a specific Role template. Placed on the rbg controllerrevision and role workload labels
+	RoleRevisionKeyFmt = RBGPrefix + "role-revision-hash-%s"
+
+	// PodGroupLabelKey identifies pods belonging to a specific pod group
+	// Value: RoleBasedName
+	PodGroupLabelKey = "pod-group.scheduling.sigs.k8s.io/name"
```
gujingit: This key has been moved to scheduler pkg in main branch, pls remove here

api/workloads/v1alpha1/groupversion_info.go
```diff
@@ -34,3 +34,7 @@ var (
 	// AddToScheme adds the types in this group-version to the given scheme.
 	AddToScheme = SchemeBuilder.AddToScheme
 )
+
+const (
+	RoleBasedGroupKind = "RoleBasedGroup"
```
gujingit: No need to add this const.

internal/controller/workloads/rolebasedgroup_controller.go
```diff
@@ -82,6 +82,9 @@ func NewRoleBasedGroupReconciler(mgr ctrl.Manager) *RoleBasedGroupReconciler {
 // +kubebuilder:rbac:groups=workloads.x-k8s.io,resources=rolebasedgroups,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=workloads.x-k8s.io,resources=rolebasedgroups/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=workloads.x-k8s.io,resources=rolebasedgroups/finalizers,verbs=update
+// +kubebuilder:rbac:groups=apps,resources=controllerrevisions,verbs=get;list;watch;create;update;patch;delete
```
gujingit, line 87: error groups value. It should be `workloads.x-k8s.io` not `apps`

bcfre: Here ControllerRevision uses Kubernetes’ built‑in native resource from the apps API group.
controllerrevisions                                                     apps/v1                                       true         ControllerRevision

api/workloads/v1alpha1/constant.go
```diff
@@ -29,6 +29,18 @@ const (
 	// to skip exclusive-topology affinity injection for that role.
 	DisableExclusiveKeyAnnotationKey = RBGPrefix + "disable-exclusive-topology"
 
+	// RevisionKey is the labels key used to store the revision hash of the
+	// RoleBasedGroup Roles's template. Place it in the rbg controllerrevision label.
+	RevisionKey = RBGPrefix + "controller-revision-hash"
```
cheyang: RevisionKey -> RevisionLabelKey

api/workloads/v1alpha1/constant.go
```diff
@@ -29,6 +29,18 @@ const (
...
+
+	// RoleRevisionKeyFmt is the labels key used to store the revision hash of
+	// a specific Role template. Placed on the rbg controllerrevision and role workload labels
+	RoleRevisionKeyFmt = RBGPrefix + "role-revision-hash-%s"
```
cheyang: RoleRevisionKeyFmt -> RoleRevisionLabelKeyFmt.

pkg/reconciler/deploy_reconciler.go
```diff
@@ -61,13 +61,22 @@ func (r *DeploymentReconciler) Reconciler(
 		return fmt.Errorf("convert deployApplyConfig to deployment error: %s", err.Error())
 	}
 
-	equal, err := semanticallyEqualDeployment(oldDeploy, newDeploy, false)
-	if equal {
-		logger.Info("deployment equal, skip reconcile")
-		return nil
+	semanticallyEqual, err := semanticallyEqualDeployment(oldDeploy, newDeploy, false)
+	if err != nil {
+		logger.Info("deployment semanticallyEqual")
```
cheyang: Shouldn’t we return the error if it’s non-nil?

bcfre: In the previous version, the err value was used to pass the differences between the old and new objects, not to indicate an actual processing error. This approach is preserved here for consistency, we can adjust the return value of this semantic consistency in the next PR.

```
	equal, err := semanticallyEqualDeployment(oldDeploy, newDeploy, false)
	if equal {
		logger.Info("deployment equal, skip reconcile")
		return nil
	}

	logger.Info(fmt.Sprintf("deployment not equal, diff: %s", err.Error()))

```

cheyang: I suggest adding comments on that because it's really confusing to the developer.

pkg/utils/revision_utils.go
```diff
@@ -51,3 +69,220 @@ func GetHighestRevision(revisions []*appsv1.ControllerRevision) *appsv1.Controll
...
+	revisions, err := ListRevisions(ctx, client, rbg, selector)
+	if err != nil {
+		return nil, err
+	}
+	highestRevision := GetHighestRevision(revisions)
+	revision := int64(1)
+	if highestRevision != nil {
+		revision = highestRevision.Revision + 1
+	}
+	rawPatch, err := getRBGPatch(rbg)
```
cheyang, line 151: Please add a test case that asks: If the contents of the RoleBasedGroup remain identical except for the value associated with workloadsv1alpha1.RevisionLabelKey, will two successive calls to NewRevision produce different rgbHash values?
In other words, should we examine rawPatch to ensure it is not inadvertently influencing the hash?

cheyang: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubectl/pkg/polymorphichelpers/rollback.go#L422

config/rbac/rbgs_controller_role.yaml
```diff
@@ -152,13 +152,15 @@ rules:
     resources:
       - statefulsets/finalizers
       - deployments/finalizers
+      - controllerrevisions/finalizers
```
cheyang: Is `controllerrevisions/finalizers` necessary?

internal/controller/workloads/rolebasedgroup_controller.go
```diff
@@ -100,6 +103,31 @@ func (r *RoleBasedGroupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		if err := r.client.Create(ctx, expectedRevision); err != nil {
+			logger.Error(err, fmt.Sprintf("failed to create revision %v", expectedRevision))
+			r.recorder.Event(rbg, corev1.EventTypeWarning, FailedCreateRevision, "Failed create revision for RoleBasedGroup")
+			return ctrl.Result{}, err
+		} else {
+			logger.Info(fmt.Sprintf("Create revision [%s] successfully", expectedRevision.Name))
+			r.recorder.Event(rbg, corev1.EventTypeNormal, SucceedCreateRevision, "Successful create revision for RoleBasedGroup")
+		}
+	}
+	expectedRoleRevisionHash, err := utils.GetRolesRevisionHash(expectedRevision)
```
cheyang: Do we need a stable JSON canonicalization (e.g. sort map keys) before hashing? and make sure we are using a single global hash for the RBG.

bcfre: json.Marshal() will sort by default.
```
// The map keys 
// are sorted and used as JSON object keys by applying the following rules,
// subject to the UTF-8 coercion described for string values above:
//   - keys of any string type are used directly
//   - keys that implement [encoding.TextMarshaler] are marshaled
//   - integer keys are converted to strings
```

internal/controller/workloads/rolebasedgroup_controller.go
```diff
@@ -341,6 +370,41 @@ func (r *RoleBasedGroupReconciler) ReconcileScalingAdapter(
...
+func (r *RoleBasedGroupReconciler) getOrCreateRevisionIfNotExist(ctx context.Context, rbg *workloadsv1alpha1.RoleBasedGroup) (*appsv1.ControllerRevision, error) {
+	selector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			workloadsv1alpha1.SetNameLabelKey: rbg.Name,
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
+	revisions, err := utils.ListRevisions(ctx, r.client, rbg, selector)
```
cheyang, line 382: I suggest eliminating the duplicate call to ListRevisions by extracting the listing logic into a single shared helper and letting both NewRevision and getOrCreateRevisionIfNotExist reuse it.

cheyang (changes requested):

internal/controller/workloads/rolebasedgroup_controller.go
```diff
@@ -100,6 +102,32 @@ func (r *RoleBasedGroupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	ctx = ctrl.LoggerInto(ctx, logger)
 	logger.Info("Start reconciling")
 
+	currentRevision, err := r.getOrCreateRevisionIfNotExist(ctx, rbg)
+	if err != nil {
+		logger.Error(err, "Failed get or create revision")
+		return ctrl.Result{}, err
+	}
+	expectedRevision, err := utils.NewRevision(ctx, r.client, rbg)
```
cheyang: I think  currentRevision is already the latest revision returned by getOrCreateRevisionIfNotExist, pass it directly to NewRevision as an input parameter instead of recomputing it—this removes the duplicate ListRevisions call.

internal/controller/workloads/rolebasedgroup_controller.go
```diff
@@ -341,6 +370,41 @@ func (r *RoleBasedGroupReconciler) ReconcileScalingAdapter(
...
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
+	revisions, err := utils.ListRevisions(ctx, r.client, rbg, selector)
+	if err != nil {
+		return nil, err
+	}
+	revisions, err = utils.CleanExpiredRevision(ctx, r.client, rbg, revisions)
```
cheyang: I suggest moving the cleanup logic for expired revisions (CleanExpiredRevision) into its own step within the reconcile loop; do not embed it inside getOrCreateRevisionIfNotExist.

cheyang (changes requested):

internal/controller/workloads/rolebasedgroup_controller.go
cheyang: Just like https://github.com/kubernetes-sigs/lws/blob/main/pkg/controllers/leaderworkerset_controller.go#L179

PR #42, opened 2025-09-25
feat: add order

tlipoca9:
### Ⅰ. Motivation

Improve the role dependency sorting algorithm in RoleBasedGroup to support parallel execution of roles at the same dependency level, enhancing startup efficiency for distributed workloads. The original topological sorting algorithm could only execute roles sequentially, unable to leverage parallelism between roles with identical dependencies.

#### Details

In the current implementation, roles have implicit dependencies based on their definition order. In LLM scenarios, when loading a large model from S3 (typically taking 10-15 minutes), if multiple roles (like prefill role and decode role) need to be started simultaneously, waiting 10-15 minutes to start the second role is inefficient and unreasonable.

### Ⅱ. Modifications

1. **Dependency Sorting Algorithm Refactor**: Modified `dependencyOrder` function to return layered lists `[][]string` instead of a single sorted list, enabling parallel execution of roles at the same dependency level.
2. **Interface Changes**: Updated `DependencyManager.SortRoles` interface to return `[][]*workloadsv1alpha1.RoleSpec` instead of `[]*workloadsv1alpha1.RoleSpec`.
3. **Controller Adaptation**: Modified `PodReconciler` and `RoleBasedGroupReconciler` to support layered role execution, using error aggregation instead of immediate error returns.
4. **Enhanced Test Cases**: Added test cases for identical dependencies to validate parallel execution correctness.

NONE

- Added "same dependency" test cases to verify multiple roles with identical dependencies are correctly assigned to the same execution level.
- Updated existing test case expected results to align with the new layered sorting algorithm.

1. Run unit tests: `go test ./pkg/dependency/... -v`
2. Create RoleBasedGroup instances containing multiple roles with identical dependencies, verify roles can start in parallel.
3. Check controller logs to confirm roles at the same level are processed concurrently.

- Algorithm changed from depth-first search to dependency-depth-based level assignment.
- Uses `stderrors.Join` to aggregate errors within the same level instead of immediate returns.
- Maintains backward-compatible API design while extending return types.

## Checklist

### DIFF
--- a/internal/controller/workloads/pod_controller.go
+++ b/internal/controller/workloads/pod_controller.go
@@ -1,6 +1,7 @@
 package workloads
 
 import (
+	stderrors "errors"
 	"fmt"
 
 	"golang.org/x/net/context"
@@ -73,14 +74,24 @@ func (r *PodReconciler) restartRBG(ctx context.Context, rbg *workloadsv1alpha1.R
 	if err != nil {
 		return err
 	}
-	for _, role := range sortedRoles {
-		recon, err := reconciler.NewWorkloadReconciler(role.Workload, r.scheme, r.client)
-		if err != nil {
-			return err
+	for _, roleList := range sortedRoles {
+		var joinedError error
+
+		for _, role := range roleList {
+			recon, err := reconciler.NewWorkloadReconciler(role.Workload, r.scheme, r.client)
+			if err != nil {
+				joinedError = stderrors.Join(joinedError, err)
+				continue
+			}
+			// 3. recreate role
+			if err := recon.RecreateWorkload(ctx, rbg, role); err != nil {
+				joinedError = stderrors.Join(joinedError, err)
+				continue
+			}
 		}
-		// 3. recreate role
-		if err := recon.RecreateWorkload(ctx, rbg, role); err != nil {
-			return err
+
+		if joinedError != nil {
+			return joinedError
 		}
 	}
 
--- a/internal/controller/workloads/rolebasedgroup_controller.go
+++ b/internal/controller/workloads/rolebasedgroup_controller.go
@@ -18,6 +18,7 @@ package workloads
 
 import (
 	"context"
+	stderrors "errors"
 	"fmt"
 	"reflect"
 	"sync"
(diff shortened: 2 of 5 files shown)


internal/controller/workloads/pod_controller.go
```diff
@@ -73,14 +74,24 @@ func (r *PodReconciler) restartRBG(ctx context.Context, rbg *workloadsv1alpha1.R
...
 		return err
 	}
-	for _, role := range sortedRoles {
-		recon, err := reconciler.NewWorkloadReconciler(role.Workload, r.scheme, r.client)
-		if err != nil {
-			return err
+	for _, roleList := range sortedRoles {
+		var joinedError error
+
+		for _, role := range roleList {
```
gujingit, line 80: It seems that we still reconcile roles serially here? Should we switch to using go routines?

internal/controller/workloads/pod_controller.go
```diff
@@ -1,6 +1,7 @@
 package workloads
 
 import (
+	stderrors "errors"
```
bcfre: Why do we need to rename "errors" to "stderrors"

internal/controller/workloads/pod_controller.go
```diff
@@ -73,14 +74,24 @@ func (r *PodReconciler) restartRBG(ctx context.Context, rbg *workloadsv1alpha1.R
 	if err != nil {
 		return err
 	}
-	for _, role := range sortedRoles {
-		recon, err := reconciler.NewWorkloadReconciler(role.Workload, r.scheme, r.client)
-		if err != nil {
-			return err
+	for _, roleList := range sortedRoles {
+		var joinedError error
```
bcfre: Maybe changing joinedError to errs or allErrs or aggregatedErrs would be more intuitive

internal/controller/workloads/rolebasedgroup_controller.go
```diff
@@ -121,63 +122,75 @@ func (r *RoleBasedGroupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			dynamicWatchCustomCRD(roleCtx, role.Workload.Kind)
+			// Check dependencies first
+			ready, err := dependencyManager.CheckDependencyReady(roleCtx, rbg, role)
+			if err != nil {
+				r.recorder.Event(rbg, corev1.EventTypeWarning, FailedCheckRoleDependency, err.Error())
+				return ctrl.Result{}, err
+			}
+			if !ready {
+				logger.Info("Dependencies not met, requeuing", "role", role.Name)
+				return ctrl.Result{RequeueAfter: 5}, nil
```
bcfre: Since the operator is event-driven, changes will naturally trigger reconciliation. If there are no incoming events (i.e., the underlying state hasn’t changed), is periodic reconciliation here still needed?

internal/controller/workloads/pod_controller.go
tlipoca9: I don't think it's necessary to use goroutines in this context. Considering that the normal reconciliation logic typically executes very quickly, introducing goroutines here seems unnecessary. However, if you believe it's genuinely required, I'm willing to make the necessary modifications.

internal/controller/workloads/pod_controller.go
tlipoca9: it is unnecessary, fix

internal/controller/workloads/pod_controller.go
tlipoca9: ok

internal/controller/workloads/rolebasedgroup_controller.go
tlipoca9: You're right, requeue can probably be removed here. 
I'm not entirely sure if my understanding is correct：when a pod's status changes, it will trigger event notification through the pod_controller.

pkg/dependency/dependency_test.go
```diff
@@ -73,7 +79,10 @@ func TestDependencyOrder(t *testing.T) {
 				"b": {"c"},
 				"c": {},
 			},
-			want:    []string{"c", "a", "b"},
+			want: [][]string{
+				{"c"},
+				{"a", "b"},
+			},
```
bcfre, line 85: Please add some more complex cases, such as
```go
{
			name: "complex dependencies",
			dependencies: map[string][]string{
				"a": {"c", "b", "d"},
				"b": {"c"},
				"c": {},
				"d": {},
				"e": {"f"},
				"f": {"d"},
			},
			want: [][]string{
				{"c", "d"},
				{"b", "f"},
				{"a", "e"},
			},
			wantErr: false,
		},
```

internal/controller/workloads/rolebasedgroup_controller.go
```diff
@@ -18,6 +18,7 @@ package workloads
 
 import (
 	"context"
+	stderrors "errors"
```
bcfre, line 21: Please remove the rename

internal/controller/workloads/rolebasedgroup_controller.go
bcfre: Yes, as far as I know. To keep the impact scope minimal, let’s leave this logic untouched for now and handle the optimization in a follow-up PR.

internal/controller/workloads/pod_controller.go
cheyang: I agree with @tlipoca9 because it's eliminating dependency blocking through layered processing rather than using goroutines here.

cheyang:
@tlipoca9 Thank you for optimizing the RBG logic! We truly appreciate your expertise and look forward to your continued contributions to the project. Please don't hesitate to reach out if you have any questions or need support. If possible， Please let me know how to find you by email or slack!

PR #53, opened 2025-10-14
fix: change stateful set service name to meet k8s requirements

liubing0427:
### Ⅰ. Motivation
The name of a K8s Service object must be a valid [RFC 1035 label name](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#rfc-1035-label-names). Must start with an alphabetic character. But now the code uses the same name as rbg name，rbg name can start with numeric. When create rbg use a name start with numeric, will error.

### Ⅱ. Modifications
service name use a new function

NONE

## Checklist

### DIFF
--- a/api/workloads/v1alpha1/helper.go
+++ b/api/workloads/v1alpha1/helper.go
@@ -46,7 +46,6 @@ func (rbg *RoleBasedGroup) GetServiceName(role *RoleSpec) string {
 	}
 	return fmt.Sprintf("%s-%s", rbg.Name, role.Name)
 }
-}
 
 func (rbg *RoleBasedGroup) GetRole(roleName string) (*RoleSpec, error) {
 	if roleName == "" {


api/workloads/v1alpha1/helper.go
```diff
@@ -39,6 +39,10 @@ func (rbg *RoleBasedGroup) GetWorkloadName(role *RoleSpec) string {
 	return fmt.Sprintf("%s-%s", rbg.Name, role.Name)
 }
 
+func (rbg *RoleBasedGroup) GetServiceName(role *RoleSpec) string {
```
cheyang, line 43: The change introduces a new service name format (svc-<rbg.Name>-<role.Name>) instead of reusing the workload name (<rbg.Name>-<role.Name>).  This breaks backward compatibility because Existing Services created by prior versions will no longer be recognized by the new code. The reconciler will attempt to query Services with the svc- prefix, fail with IsNotFound, and recreate them, potentially deleting old Services unintentionally.

How about checking if the rbg object is created by old version controller?

```golang
func (rbg *RoleBasedGroup) GetServiceName(role *RoleSpec) string {
    oldName := fmt.Sprintf("%s-%s", rbg.Name, role.Name)
    if exists, _ := CheckServiceExists(oldName); exists {
        return oldName
    }
    return fmt.Sprintf("svc-%s-%s", rbg.Name, role.Name)
}
```

cheyang: Kubernetes Service names have a 63-character limit. Adding the svc- prefix risks exceeding this limit if rbg.Name or role.Name. what's your plan to handle this logic?

liubing0427: I have changed the logic, only the name start with numeric add prefix "s-".
Regarding the problem of excessive length, if it is handled, the service name will become less predictable, so will not deal with it here.

Issue #56, opened 2025-10-15
[Bug] Update rbg and remove podGroupPolicy, but existing podgroup remains


ShirleyDing:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

Firstly, create a RoleBasedGroup with gang-scheduling enabled, then, update the RoleBasedGroup and disable gang-scheduling, but PodGroup created by the RoleBasedGroup remains.

### Steps to reproduce
1. Deploy a RoleBasedGroup
```
apiVersion: workloads.x-k8s.io/v1alpha1
kind: RoleBasedGroup
metadata:
  name: gang-scheduling
spec:
  podGroupPolicy:
    volcanoScheduling:
      priorityClassName: test
      queue: test
  roles:
    - name: role-sts
      replicas: 1
      template:
        spec:
          containers:
            - name: sts
              image: anolis-registry.cn-zhangjiakou.cr.aliyuncs.com/openanolis/nginx:1.14.1-8.6
              ports:
                - containerPort: 80
              resources:
                requests:
                  nvidia.com/gpu: "1"
                limits:
                  nvidia.com/gpu: "1"

    - name: role-deploy
      replicas: 1
      workload:
        apiVersion: apps/v1
        kind: Deployment
      template:
        spec:
          containers:
            - name: deploy
              image: anolis-registry.cn-zhangjiakou.cr.aliyuncs.com/openanolis/nginx:1.14.1-8.6
              ports:
                - containerPort: 80
              resources:
                requests:
                  nvidia.com/gpu: "1"
                limits:
                  nvidia.com/gpu: "1"
```
2. the corresponding volcano podgroup created together
rbg:
<img width="244" height="40" alt="Image" src="https://github.com/user-attachments/assets/4af7d645-7986-444e-a755-d5dab3ff591a" />

volcano podgroup:
[...]

PR #57, opened 2025-10-15
fix: delete corresponding podgroup created by rbg when gang-schedulin…

ShirleyDing:
### Ⅰ. Motivation
bugfix

### Ⅱ. Modifications
podgroup_manager and rolebasedgroup_controller reconcile logic

fixes [#56](https://github.com/sgl-project/rbg/issues/56)

1. create a RoleBasedGroup with gang-scheduling enabled
2. update the RoleBasedGroup and remove spec.podGroupPolicy
3. check if the podGroup created by the RoleBasedGroup is deleted

## Checklist

### DIFF
--- a/internal/controller/workloads/rolebasedgroup_controller.go
+++ b/internal/controller/workloads/rolebasedgroup_controller.go
@@ -141,14 +141,11 @@ func (r *RoleBasedGroupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, err
 	}
 
-	podGroupExist := scheduler.GetPodGroupAndLoadCrdName(rbg, runtimeController, &watchedWorkload, r.apiReader)
 	// Process PodGroup
-	if podGroupExist {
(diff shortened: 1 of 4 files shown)


bcfre:
Thank you for identifying this issue and providing a possible solution.
I’d like to add another angle to consider: Should we support this field being mutable? @Syspretor @cheyang @ShirleyDing 

The specific background is that when we use PodGroup for gang scheduling, we inject relevant gang scheduling information into the pod’s labels or annotations (https://github.com/sgl-project/rbg/blob/main/pkg/scheduler/podgroup_manager.go#L52).

If changes to this field are allowed, it could lead to modifications in the workload spec, which in turn would trigger the underlying pods to be recreated/updated.
We should consider whether pod recreation caused solely by enabling or disabling gang scheduling behavior is actually expected.

cheyang:
@ShirleyDing Thanks for your fix! Look forward to your next PR!

PR #58, opened 2025-10-16
bugfix: add max len check for workloadName & serviceName

gujingit:
feat: svcName changed to new format "s-rbg.Name-role.Name"

### Ⅰ. Motivation

bugfix: add max len check for workloadName & serviceName

### Ⅱ. Modifications

bugfix: add max len check for workloadName & serviceName
feat: change serviceName in new format "s-rbg.Name-role.Name"

NONE

make test

## Checklist

### DIFF
--- a/api/workloads/v1alpha1/helper.go
+++ b/api/workloads/v1alpha1/helper.go
@@ -5,7 +5,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"unicode"
+	"strings"
 )
 
 func (rbg *RoleBasedGroup) GetCommonLabelsFromRole(role *RoleSpec) map[string]string {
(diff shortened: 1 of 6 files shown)


api/workloads/v1alpha1/helper.go
```diff
@@ -37,14 +37,32 @@ func (rbg *RoleBasedGroup) GetGroupSize() int {
...
+	return workloadName
 }
 
+// GetServiceName Because ServiceName needs to follow DNS naming conventions,
+// which do not allow names to start with a number. Therefore, the s- prefix
+// is added to the service name to meet this requirement.
 func (rbg *RoleBasedGroup) GetServiceName(role *RoleSpec) string {
-	if len(rbg.Name) > 0 && unicode.IsDigit(rune(rbg.Name[0])) {
-		return fmt.Sprintf("s-%s-%s", rbg.Name, role.Name)
+	svcName := fmt.Sprintf("s-%s-%s", rbg.Name, role.Name)
```
cheyang, line 58: The change introduces a new service name format (s-<rbg.Name>-<role.Name>) instead of reusing the workload name (<rbg.Name>-<role.Name>).  This breaks backward compatibility because Existing Services created by prior versions will no longer be recognized by the new code. The reconciler will attempt to query Services with the svc- prefix, fail with IsNotFound, and recreate them, potentially deleting old Services unintentionally.

How about checking if the rbg object is created by old version controller?

```golang
func (rbg *RoleBasedGroup) GetServiceName(role *RoleSpec) string {
    oldName := fmt.Sprintf("%s-%s", rbg.Name, role.Name)
    if exists, _ := CheckServiceExists(oldName); exists {
        return oldName
    }
    return fmt.Sprintf("s-%s-%s", rbg.Name, role.Name)
}
```

cheyang (approved):
/lgtm
/approve

PR #59, opened 2025-10-17
KEP-30: add role coordination kep

gujingit:
### Ⅱ. Modifications

Add Role coordination KEP

NONE

## Checklist

### DIFF
--- a/keps/30-role-coordination/README.md
+++ b/keps/30-role-coordination/README.md
@@ -11,7 +11,8 @@
     - [Non-Goals](#non-goals)
 - [Proposal](#proposal)
     - [User Stories](#user-stories)
-        - [Story 1: Coordinated Rolling Update](#story-1-coordinated-rolling-update)
+        - [Story 1](#story-1)
+        - [Story 2](#story-2)
     - [Implementation Details](#implementation-details)
         - [API Changes](#api-changes)
         - [Example](#yaml-example)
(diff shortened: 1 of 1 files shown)


keps/30-role-coordination/kep.yaml
```diff
@@ -0,0 +1,51 @@
+title: KEP Template
+kep-number: NNNN
+authors:
+  - "@jane.doe"
```
cheyang: Please update the information of this file.

keps/30-role-coordination/README.md
```diff
@@ -0,0 +1,270 @@
...
+## Design Details
+
+The implementation will modify the main reconciliation loop
+in [RoleBasedGroupReconciler] to check for a coordination strategy. If present, it will execute the coordinated
+update logic; otherwise, it will fall back to the existing independent role update behavior.
+
+Each coordination step will:
+
+1. Apply the specified strategies to the relevant roles
+2. Monitor the status of those roles
```
cheyang: I suggest adding the detail design to instruct the router component to split requests between the old and new Pods in real time during a RoleBasedGroup coordinated upgrade, so that it  can integrate with SGLang Router's rolling update workflow

cheyang (changes requested):

keps/30-role-coordination/README.md
```diff
@@ -0,0 +1,334 @@
...
+        - role: planner
+          updateStrategy:
+            batchSize: 1
+```
+
+- Updating with pause
+40 prefill pods and 10 decode pods were upgraded in a rolling update with a 4:1 ratio of prefill to decode.
+After rolling updates of 20 prefill pods and 5 decode pods, the process was paused.
+```yaml
+apiVersion: workloads.x-k8s.io/v1alpha1
```
cheyang, line 179: I recommend retaining support for multiple strategy types such as `updateStrategy`, `scaleStrategy`, and `gangStrategy`, while using a unified `config` structure internally implemented as a map. The API would be:

```yaml
apiVersion: workloads.x-k8s.io/v1alpha1
kind: RoleBasedGroup
spec:
  coordinations:
    - name: "prefill-decode"
      strategies:
        - role: prefill
          strategyType: updateStrategy   # Explicit strategy type
          config:                        # Universal map-based config
            partition: 20
            batchSize: 2
        - role: decode
          strategyType: updateStrategy
          config:
            partition: 5
            batchSize: 1
            
    - name: "router-planner"
      strategies:
        - role: router
          strategyType: gangStrategy    # Different strategy type
          config:                       # Same config structure
            minAvailable: 2
            timeout: 10m
        - role: planner
          strategyType: scaleStrategy   # Third strategy type
          config:
            minReplicas: 3
            maxReplicas: 10
```

```yaml
apiVersion: workloads.x-k8s.io/v1alpha1
kind: RoleBasedGroup
metadata:
  name: role-coordination
spec:
  coordination:
    - name: "prefill-decode-update" 
      type: updateStrategy # Explicit strategy type
      strategy:
      - role: prefill  
        config:   # Universal map-based config
          type: Recreate # [Recreate, InplaceIfPossible], default is Recreate
          batchSize: 2
      - role: decode
        config:
          type: Recreate # [Recreate, InplaceIfPossible], default is Recreate
          batchSize: 1
    # other coordination strategy type demo
    - name: "prefill-decode-scaling" # Support later
      type: scalingStrategy    # Different strategy type
      strategies:
        - role: prefill
          config:                       # Same config structure
            batchSize: 2
        - role: decode
          config:
            batchSize: 1
    - name: "prefill-decode-gang" # Other type will support on demand
      type: gangStrategy    # Different strategy type
      strategies:
        - role: prefill
          config:                       # Same config structure
            minAvailable: 2
            timeout: 10m
        - role: decode
          config:
            minAvailable: 2
            timeout: 10m
```


### Key Implementation Features:

1. **Unified Config Structure**
   ```yaml
   config:  # Common structure for all strategies
     key1: value1
     key2: value2
     ...
   ```

2. **Strategy Type Flexibility**
   ```go
   type Strategy struct {
       Role        string
       StrategyType string    // updateStrategy|scaleStrategy|gangStrategy
       Config      map[string]string
   }
   ```

3. **Validation Handling**
   ```go
   func Validate(config map[string]string, strategyType string) error {
       switch strategyType {
       case "updateStrategy":
           return validateUpdateConfig(config)
       case "gangStrategy":
           return validateGangConfig(config)
       case "scaleStrategy":
           return validateScaleConfig(config)
       default:
           return fmt.Errorf("unknown strategy type: %s", strategyType)
       }
   }
   ```


This maintains API consistency while enabling diverse strategy implementations through:
- Single `config` field with flexible key-value pairs
- Explicit `strategyType` designation for behavior dispatch
- Backward compatibility with existing updateStrategy configurations
- Clean extension path for future strategy types

ZYecho11: Why is it necessary to set gangPolicy under the coordinates field? What is User Story？

cheyang: I need to deploy a minimal structure consisting of at least one router, two prefiller, and two decoder instances using RBG. 
```yaml
apiVersion: workloads.x-k8s.io/v1alpha1
kind: RoleBasedGroup
spec:
  coordinations:
    - name: "minimal-deployment"
      strategies:
        - role: router
          strategyType: gangStrategy
          config:
            minAvailable: 1   # Minimum 1 router
            
        - role: prefill
          strategyType: gangStrategy
          config:
            minAvailable: 2   # Minimum 2 prefill
            
        - role: decode
          strategyType: gangStrategy
          config:
            minAvailable: 2   # Minimum 2 decode
```

ZYecho11: isn't this information already defined in the podgroup and podgroupPolicy? Why do we need to repeat the definition？

cheyang: The  difference is that coordination needs to support *multiple atomic gangs*. For example:  
- Role `prefill` requires 22 replicas  
- Role `decode` requires 10 replicas  
- But each atomic gang must contain exactly 2 `prefill` + 1 `decode` pods  each time to place in the same node.

This necessitates *coordinated sequential gangs* – a single PodGroup cannot express this multi-batch requirement.

ZYecho11: thanks, get it

keps/30-role-coordination/README.md
```diff
@@ -0,0 +1,322 @@
...
+#### Coordinated-roles Upgrading Yaml Example
+- Automatic updating without pause, it will be upgraded in a rolling update with a 4:1 ratio of prefill to decode.
+```yaml
+apiVersion: workloads.x-k8s.io/v1alpha1
+kind: RoleBasedGroup
+metadata:
+  name: role-coordination
+spec:
+  coordination:
+    - name: "prefill-decode-update"  # strategy 1: reconcile prefill & decode at 4:1 ratio
```
veophi: How about
```yaml
coordination:
- name: prefill-decode-update
  type: RollingUpdate
  roles:
  - prefill
  - decode
  strategy:
    maxSkew: 1% # the max skew of updated replicas between prefill and decode. For example, one rbg with (200p, 100d) allows `abs(updated_prefill/200 - updated_decode/100) < 1%`.
    partition: 99%
```

keps/30-role-coordination/README.md
```diff
@@ -0,0 +1,322 @@
...
+package controller
+
+func (r *RoleBasedGroupReconciler) Reconcile() {
+	if r.spec.Coordination != nil {
+		r.executeCoordinationStrategy()
+	}
+	// existing independent role update behavior
+}
+
+func (r *RoleBasedGroupReconciler) executeCoordinationStrategy() {
```
veophi: Behavior when **a single Pod** becomes unhealthy during rollout?

If one Pod enters an Unhealthy state (e.g., readiness/failure probe fails due to some node problems) during a rolling upgrade, will the rbg controller block the entire upgrade — even when `maxUnavailable` is explicitly set to >1?

keps/30-role-coordination/README.md
gujingit: This design will binds the rolling update ratio to the replicas of prefill and decode. If the prefill or decode has HPA , their replica count is not accurate, the update ratio will not meet expectations.

gujingit: Also this design can not support this case. There are 40 Prefill and 10 Decode pods and users want to update prefill:decode = 2:1.

keps/30-role-coordination/README.md
gujingit: The rolling update process of the rbg controller follows the K8s sts design. If a old version pod enters an unhealthy state, it will not block the upgrade. If a new version pod (only single pod) enters an unhealthy state, it will block the upgrade process.

PR #95, opened 2025-11-11
chore(hack): add script to generate manifests.yaml

Syspretor:
### Ⅰ. Motivation
Currently, during releases we only update the Helm content, and the manifest.yaml used by kubectl is not updated. This PR support to generate the manifest from the Helm chart to keep the release manifest.yaml in sync.
1. Rename `$RBG_HOME/tools/release/version-update.sh` to  `$RBG_HOME/tools/release/update-release.sh`
2. Add codes to generate the manifest from the Helm chart 

fixes #XXXX

## Checklist

### DIFF
--- a/tools/release/update-release.sh
+++ b/tools/release/update-release.sh
@@ -8,6 +8,9 @@ cd "$REPO_ROOT" || { echo "Error: Failed to change to repo root"; exit 1; }
 # Set paths relative to repo root
 CHARTS_DIR="deploy/helm"
 
+# Set paths relative to manifest root
+MANIFEST_DIR="deploy/kubectl"
+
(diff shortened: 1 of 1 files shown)


hack/deploy/generate-kubectl-manifests.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+# USAGE:
+#   RBG_HOME=/path/to/rbg
+#   cd $RBG_HOME
+#   bash hack/deploy/generate-kubectl-manifests.sh
```
cheyang: Please merge this script into tools/release/version-update.sh

PR #107, opened 2025-11-18
rbgs support pod group policy

nightmeng:
### Ⅰ. Motivation

  - Fix the bug where a RoleBasedGroupSet with podGroupPolicy creates RoleBasedGroups that do not inherit that policy, so PodGroup metadata never reaches the RoleBasedGroup and gang scheduling breaks.

  ### Ⅱ. Modifications

  - In the RoleBasedGroupSet reconciler, detect whether podGroupPolicy is set on the RBGS and, if so, propagate the field to every generated RoleBasedGroup so it can correctly handle PodGroup creation.

  NONE

  - No new tests; the change is a small wiring fix and is already covered by existing RBGS/RBG e2e flows.

  - Deploy the updated controller, create an RBGS that specifies podGroupPolicy, and confirm each resulting RoleBasedGroup now contains the same podGroupPolicy.
  - Observe via kubectl get podgroup (or Volcano equivalent) that PodGroups are created/updated for each RoleBasedGroup.
  - Remove the RBGS and ensure the associated RoleBasedGroups and PodGroups are garbage collected.

  - Double-check that the propagation only happens when the RBGS actually specifies podGroupPolicy, to avoid altering RoleBasedGroups that do not need it.

  ## Checklist
  - NONE

### DIFF
--- a/internal/controller/workloads/rolebasedgroupset_controller.go
+++ b/internal/controller/workloads/rolebasedgroupset_controller.go
@@ -479,7 +479,8 @@ func newRBGForSet(rbgset *workloadsv1alpha1.RoleBasedGroupSet, index int) *workl
 			// The OwnerReference will be set in the scaleUp function.
 		},
 		Spec: workloadsv1alpha1.RoleBasedGroupSpec{
-			Roles: rbgset.Spec.Template.Roles,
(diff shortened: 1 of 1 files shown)


Syspretor:
@nightmeng Thanks for this contribution!

PR #108, opened 2025-11-20
chore(rbg): migrate role.leaderWorkerSet field type from value to poi…

Syspretor:
### Ⅰ. Motivation

On the one hand, RBG lets users flexibly define the workload for each role (e.g., Deployment, StatefulSet, LeaderWorkerSet, InstanceSet). On the other hand, RBG supports both single-node replicated deployments (via role.template) and multi-node leader/worker deployments (via role.leaderWorkerSet), and will support even more flexible configurations in the future (roles.components). Therefore, we need to migrate the template and leaderWorkerSet fields to pointer types to:

- avoid default values misleading users
- facilitate validation of the role’s declared deployment shape across different workloads.

#### API Layer

- RoleSpec.LeaderWorkerSet -> `LeaderWorkerTemplate` -> `*LeaderWorkerTemplate`
- RoleSpec.LeaderWorkerSet.PatchLeaderTemplate `runtime.RawExtension` -> `*runtime.RawExtension`
- RoleSpec.LeaderWorkerSet.PatchWorkerTemplate `runtime.RawExtension` -> `*runtime.RawExtension`

#### Controller Logic:
- Handle nil LeaderWorkerSet gracefully with zero-value fallback
- Maintains backward compatibility: existing behavior unchanged
- Internal functions updated to use pointer throughout
- Add a Validate method to the WorkloadReconciler interface to let each workload implement its own validation of the role’s deployment shape:

  - DeploymentReconciler: role.Template must not be empty (nil).
  - StatefulSetReconciler: role.Template must not be empty (nil).
  - LeaderWorkerSetReconciler: if Template is empty (nil), neither PatchLeaderTemplate nor PatchWorkerTemplate may be empty (nil).

## Checklist


api/workloads/v1alpha1/helper.go
```diff
@@ -28,6 +28,10 @@ func (rbg *RoleBasedGroup) GetGroupSize() int {
 	ret := 0
 	for _, role := range rbg.Spec.Roles {
 		if role.Workload.String() == LeaderWorkerSetWorkloadType {
+			if role.LeaderWorkerSet == nil {
+				ret += 1 * int(*role.Replicas)
```
bcfre, line 32: This will cause a panic if it's a nil pointer. Can we guarantee that it won't be nil here?

internal/controller/workloads/rolebasedgroup_controller.go
```diff
@@ -168,10 +148,20 @@ func (r *RoleBasedGroupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+				"Failed to reconcile role %s, err: %v", role.Name, err,
 			)
 			return ctrl.Result{}, err
 		}
+
+		if err := reconciler.Validate(ctx, &role); err != nil {
+			logger.Error(err, "Failed to validate role declaration")
+			r.recorder.Eventf(
+				rbg, corev1.EventTypeWarning, FailedReconcileWorkload,
+				"Failed to validate role %s declaration, err: %v", role.Name, err,
```
bcfre, line 160: Should we hard‑code the detailed event reasons? Keeping event reasons stable helps Kubernetes aggregate events and reduces pressure on the cluster.

PR #113, opened 2025-11-26
feat(rbg): support role-level metadata

Syspretor:
### Ⅰ. Motivation

Support configuring Role-level labels/annotations; these settings will be propagated to the labels/annotations of workloads created from the Role definition.

### Ⅱ. Modifications

Introducing new API Fields under `rbg.spec.role`

```go
// RoleSpec defines the specification for a role in the group
type RoleSpec struct {
	// Unique identifier for the role
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:MinLength=1
	Name string `json:"name"`

	// Map of string keys and values that can be used to organize and categorize
	// (scope and select) objects. May match selectors of replication controllers
	// and services.
	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels
	// +optional
	Labels map[string]string `json:"labels,omitempty"`

	// Annotations is an unstructured key value map stored with a resource that may be
	// set by external tools to store and retrieve arbitrary metadata. They are not
	// queryable and should be preserved when modifying objects.
	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations
	// +optional
	Annotations map[string]string `json:"annotations,omitempty"`

	// ......
}
```

fixes #XXXX

1. create RBG with the following yaml contains a custom label declaration of a role:
```yaml
apiVersion: workloads.x-k8s.io/v1alpha1
kind: RoleBasedGroup
metadata:
  name: demo
spec:
  roles:
  - name: engine
    labels:
      custom-label-key:"custome-label-value"
    replicas: 10
    template:
      spec:
        containers:
        - name: main
          image: nginx:latest
```
2. check the custom label of Statefulset created by RBG
```shell
# kubectl get sts demo-engine -o jsonpath="{.metadata.labels['custom-label-key']}{'\n'}"
custome-label-value
```

## Checklist


api/workloads/v1alpha1/rolebasedgroup_types.go
```diff
@@ -200,6 +200,10 @@ type RoleSpec struct {
 	// +kubebuilder:validation:MinLength=1
 	Name string `json:"name"`
 
+	// RoleMetadata contains standard metadata labels and annotations that will be applied to the role's workload.
+	// +optional
+	RoleMetadata `json:"metadata,omitempty"`
```
cheyang: How about defining labels and annotations instead of RoleMetadata? I am concerned that metadata also contains `name`. WTDY?

```golang
// Labels is a map of string keys and values that can be used to organize and categorize
	// +optional
	Labels map[string]string `json:"labels,omitempty"`
	// Annotations is an unstructured key value map stored with a resource that may be
	// set by external tools to store and retrieve arbitrary metadata. They are not
	// queryable and should be preserved when modifying objects.
	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations
	// +optional
	Annotations map[string]string `json:"annotations,omitempty"`
	// TopologyConstraint defines topology placement requirements for PodClique.
	// Must be equal to or stricter than parent resource constraints.
	// +optional
```

PR #125, opened 2025-12-11
example(rbg): add example for mooncake integration

Syspretor:
### Ⅰ. Motivation
add example for mooncake integration with RBG

fixes #XXXX

## Checklist

### DIFF
--- a/examples/mooncake/Dockerfile.mooncake
+++ b/examples/mooncake/Dockerfile.mooncake
@@ -1,0 +1,32 @@
+# Stage 1: Builder
+# This stage downloads dependencies, and install Mooncake from PyPI.
+FROM ubuntu:22.04
+
+ENV DEBIAN_FRONTEND=noninteractive
+ENV PYTHONUNBUFFERED=1
+
+# Install only runtime dependencies
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    netcat \
+    curl \
+    python3 \
+    python3-pip \
+    python3-venv \
+    infiniband-diags \
+    ibverbs-providers \
+    rdma-core \
+    libibverbs1 \
+    libibverbs-dev \
+    ibverbs-utils \
+    librdmacm1 \
+    libnuma1 \
+    liburing2 \
+    libyaml-0-2 \
+    && apt-get clean && rm -rf /var/lib/apt/lists/*
+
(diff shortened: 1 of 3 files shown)


examples/mooncake/Dockerfile.mooncake
```diff
@@ -0,0 +1,32 @@
...
+    libnuma1 \
+    liburing2 \
+    libyaml-0-2 \
+    && apt-get clean && rm -rf /var/lib/apt/lists/*
+
+# Create venv and install Mooncake Transfer Engine from PyPI
+RUN python3 -m venv /opt/venv
+ENV PATH="/opt/venv/bin:$PATH"
+RUN pip install --upgrade pip setuptools wheel
+RUN pip install mooncake-transfer-engine==0.3.7
```
whybeyoung: latest release should be 0.3.7.post2

examples/mooncake/pd-disaggregated-with-mooncake.yaml
```diff
@@ -0,0 +1,257 @@
...
+              image: lmsysorg/sglang:v0.5.5
+              command:
+                - python3
+                - -m
+                - sglang_router.launch_router
+                - --pd-disaggregation
+                - --prefill
+                - http://sglang-pd-with-mooncake-demo-prefill-0.s-sglang-pd-with-mooncake-demo-prefill:8000
+                - --decode
+                - http://sglang-pd-with-mooncake-demo-decode-0.s-sglang-pd-with-mooncake-demo-decode:8000
```
whybeyoung, line 99: their is a option according to the  k8s's label selector . rather than the hard code with prefill or decode url

examples/mooncake/Dockerfile.mooncake
Syspretor: Done, it's good to keep it on the latest version.

examples/mooncake/pd-disaggregated-with-mooncake.yaml
Syspretor: We initially considered using Kubernetes labels for dynamic service discovery; however, for the Mooncake demo, this would introduce some additional complexity—such as requiring an extra ServiceAccount to be created. Therefore, we’ve opted to keep the approach simple in this demo. We plan to provide more comprehensive and flexible Router configurations within RBG in the example/router directory. Would this be acceptable?

PR #128, opened 2025-12-17
add cli: kubectl rbg llm generate

bcfre:

fixes #XXXX

## Checklist


cmd/cli/cmd/llm/generate/generate.go
```diff
@@ -0,0 +1,261 @@
...
+	cmd.Flags().StringVar(&config.ModelName, "model", "", "Model name (required)")
+	cmd.Flags().StringVar(&config.SystemName, "system", "", "GPU system type (required)")
+	cmd.Flags().IntVar(&config.TotalGPUs, "total-gpus", 0, "Total number of GPUs for deployment (required)")
+
+	// Core optional parameters
+	cmd.Flags().StringVar(&config.HuggingFaceID, "hf-id", "", "HuggingFace model ID (e.g., Qwen/Qwen2.5-7B)")
+	cmd.Flags().StringVar(&config.DecodeSystemName, "decode-system", "", "GPU system for decode workers (defaults to --system)")
+	cmd.Flags().StringVar(&config.BackendName, "backend", BackendSGLang, "Inference backend")
+	cmd.Flags().StringVar(&config.BackendVersion, "backend-version", "", "Backend version")
+	cmd.Flags().IntVar(&config.ISL, "isl", 4000, "Input sequence length")
```
Syspretor: Should these flags (ttft.tpot.isl.osl) be required? The default value isn't helpful to users.

bcfre: That sounds great!

cmd/cli/cmd/llm/generate/dependency.go
```diff
@@ -0,0 +1,107 @@
...
+		klog.Warning("Could not determine aiconfigurator version, but tool is available")
+		return nil
+	}
+
+	// Extract version number from output using regex
+	// Expected format: "aiconfigurator 0.5.0" or just "0.5.0"
+	outputStr := string(output)
+	klog.V(2).Infof("aiconfigurator version output: %s", outputStr)
+
+	versionRegex := regexp.MustCompile(`(\d+\.\d+\.\d+)`)
```
cheyang, line 54: How about using  https://github.com/Masterminds/semver ?

cmd/cli/cmd/llm/generate/executor.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	if config.RequestLatency > 0 {
+		args = append(args, "--request_latency", strconv.FormatFloat(config.RequestLatency, 'f', -1, 64))
+	}
+
+	// Add extra arguments
+	for key, value := range config.ExtraArgs {
+		args = append(args, fmt.Sprintf("--%s", key), value)
+	}
```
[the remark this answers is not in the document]
bcfre: This extended field is not currently being used.

cmd/cli/cmd/llm/generate/dependency.go
bcfre: Currently, semver doesn't support version comparisons like 0.5.0rc0 > 0.4.0; it only works with explicitly delimited pre-release identifiers. Because of this limitation, I think we shouldn't use semver.

PR #129, opened 2025-12-18
kep: rbg warmup

TrafalgarZZZ:

fixes #XXXX

## Checklist


keps/XX-rbg-warmup/README.md
```diff
@@ -0,0 +1,209 @@
...
+metadata:
+  name: warmup-demo
+  namespace: default
+spec:
+  policies:
+    parallelism: 10 # how many nodes will be warmed up at the same time
+    backoffLimit: 3 #
+    timeoutSeconds: 300
+    ttlSecondsAfterFinished: 86400
+  images:
```
zmberg: What is the relationship between images, volumes, and targets? Can the overall configuration be simplified? Is retaining only the target rbg configuration insufficient for preheating?

TrafalgarZZZ: Please take a look at the new API design. Thx.

cheyang (commented):
Thanks for putting this KEP together. The warmup concept is well-motivated and the architecture comparison table is helpful. A few things need attention before this is mergeable:

**Metadata issues in `kep.yaml`:**
- `title` is still set to "KEP Template" — should be the actual KEP title.
- `kep-number` is 30, should be 129.
- Author list doesn't include the PR author (`trafalgarz@outlook.com`).

**Empty sections that need content:**
- Test Plan (Unit / Integration / E2E) — at least outline what will be tested.
- Graduation Criteria — what gates alpha -> beta -> GA?
- "Important Notes" under Controller Logic — either fill it or remove the heading.
- Drawbacks — even if minor, acknowledging trade-offs is useful.

**Release Signoff Checklist** still references `kubernetes/enhancements`, `kubernetes/website`, and conformance test links from upstream Kubernetes. These should be adapted (or removed) for the RBG project.

**Design questions:**

1. How is the mutual exclusivity between `targetNodes` and `targetRoleBasedGroup` enforced? CRD-level CEL validation, webhook, or controller-side check?

2. In the YAML example, `restartPolicy` sits at the same level as `spec` inside `podTemplate`. Standard PodTemplateSpec puts it inside `spec`. Which structure do you intend?

3. The commented-out `dataCopy` block — is this deferred to a later phase? Worth a sentence clarifying scope.

4. Failure semantics: when a per-node warmup pod fails, does `backoffLimit` apply per-node or globally? What happens to already-warmed nodes if the overall job times out?

5. The non-goals section says coordination between warmup and upgrade is out-of-scope, but the motivation section explicitly frames warmup as an upgrade accelerator. Could you clarify the boundary? e.g., "warmup is triggered manually; automatic trigger before upgrade is future work."

**Status conditions:** The design only mentions `phase` + counters. Consider adding standard `conditions[]` (e.g., `NodesSelected`, `WarmupInProgress`, `WarmupComplete`) for observability and programmatic consumption.

**Minor:**
- "upgradation" is non-standard English — suggest using "upgrade" instead.
- `status: implementable` in `kep.yaml` seems premature given the empty test plan and graduation criteria.

Overall the core design (WarmUp Pod per node, init containers for image pull, sidecar containers for custom actions) is solid. The alternative analysis is thorough. Once the gaps above are addressed this should be in good shape.

cheyang (commented):
## KEP-129 Review: Image and Volume Warmup for RoleBasedGroup

Thanks @TrafalgarZZZ for this proposal. The motivation is clear and the architecture comparison across three alternatives is well done. I have several comments below, ranging from metadata fixes to design completeness gaps.

---

### Metadata Issues (kep.yaml)

1. **`kep-number` is `30` instead of `129`.** This appears to be a copy-paste artifact from KEP-30. Please update to `kep-number: 129`.
2. **`title` is still `KEP Template`** -- while I notice KEP-30 also has this, it would be clearer to set it to the actual title (`KEP-129: Image and Volume Warmup for RoleBasedGroup`).

### Missing / Empty Sections

The following sections in the KEP are declared in the table of contents but have no content:

| Section | Status |
|---|---|
| **Test Plan** (Unit / Integration / E2E) | Empty -- at least an outline of what will be tested is expected |
| **Graduation Criteria** | Empty -- describe alpha, beta, GA gates |
| **Implementation History** | Empty |
| **Drawbacks** | Empty |
| **Important Notes** | Empty (under Controller Logic) |

Even at the proposal stage, these sections should have at least a rough sketch so reviewers can evaluate feasibility and scope.

### API Design Feedback

3. **`restartPolicy` placement in `customizedAction.podTemplate`:** In the YAML example, `restartPolicy` is placed directly under `podTemplate`, but for a Pod template spec it should be under `podTemplate.spec`. The current structure is:
   ```yaml
   customizedAction:
     podTemplate:
       restartPolicy: Never|OnFailure   # <-- wrong level
       spec:
         containers: ...
   ```
   Should be:
   ```yaml
   customizedAction:
     podTemplate:
       spec:
         restartPolicy: Never|OnFailure
         containers: ...
   ```

4. **`dataCopy` is entirely commented out** with no explanation. Is this deferred to a later phase? Dropped? Please add a note clarifying the status and expected timeline. If it's a non-goal for alpha, listing it explicitly in Non-Goals would help.

5. **`backoffLimit: 3`** -- the description should clarify the retry semantics. Is this per-node retry count or for the entire warmup job? What happens when the backoff limit is reached for a single node (does it mark the node as failed and continue, or does the whole job fail)?

### Design Gaps

6. **Resource consumption:** WarmUp pods will consume node resources (CPU/memory requests). The KEP should discuss how to minimize resource footprint (e.g., minimal resource requests for image-preload init containers) and what happens if a node has insufficient resources to schedule the warmup pod alongside existing workloads.

7. **Cleanup / garbage collection:** The `ttlSecondsAfterFinished` field is mentioned but the KEP doesn't describe the cleanup mechanism. Who cleans up warmup pods -- the controller or Kubernetes TTL? What happens if the controller crashes mid-warmup?

8. **Failure handling details:** The state machine describes a `Failed` terminal state, but doesn't discuss retry behavior for individual node failures or partial success scenarios. If 99/100 nodes succeed, can the user retry only the failed node?

9. **RBAC requirements:** No mention of what RBAC permissions are needed for the new `RoleBasedGroupWarmup` CRD, or what service account the warmup pods run under.

10. **Coordination with RBG upgrades:** The Non-Goals section says co-design with RBG upgrades is out of scope, but the Motivation section explicitly argues that warmup improves upgrade efficiency. The KEP should at least describe the intended manual workflow (user creates WarmUp CR, waits for completion, then triggers upgrade).

### Minor

11. **Typo:** "upgradation" (used in Motivation and Non-Goals) should be "upgrade process" or "upgrade". "Upgradation" is not standard English.

12. **Alternatives table clarity:** The table shows Volume WarmUp (volume changed) as supported only by WarmUp Pod but doesn't explain *why* the other two approaches can't handle changed volumes. A brief note would strengthen the comparison.

---

**Overall:** The problem statement is well-motivated and the architecture choice is justified with a good comparison table. However, several sections critical for KEP completeness are empty, and the API spec needs corrections. I'd suggest filling in the Test Plan, Graduation Criteria, and addressing the design gaps above before marking as implementable.

cheyang (approved):
/lgtm
/approve

PR #148, opened 2026-01-21
[WIP] add EPD

liusy58:

fixes #XXXX

## Checklist

### DIFF
--- a/keps/75-EPD/README.md
+++ b/keps/75-EPD/README.md
@@ -1,0 +1,158 @@
+# KEP-75 SGLang EPD Integration
+
+<!--
+This KEP introduces the integration of Encoder-Prefill-Decode (EPD) Disaggregation 
+within the Role-Based Group (RBG) framework for SGLang.
+-->
+
+<!-- toc -->
+- [Motivation](#motivation)
+- [Proposal](#proposal)
+    - [User Stories](#user-stories)
+    - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+    - [EPD Architecture](#epd-architecture)
+    - [Role Definitions](#role-definitions)
+    - [RBG Deployment Example](#rbg-deployment-example)
+- [Benchmark](#benchmark)
+- [Test Plan](#test-plan)
+    - [Integration Tests](#integration-tests)
+    - [End to End Tests](#end-to-end-tests)
+<!-- /toc -->
+
+## Motivation
+
+Vision-Language Models (VLMs), such as Qwen2.5-VL and Llama-Vision, introduce unique computational challenges that standard collocated inference architectures struggle to handle efficiently:
+
+1.  **ViT Scaling Inefficiency**: Vision Transformers (ViT) do not scale linearly with Tensor Parallelism (TP). Increasing TP for ViT often degrades performance due to communication overhead.
+2.  **Resource Imbalance**: Vision processing (encoding) is compute-intensive but only occurs during the prefill phase. In requests with multiple images, the vision encoder becomes a significant bottleneck for Time To First Token (TTFT).
+3.  **Static Resource Allocation**: In traditional deployments, vision and language components share the same GPU resources, preventing independent scaling based on workload characteristics.
+
+This KEP aims to integrate **Encoder-Prefill-Decode (EPD) Disaggregation** as a core pattern in RBG-deployed SGLang services. By separating vision encoding into a dedicated role, users can scale encoders horizontally to handle image-heavy workloads, significantly improving service SLOs.
+
+## Proposal
+
+The proposal introduces a three-tier disaggregated architecture:
+- **Encoder**: Independent nodes for ViT processing.
+- **Prefill**: Language nodes that retrieve embeddings from the Encoders.
+- **Decode**: Dedicated nodes for token generation.
+
+This enables independent scaling of the vision encoding capacity without altering the language model's configuration.
+
+### User Stories
+
+#### Story 1
+As a multi-modal service provider, I want to handle requests containing 8+ images. By deploying multiple `vlm-encoder` replicas in an RBG, I want the system to parallelize image encoding across these nodes, reducing the TTFT from seconds to milliseconds.
+
+### Risks and Mitigations
+
+-   **Network Latency**: Moving large vision embedding tensors between nodes adds overhead.
+    -   *Mitigation*: Support high-performance transfer backends like Mooncake (GPU-Direct RDMA) or ZMQ to ensure transfer time is significantly lower than the compute time saved.
+-   **Resource Utilization**: Dedicated encoders may be underutilized during text-only requests.
+    -   *Mitigation*: Use RBG's horizontal scaling capabilities to adjust encoder counts based on real-time multimodal traffic.
+
+## Design Details
+
+### EPD Architecture
+
+The EPD workflow follows a specific request flow:
+1.  **Client Request**: Arrives at the `sglang-router`.
+2.  **Image Distribution**: The `vlm-prefill` node splits image inputs and distributes them to the `vlm-encoder` pool.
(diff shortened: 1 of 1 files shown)


keps/75-EPD/README.md
```diff
@@ -0,0 +1,158 @@
...
+    - name: vlm-encoder
+      replicas: 2
+      template:
+        spec:
+          containers:
+            - name: sglang
+              image: lmsysorg/sglang:dev
+              command:
+                - python3
+                - -m sglang.launch_server
```
whybeyoung, line 94: need test?  maybe sperate it with newline.
```
- python3
- -m
- sglang.launch_server
....
```

keps/75-EPD/README.md
```diff
@@ -0,0 +1,158 @@
...
+                - --model-path /models/Qwen2.5-VL-7B-Instruct
```
whybeyoung, line 95: the same with above

keps/75-EPD/README.md
```diff
@@ -0,0 +1,158 @@
...
+                - --encoder-only
+                - --enable-prefix-mm-cache
+                - --port 30002
```
whybeyoung, line 98: the same with above

keps/75-EPD/README.md
```diff
@@ -0,0 +1,158 @@
...
+      template:
+        spec:
+          containers:
+            - name: sglang
+              image: lmsysorg/sglang:dev
+              env:
+                - name: ENCODER_URLS
+                  value: "http://s-sglang-vlm-epd-vlm-encoder-0:30002 http://s-sglang-vlm-epd-vlm-encoder-1:30002"
+              command:
+                - sh
```
whybeyoung, line 115: It is not very reasonable for sh to act as the No.1 process of a container.

keps/75-EPD/README.md
```diff
@@ -0,0 +1,158 @@
...
+    - name: vlm-prefill
+      dependencies: [ "vlm-encoder" ]
+      replicas: 1
+      template:
+        spec:
+          containers:
+            - name: sglang
+              image: lmsysorg/sglang:dev
+              env:
+                - name: ENCODER_URLS
```
whybeyoung, line 112: For  ENCODER_URLS, It seems unreasonable to configure a static Encoder address in prefill—this won't be the way it's used in production, right? Perhaps we should implement its service discovery automatically in gateway mode instead?

keps/75-EPD/README.md
```diff
@@ -0,0 +1,158 @@
...
+    - name: vlm-decode
+      replicas: 1
+      template:
+        spec:
+          containers:
+            - name: sglang
+              image: lmsysorg/sglang:dev
+              command:
+                - python3
+                - -m sglang.launch_server
```
whybeyoung, line 136: seperate it.

keps/75-EPD/README.md
```diff
@@ -0,0 +1,158 @@
...
+                - --model-path /models/Qwen2.5-VL-7B-Instruct
+                - --disaggregation-mode decode
+                - --port 30001
+              resources:
+                limits:
+                  nvidia.com/gpu: "1"
+
+    - name: sglang-router
```
whybeyoung, line 144: rename to sgl-gateway

keps/75-EPD/README.md
```diff
@@ -0,0 +1,158 @@
...
+      dependencies: [ "vlm-prefill", "vlm-decode" ]
+      replicas: 1
+      template:
+        spec:
+          containers:
+            - name: router
+              image: lmsysorg/sglang:dev
+              command:
+                - python3
+                - -m sglang_router.launch_router
```
whybeyoung, line 154: seperate it

whybeyoung (changes requested):
Overall, it needs to be verified successfully in the local Kubernetes environment. In addition, it is recommended that the EPD feature support service discovery in the gateway.

cheyang (commented):
Hey @liusy58 - checking in on this KEP. It's been about 5 months since @whybeyoung's review. Are you still planning to continue this work?

A few things that need attention before this can move forward:

**Incomplete sections:**
- The `Benchmark` section referenced in the TOC is empty. Even for a KEP, expected/target numbers would strengthen the proposal.
- `Test Plan` (both Integration Tests and End-to-End Tests) is blank. At minimum, outline what you intend to validate.

**Previous review feedback unaddressed:**
- whybeyoung asked for local Kubernetes verification and service discovery support in the gateway. No follow-up commits since then.

**YAML manifest issue in the encoder role:**
```yaml
command:
  - python3
  - -m sglang.launch_server   # <-- single arg with space
  - --encoder-only
```
In a Kubernetes pod spec, each list item is a separate exec argument. `-m sglang.launch_server` as one string won't work - Python expects `-m` and `sglang.launch_server` as two separate args. Either split them or wrap in `sh -c` like you did for the prefill role.

**Minor:**
- The decode role is missing `--language-only` (or equivalent) in the command - is that intentional? The table says decode uses `--disaggregation-mode decode` but the YAML doesn't include it either.

Let us know if you'd like to pick this back up or if priorities have shifted. Happy to re-review once updated.

keps/75-EPD/README.md
```diff
@@ -0,0 +1,158 @@
...
+                - --pd-disaggregation
+                - --prefill http://s-sglang-vlm-epd-vlm-prefill:30000
+                - --decode http://s-sglang-vlm-epd-vlm-decode:30001
+                - --port 8000
```
cheyang, line 158: The table of contents references **Benchmark**, **Integration Tests**, and **End-to-End Tests** sections, but the document body ends after the RBG deployment YAML (line 158) without any of those sections present. Even as a WIP, having stub headings keeps the structure navigable and signals what's still planned.

keps/75-EPD/README.md
```diff
@@ -0,0 +1,158 @@
...
+apiVersion: workloads.x-k8s.io/v1alpha1
+kind: RoleBasedGroup
+metadata:
+  name: sglang-vlm-epd
+spec:
+  roles:
+    - name: vlm-encoder
+      replicas: 2
+      template:
+        spec:
```
cheyang, line 88: The `vlm-encoder`, `vlm-decode`, and `sglang-router` roles use list-style `command` entries like `- -m sglang.launch_server` and `- --prefill http://...`. In a Kubernetes pod spec each list item becomes a separate exec argument, so Python would receive `"-m sglang.launch_server"` as one token and fail to recognize the `-m` flag. The `vlm-prefill` role already handles this correctly with `sh -c "..."`. Consider applying the same `sh -c` pattern to all roles, or splitting flags and values into individual list items (e.g., `- -m` / `- sglang.launch_server`).

keps/75-EPD/README.md
```diff
@@ -0,0 +1,158 @@
...
+| **Decode** | `vlm-decode` | `--disaggregation-mode decode` | Dedicated to auto-regressive token generation. |
+| **Router** | `sglang-router` | N/A | Entry point for load balancing and PD coordination. |
+
+### RBG Deployment Example
+
+The following RBG manifest deploys a Qwen2.5-VL-7B EPD cluster with 2 Encoders, 1 Prefiller, and 1 Decoder.
+
+```yaml
+apiVersion: workloads.x-k8s.io/v1alpha1
+kind: RoleBasedGroup
```
cheyang, line 80: Compared to KEP-74 (Mooncake Integration), this manifest omits readinessProbes and model volume definitions. Adding at least a TCP socket readinessProbe and a PVC volume mount for the model path would make the example deployment-ready and consistent with existing KEPs.

keps/75-EPD/README.md
cheyang, line 158: The KEP template and sibling KEPs include an **Alternatives** section explaining why other approaches were ruled out. For EPD this could briefly cover why a two-tier PD split (without separate encoders) was insufficient, strengthening the motivation.

PR #164, opened 2026-02-26
refactor(rbg): implement v1alpha2 apiversion

Syspretor:

fixes #XXXX

## Checklist


api/workloads/v1alpha2/groupversion_info.go
```diff
@@ -0,0 +1,36 @@
+/*
+Copyright 2025.
```
cheyang, line 2: 2025 -》2026

pkg/utils/revision_utils.go
```diff
@@ -115,7 +116,7 @@ func ApplyRevision(
 
 func CleanExpiredRevision(
 	ctx context.Context, client client.Client,
-	rbg *workloadsv1alpha1.RoleBasedGroup) ([]*appsv1.ControllerRevision, error) {
+	rbg *workloadsv1alpha2.RoleBasedGroup) ([]*appsv1.ControllerRevision, error) {
 	selector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{
 		MatchLabels: map[string]string{
 			workloadsv1alpha1.SetNameLabelKey: rbg.Name,
```
cheyang, line 122: Since SetNameLabelKey is a version-independent constant shared across v1alpha1 and v1alpha2, consider extracting it into a shared package (e.g., a top-level common package) rather than referencing it from a specific API version. 

This avoids confusion about which version to import and prevents inconsistencies like the mix of workloadsv1alpha1.SetNameLabelKey and workloadsv1alpha2.SetNameLabelKey seen in this PR.

PR #185, opened 2026-03-09
fix(role-instance): restart policy does not take effect

### LINKED ISSUE #165: [Bug] instance restart policy not working
### Checklist

- [ ] 1. I have searched related issues but cannot get the expected help.
- [ ] 2. The bug has not been fixed in the latest version.
- [ ] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [ ] 5. Please use English, otherwise it will be closed.

### Describe the bug

When use instance workload, after setting `RecreateInstanceOnPodRestart` to the instance, controller would not recreate the failed instance once one pod is failed or deleted.

The reason why is the `RestartPolicy` not cited in instance controller, therefore, the instance controller can not recognize this policy.

https://github.com/sgl-project/rbg/blob/main/api/workloads/v1alpha1/instance_types.go#L39

### Reproduction

Construct a rbg with instance workload, and set the size of the instance to more than one. And try to delete one pod, the other pods in the same instance will not restart nor be recreated, which is unexpected.

### Environment

latest version, v0.6.0

Syspretor:

fixes #165 

## Checklist

### DIFF
--- a/pkg/reconciler/instance/sync/instance_scale.go
+++ b/pkg/reconciler/instance/sync/instance_scale.go
@@ -200,10 +200,20 @@ func shouldRecreateInstance(instance *v1alpha1.Instance, pods []*v1.Pod) bool {
 
 	// Check if any pod has been deleted:
 	// Only trigger recreate if Instance was previously Ready (stable state)
-	// This avoids triggering recreate during initial creation or scaling
-	if wasInstanceReady(instance) {
+	// and spec is not being changed (Generation == ObservedGeneration).
+	// This avoids triggering recreate during initial creation or scaling up.
+	if wasInstanceReady(instance) && instance.Generation == instance.Status.ObservedGeneration {
 		expectedPodCount := getExpectedPodCount(instance)
-		if len(pods) < expectedPodCount {
+		// Calculate active pods (exclude those being deleted).
+		// Pods with DeletionTimestamp != nil are in Terminating state and should not
+		// be counted as active, otherwise the controller may fail to detect pod deletion.
+		activeCount := 0
+		for _, p := range pods {
+			if p.DeletionTimestamp == nil {
+				activeCount++
+			}
+		}
+		if activeCount < expectedPodCount {
 			return true
 		}
 	}


pkg/reconciler/instance/sync/instance_scale.go
```diff
@@ -163,3 +172,62 @@ func (c *realControl) createOnePod(ctx context.Context, instance *v1alpha1.Insta
...
+			return true
+		}
+	}
+
+	// Check if any pod has been deleted:
+	// Only trigger recreate if Instance was previously Ready (stable state)
+	// This avoids triggering recreate during initial creation or scaling
+	if wasInstanceReady(instance) {
+		expectedPodCount := getExpectedPodCount(instance)
+		if len(pods) < expectedPodCount {
```
cheyang: When a user scales up the Instance (increasing component sizes), the Instance may still be in Ready state (progressive availability), but len(pods) < expectedPodCount will evaluate to true, causing unnecessary deletion of existing pods. Can we also check `if instance.Generation == instance.Status.ObservedGeneration`  to make it safe?

```golang
// Only check for missing pods when spec is not being changed
if wasInstanceReady(instance) && instance.Generation == instance.Status.ObservedGeneration {
    expectedPodCount := getExpectedPodCount(instance)
    if len(pods) < expectedPodCount {
        return true
    }
}
```

pkg/reconciler/instance/sync/instance_scale.go
```diff
@@ -163,3 +172,62 @@ func (c *realControl) createOnePod(ctx context.Context, instance *v1alpha1.Insta
...
+
+	// Check each pod for container restart
+	for _, pod := range pods {
+		if utils.ContainerRestarted(pod) {
+			return true
+		}
+	}
+
+	// Check if any pod has been deleted:
+	// Only trigger recreate if Instance was previously Ready (stable state)
```
cheyang, line 202: We should also consider the pod is terminating for a long time. The current check len(pods) < expectedPodCount includes pods where DeletionTimestamp != nil. If a pod is evicted or manually deleted, it remains in the pods list during termination. The controller will see Count == Expected and fail to trigger recreation, causing the job to hang indefinitely. 

```golang
// Calculate active pods (exclude those being deleted)
activeCount := 0
for _, p := range pods {
    if p.DeletionTimestamp == nil {
        activeCount++
    }
}
if activeCount < expectedPodCount {
    return true
}
```

PR #188, opened 2026-03-10
chore(rbg): implement v1alpha2 as storage version

Syspretor:

fixes #XXXX

## Checklist

### DIFF
--- a/internal/controller/workloads/discovery_config_mode_test.go
+++ b/internal/controller/workloads/discovery_config_mode_test.go
@@ -19,7 +19,109 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
-func TestReconcileDiscoveryConfigMap(t *testing.T) {
+func TestEnsureDiscoveryConfigMode(t *testing.T) {
+	type testCase struct {
+		name               string
+		mutateRBG          func(*workloadsv1alpha2.RoleBasedGroup)
+		buildExtraObjects  func(*workloadsv1alpha2.RoleBasedGroup) []runtime.Object
+		wantRequeue        bool
+		wantMode           constants.DiscoveryConfigMode
+		wantModeAnnotation bool
+	}
+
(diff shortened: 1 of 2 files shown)


internal/controller/workloads/rolebasedgroup_controller.go
```diff
@@ -970,13 +976,13 @@ func (r *RoleBasedGroupReconciler) CalculateScalingForAllCoordination(
 // getScheduledReplicas queries the number of scheduled pods (with nodeName) for a given role.
 func (r *RoleBasedGroupReconciler) getScheduledReplicas(
 	ctx context.Context,
-	rbg *workloadsv1alpha1.RoleBasedGroup,
-	roleName string,
+	rbg *workloadsv1alpha2.RoleBasedGroup,
+	rollName string,
```
cheyang: rollName -> roleName

pkg/reconciler/pod_reconciler.go
```diff
@@ -123,7 +124,7 @@ func (r *PodReconciler) ConstructPodTemplateSpecApplyConfiguration(
 		return nil, err
 	}
 
-	scheduler.InjectPodGroupProtocol(rbg, podTemplateApplyConfiguration)
+	scheduler.InjectPodGroupProtocolV2(rbg, podTemplateApplyConfiguration)
```
cheyang: PodGroup reconciliation step is removed from RoleBasedGroupReconciler.Reconcile() (the old Step 2), but InjectPodGroupProtocolV2() is still being called in pod_reconciler.go. Since IsKubeGangSchedulingV2() and IsVolcanoGangSchedulingV2() currently always return false, this creates a half-removed feature state. I suggest removing the InjectPodGroupProtocolV2 call and related scheduler code if gang scheduling is intentionally redesigned to a later version.

PR #195, opened 2026-03-13
feature(rbg): support gang scheduling in v1alpha2

Syspretor:

fixes #XXXX

## Checklist

### DIFF
--- a/pkg/scheduler/podgroup_manager_test.go
+++ b/pkg/scheduler/podgroup_manager_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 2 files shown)


pkg/scheduler/k8s-scheduler-plugin/manager.go
```diff
@@ -0,0 +1,191 @@
...
+
+	if apierrors.IsNotFound(err) {
+		if createErr := m.client.Create(ctx, podGroup); createErr != nil {
+			logger.Error(createErr, "create pod group error")
+			return createErr
+		}
+		return nil
+	}
+
+	if podGroup.Spec.MinMember != int32(rbg.GetGroupSize()) {
```
cheyang: if the user changes the rbg.workloads.x-k8s.io/group-gang-scheduling-timeout annotation,
 the existing PodGroup will never be updated to reflect the new timeout value.

 Is it by design？

cmd/rbgs/main.go
```diff
@@ -135,6 +138,11 @@ func main() {
 		"The number of worker threads used by the the RBGS controller.",
 	)
 	flag.DurationVar(&cacheSyncTimeout, "cache-sync-timeout", 120*time.Second, "Informer cache sync timeout.")
+	flag.StringVar(
```
cheyang, line 141: Please validate --scheduler-name explicitly instead of silently falling back to scheduler-plugins.

 Right now cmd/rbgs/main.go passes the raw flag value into scheduler.SchedulerPluginType(. ..), and NewPodGroupManager() defaults to the kube scheduler-plugins implementation for any unknown
 value. That means a typo like volcanoo will not fail fast; it will silently select the wrong backend.

 This is a risky misconfiguration path because users may think they deployed Volcano support while the controller is actually reconciling scheduler-plugins PodGroups.

cheyang (approved):
/lgtm
/approve

PR #206, opened 2026-03-16
refactor(rbgs): change spec api field

Syspretor:

fixes #XXXX

## Checklist

### DIFF
--- a/test/wrappers/v1alpha1/rbgs_wrapper.go
+++ b/test/wrappers/v1alpha1/rbgs_wrapper.go
@@ -43,9 +43,7 @@ func BuildBasicRoleBasedGroupSet(name, ns string) *RoleBasedGroupSetWrapper {
 			},
 			Spec: workloadsv1alpha.RoleBasedGroupSetSpec{
 				Replicas: ptr.To(int32(1)),
-				Template: workloadsv1alpha.RoleBasedGroupTemplateSpec{
-					Spec: BuildBasicRoleBasedGroup(name, ns).Obj().Spec,
-				},
+				Template: BuildBasicRoleBasedGroup(name, ns).Obj().Spec,
 			},
 		},
 	}


internal/controller/workloads/rolebasedgroupset_controller.go
```diff
@@ -482,7 +482,7 @@ func newRBGForSet(rbgset *workloadsv1alpha2.RoleBasedGroupSet, index int) *workl
 			// The OwnerReference will be set in the scaleUp function.
 		},
 		Spec: workloadsv1alpha2.RoleBasedGroupSpec{
```
cheyang, line 536: groupTemplate.metadata is introduced in the API, but the controller does not currently use it when creating or updating child RBGs.

 RoleBasedGroupTemplateSpec now contains both metadata and spec, which suggests users can define metadata for generated child RoleBasedGroup objects. However, the controller still only
 copies:
 - generated name / namespace
 - controller-managed labels
 - groupTemplate.spec.roles
 - one propagated annotation from the parent set

 So at the moment, fields like:
 - spec.groupTemplate.metadata.labels
 - spec.groupTemplate.metadata.annotations

 appear to be supported by the API, but are ignored by reconciliation.

 I think this is a correctness issue because the schema and the controller behavior are now out of sync.

internal/controller/workloads/rolebasedgroupset_controller.go
```diff
@@ -375,7 +375,7 @@ func (r *RoleBasedGroupSetReconciler) needsUpdate(
 	rbgset *workloadsv1alpha2.RoleBasedGroupSet, rbg *workloadsv1alpha2.RoleBasedGroup,
 ) bool {
 	// Check if the template spec has changed using order-insensitive comparison
```
cheyang, line 377: Even after metadata application is added, the update path is currently incomplete: needsUpdate() only compares roles plus a small subset of propagated annotations from the parent
 RoleBasedGroupSet.

 That means a change like:

 ```yaml
   spec:
     groupTemplate:
       metadata:
       labels:

 ```

 would not trigger updates for existing child RBGs.

 Please make sure the update detection logic and the create/update application logic cover the same groupTemplate fields.

internal/controller/workloads/rolebasedgroupset_controller_test.go
```diff
@@ -250,9 +252,11 @@ func TestRoleBasedGroupSetReconciler_needsUpdate(t *testing.T) {
 					Namespace: "default",
 				},
 				Spec: workloadsv1alpha2.RoleBasedGroupSetSpec{
-					Template: workloadsv1alpha2.RoleBasedGroupSpec{
-						Roles: []workloadsv1alpha2.RoleSpec{
-							{Name: "new-role"},
+					GroupTemplate: workloadsv1alpha2.RoleBasedGroupTemplateSpec{
```
cheyang, line 255: The tests currently cover the field rename, but not the new metadata semantics introduced by this PR.

cheyang (approved):
/lgtm
/approve

PR #225, opened 2026-03-19
v1alpha2 conversion webhook

diw-zw:

fixes #XXXX

## Checklist


deploy/helm/rbgs/templates/clusterrole.yaml
```diff
@@ -27,13 +27,24 @@ rules:
...
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - create
+  - get
+  - list
+  - update
+  - watch
 - apiGroups:
```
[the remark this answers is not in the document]
diw-zw: fixed

Syspretor (approved):
/lgtm

internal/controller/workloads/coordinatedpolicy_migration_controller.go
```diff
@@ -0,0 +1,156 @@
...
+// buildCoordinatedPolicyApplyConfiguration converts v1alpha1 CoordinationRequirements
+// into a CoordinatedPolicy apply configuration owned by the given RoleBasedGroup.
+func buildCoordinatedPolicyApplyConfiguration(
+	rbg *workloadsv1alpha2.RoleBasedGroup,
+	coords []workloadsv1alpha1.Coordination,
+) *applyv1alpha2.CoordinatedPolicyApplyConfiguration {
+	rules := make([]*applyv1alpha2.CoordinatedPolicyRuleApplyConfiguration, 0, len(coords))
+	for _, c := range coords {
+		rule := applyv1alpha2.CoordinatedPolicyRule().
+			WithRoles(c.Roles...)
```
cheyang, line 94: `CoordinatedPolicySpec.policies` is declared as a map list keyed by `name`, and `CoordinatedPolicyRule.name` is also required in the CRD schema. This migration path builds each rule with `WithRoles(...)` and optional `WithStrategy(...)`, but never sets `WithName(...)`. In a real cluster that means the SSA payload for the generated `CoordinatedPolicy` does not satisfy the schema / list-map key contract, so applying it can fail for any v1alpha1 object that still carries coordination data. Please populate a stable rule name when translating each v1alpha1 coordination entry.

config/default/kustomization.yaml
```diff
@@ -47,6 +47,39 @@ resources:
...
+    kind: Service
+    version: v1
+    name: webhook-service   # Kustomize will resolve this to rbgs-webhook-service at render time
+    fieldPath: metadata.name
+  targets:
+  - select:
+      kind: Deployment
+      name: rbgs-controller-manager
+    fieldPaths:
+    - spec.template.spec.containers.[name=rbgs].args.3
```
cheyang: These replacements write to `spec.template.spec.containers.[name=rbgs].args.3` and `.args.4`, but `config/manager/manager.yaml` currently defines only three args entries (`0..2`). That leaves the default `config/default` render path inconsistent and is likely to break `kustomize build` for the standard install path. Either add the corresponding args to the Deployment first, or remove/adjust these replacements so they only target fields that actually exist.

Syspretor (approved):

PR #231, opened 2026-03-23
Chore(rbg): add v1alpha2 related examples

Syspretor:

fixes #XXXX

## Checklist

### DIFF
--- a/examples/inference/agg-leader-worker.yaml
+++ b/examples/inference/agg-leader-worker.yaml
@@ -27,24 +27,23 @@ apiVersion: workloads.x-k8s.io/v1alpha2
 kind: RoleBasedGroup
 metadata:
   name: agg-lwd
-  namespace: default
   labels:
     app: llm-inference-tp
     deployment-type: aggregated-tensor-parallel
     model: qwen-0.6b
 spec:
   roles:
-    # Backend: 2 instances, tensor parallel size = 4 (1 leader + 3 workers)
+    # Backend: 2 instances, tensor parallel size = 2 (1 leader + 2 workers)
     - name: backend
-      replicas: 2
-      restartPolicy: RecreateRoleInstanceOnPodRestart
+      replicas: 1
+      restartPolicy: None #RecreateRoleInstanceOnPodRestart
       rolloutStrategy:
         type: RollingUpdate
         rollingUpdate:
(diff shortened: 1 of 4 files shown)


doc/features/exclusive-topology.md
```diff
@@ -36,7 +36,7 @@ In this example, all pods (leader, worker, and lws) will be scheduled on the sam
 ## Example: RoleBasedGroupSet with Exclusive Topology
 
 RoleBasedGroupSet extends the exclusive topology feature to multiple groups, where each group is scheduled exclusively
-on a topology domain. See example [rbgs-exclusive-topology.yaml](../../examples/rbgs/exclusive-topology.yaml).
+on a topology domain. See example [rbgs-exclusive-topology.yaml](../../examples/workload/rbgs/exclusive-topology.yaml).
```
cheyang: `doc/features/exclusive-topology.md` now links to
 `../../examples/workload/rbgs/exclusive-topology.yaml`, but I could not find
 that path in the repo.

examples/inference/agg-standalone.yaml
```diff
@@ -0,0 +1,96 @@
...
+                  - python3
+                  - -m
+                  - sglang.launch_server
+                  - --model-path
+                  - "Qwen/Qwen3-0.6B"
+                  - --host
+                  - "0.0.0.0"
+                  - --port
+                  - "8001"
+                  - --tp-size
```
cheyang, line 69: The command ends with  but does not provide a value after it. That makes the backend startup command incomplete, so this example is not runnable as committed. Could you add the intended value here, or remove the flag if it is not required?

examples/inference/agg-leader-worker.yaml
```diff
@@ -0,0 +1,89 @@
...
+        size: 4   # 1 leader + 3 workers per instance
+        template:
+          metadata:
+            labels:
+              app: llm-inference-tp
+          spec:
+            containers:
+              - name: sglang
+                image: lmsysorg/sglang:v0.5.9-cu124
+                spec:
```
cheyang: The container structure looks malformed here. Inside the  item, there is another nested  block, so  ends up attached at the wrong level. I think , , and  should be defined directly on the same container item instead of under a second nested container list.

Issue #232, opened 2026-03-23
[Feature] set pod group annotations copied from rbg


JasonHe-WQ:
### Checklist

- [ ] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [ ] 2. Please use English, otherwise it will be closed.

### Motivation

Align with LWS controller behaviour, which copy annotations with `volcano.sh/` prefix from `lws` to `pod group`

### Related resources

`LWS` controller inherits annotations from `lws` resource.  However, `rbg` controller would not copy.

https://github.com/kubernetes-sigs/lws/blob/aa903eeb203252c92ee468f3bf335e0732b1bcb1/pkg/schedulerprovider/volcano_provider.go#L59-L68

https://github.com/kubernetes-sigs/lws/blob/aa903eeb203252c92ee468f3bf335e0732b1bcb1/pkg/schedulerprovider/volcano_provider.go#L111-L119

As you can observe, no annotation would be added.

https://github.com/sgl-project/rbg/blob/bd6296e97eeecc45d2772a74bd4d6ec990282076/pkg/scheduler/volcano/manager.go#L102-L115

PR #233, opened 2026-03-23
feat(scheduler): inherit podGroup annotations

JasonHe-WQ:
### Ⅰ. Motivation
align with `lws` controllers' behaviour

### Ⅱ. Modifications
coped annotations with `volcano` prefix from `rbg` to `pod group`

impl #232 

~~For some annotations, it may be covered by volcano scheduler when scheduled, e.g. `volcano.sh/job-allocated-hypernode`.~~

~~Therefore, we may need to use another map to store these annotations.~~

Or we can simplify the impl to only support copy from `rbg` when create and ignore update.

We don't update annos on `podgroup` when update `rbg` since some update may break scheduling policy and hard to investigate

## Checklist

### DIFF
--- a/pkg/scheduler/podgroup_manager_test.go
+++ b/pkg/scheduler/podgroup_manager_test.go
@@ -26,6 +26,11 @@ import (
 	volcanoschedulingv1beta1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
 )
 
+const (
+	rbgName      = "test-rbg"
+	rbgNamespace = "default"
+)
+
 func TestPodGroupScheduler_Reconcile(t *testing.T) {
 	// Define test scheme
 	scheme := runtime.NewScheme()
@@ -35,8 +40,6 @@ func TestPodGroupScheduler_Reconcile(t *testing.T) {
 	_ = apiextensionsv1.AddToScheme(scheme)
 
 	gvk := utils.GetRbgGVK()
-	rbgName := "test-rbg"     //nolint:goconst
-	rbgNamespace := "default" //nolint:goconst
 	podGroup := &schedv1alpha1.PodGroup{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      rbgName,
@@ -262,8 +265,6 @@ func TestVolcanoPodGroupScheduler_ReconcileCopiesVolcanoAnnotationsOnCreate(t *t
 	_ = volcanoschedulingv1beta1.AddToScheme(scheme)
 	_ = apiextensionsv1.AddToScheme(scheme)
 
-	rbgName := "test-rbg"
-	rbgNamespace := "default"
 	rbg := wrappersv2.BuildBasicRoleBasedGroup(rbgName, rbgNamespace).
 		WithAnnotations(
 			map[string]string{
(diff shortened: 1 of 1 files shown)


pkg/scheduler/volcano/manager.go
```diff
@@ -106,6 +111,7 @@ func (m *PodGroupManager) createOrUpdate(ctx context.Context, rbg *workloadsv1al
 			OwnerReferences: []metav1.OwnerReference{
 				*metav1.NewControllerRef(rbg, utils.GetRbgGVK()),
 			},
+			Annotations: desiredAnnotations,
 		},
```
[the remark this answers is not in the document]
JasonHe-WQ: These anno may need further discussion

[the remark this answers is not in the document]
JasonHe-WQ: no need to consider when update

pkg/scheduler/podgroup_manager_test.go
```diff
@@ -271,3 +255,61 @@ func TestPodGroupScheduler_Reconcile(t *testing.T) {
...
+
+func TestVolcanoPodGroupScheduler_ReconcileCopiesVolcanoAnnotationsOnCreate(t *testing.T) {
+	scheme := runtime.NewScheme()
+	_ = workloadsv1alpha2.AddToScheme(scheme)
+	_ = volcanoschedulingv1beta1.AddToScheme(scheme)
+	_ = apiextensionsv1.AddToScheme(scheme)
+
+	rbgName := "test-rbg"
+	rbgNamespace := "default"
+	rbg := wrappersv2.BuildBasicRoleBasedGroup(rbgName, rbgNamespace).
```
cheyang, line 268: We need also test both create and update scenario.

pkg/scheduler/volcano/manager.go
cheyang, line 137: desiredAnnotations is only applied on create, not on update. In the current update path we only reconcile MinMember, Queue, and PriorityClassName, so changes to inherited scheduler
 annotations on the RBG will never be propagated to an existing PodGroup. This also means removed annotations can stay stale. I think annotation inheritance needs to be reconciled on update
 as well, not just at creation time.

pkg/scheduler/volcano/manager.go
```diff
@@ -177,3 +180,16 @@ func (m *PodGroupManager) deletePodGroup(
 
 	return nil
 }
+
+func inheritVolcanoAnnotations(rbg *workloadsv1alpha2.RoleBasedGroup) map[string]string {
+	res := map[string]string{}
+	for k, v := range rbg.Annotations {
```
cheyang: This feels a bit too Volcano-specific. Instead of hard-coding only volcano.sh/*, could we make this generic over a built-in prefix list, e.g. `[]string{"volcano.sh/",
 "scheduling.x-k8s.io/"}`? Then the logic becomes “inherit/reconcile annotations matching supported scheduler prefixes” rather than a Volcano-only special case.



 Reference implementation

 Something like:

 ```go
   supportedSchedulerAnnotationPref ixes := []string{
   "volcano.sh/",
   "scheduling.x-k8s.io/",
   }
 ```

 Then:
 - extract annotations from rbg.Annotations whose keys match any supported prefix
 - on reconcile, remove existing annotations on the PodGroup that match those prefixes
 - re-apply the desired ones from the RBG
 - preserve unrelated annotations

pkg/scheduler/volcano/manager.go
JasonHe-WQ: Yes, but some annotations on pg would be updated once scheduled by the volcano scheduler can should not be covered by other controllers, e.g. `volcano.sh/job-allocated-hypernode`, and the others would be OK to update effectively. Therefore, the imply would be more complex.

pkg/scheduler/volcano/manager.go
JasonHe-WQ: May be instead of combined slice, each scheduler hosts its own annotation key is another choice?

PR #241, opened 2026-03-26
fix(rbg-controller): watch owned RoleBasedGroupScalingAdapter resources

sebest:
## Summary
- Add `.Owns(&RoleBasedGroupScalingAdapter{}, builder.MatchEveryOwner)` to the RBG controller's `SetupWithManager`
- Without this, if an RBGSA is manually deleted, the RBG controller won't notice until its next unrelated reconcile
- Uses `MatchEveryOwner` because RBGSA owner references don't set `Controller: true`, so the default `.Owns()` would silently never trigger

## Test plan
- [x] Extended "Create scaling adapter when not exists" test to verify owner reference correctness (name + UID)
- [x] All existing tests pass

### DIFF
--- a/internal/controller/workloads/rolebasedgroup_controller.go
+++ b/internal/controller/workloads/rolebasedgroup_controller.go
@@ -886,7 +886,7 @@ func (r *RoleBasedGroupReconciler) SetupWithManager(mgr ctrl.Manager, options co
 		Owns(&appsv1.Deployment{}, builder.WithPredicates(WorkloadPredicate())).
 		Owns(&workloadsv1alpha2.RoleInstanceSet{}, builder.WithPredicates(WorkloadPredicate())).
 		Owns(&corev1.Service{}).
-		Owns(&workloadsv1alpha2.RoleBasedGroupScalingAdapter{}, builder.MatchEveryOwner).
+		Owns(&workloadsv1alpha2.RoleBasedGroupScalingAdapter{}, builder.MatchEveryOwner, builder.WithPredicates(RBGScalingAdapterPredicate())).
 		Named("workloads-rolebasedgroup")
 
 	err := utils.CheckCrdExists(r.apiReader, utils.LwsCrdName)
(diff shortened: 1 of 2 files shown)


internal/controller/workloads/rolebasedgroup_controller.go
```diff
@@ -886,6 +886,7 @@ func (r *RoleBasedGroupReconciler) SetupWithManager(mgr ctrl.Manager, options co
 		Owns(&appsv1.Deployment{}, builder.WithPredicates(WorkloadPredicate())).
 		Owns(&workloadsv1alpha2.RoleInstanceSet{}, builder.WithPredicates(WorkloadPredicate())).
 		Owns(&corev1.Service{}).
+		Owns(&workloadsv1alpha2.RoleBasedGroupScalingAdapter{}, builder.MatchEveryOwner).
```
cheyang: This will trigger RBG reconcile on every RBGSA update, including status-only changes. The adapter controller patches status during bind/scale operations, so you'll get unnecessary reconciles.

Consider using the existing predicate:

```go
.Owns(&workloadsv1alpha2.RoleBasedGroupScalingAdapter{}, builder.WithPredicates(RBGScalingAdapterPredicate()))
```

`RBGScalingAdapterPredicate()` filters spec changes and handles delete events properly - it's already defined in rolebasedgroupscalingadapter_controller.go:308.

sebest: Good catch on the unnecessary reconciles. I've added `RBGScalingAdapterPredicate()` to filter those out.

However, we can't drop `MatchEveryOwner` here. The RBGSA owner references don't set `Controller: true` (see `ReconcileScalingAdapter` at line 796 and `UpdateAdapterOwnerReference` at line 233). Without `MatchEveryOwner`, the default `Owns()` handler uses `metav1.GetControllerOf()` which returns nil for non-controller owner refs, so the watch silently breaks and no reconcile is ever enqueued, regardless of predicates.

The two options compose fine since they modify separate fields on `OwnsInput`:
- `MatchEveryOwner`: scans all owner refs (needed because `Controller` isn't set)
- `WithPredicates(RBGScalingAdapterPredicate())`: filters out status-only updates

Long term we should add `Controller: ptr.To(true)` to the owner ref and drop `MatchEveryOwner`, but that needs a migration strategy for existing clusters. Happy to do that in a follow-up.

PR #242, opened 2026-03-26
feat(scaling-adapter): add Labels field for user-specified RBGSA labels

sebest:
## Summary
- Add a `Labels` field to `ScalingAdapter` spec allowing users to specify custom labels on auto-created RBGSAs
- Custom labels are merged with default labels during RBGSA creation

## Test plan
- [x] Unit tests included
- [x] All existing tests pass


sebest:
@Syspretor I don't think the linter error is caused by my PR but I'd like to have a second opinion, let me know what you think.

Syspretor:
> @Syspretor can I get a review ?

@sebest I’ve reviewed it and I think this PR looks good, so I’ve merged it. Thanks for your contribution!

PR #251, opened 2026-03-28
Security issue: unsafe json string in pulling job's annotations; unsa…

diw-zw:
Fixed some security issues:

1. Use a safer approach to generate JSON strings in the llm pull CLI tool code.
2. Add validation for user-provided file paths in the benchmark dashboard handler.

### DIFF
--- a/cmd/cli/cmd/llm/benchmark/dashboard/handlers.go
+++ b/cmd/cli/cmd/llm/benchmark/dashboard/handlers.go
@@ -101,6 +101,12 @@ func (s *Server) handleExperimentRoutes(w http.ResponseWriter, r *http.Request) 
 
 	experimentName := parts[0]
 
+	// Validate experiment name to prevent path traversal
+	if !isValidName(experimentName) {
+		s.writeError(w, http.StatusBadRequest, "Invalid experiment name")
+		return
+	}
+
 	if len(parts) == 1 {
 		// GET /api/experiments/{name}
 		s.handleGetExperiment(w, r, experimentName)
@@ -171,6 +177,12 @@ func (s *Server) handleGetResult(w http.ResponseWriter, r *http.Request, experim
 		return
 	}
 
+	// Validate inputs to prevent path traversal
+	if !isValidName(experimentName) || !isValidName(fileName) {
+		s.writeError(w, http.StatusBadRequest, "Invalid experiment name or filename")
+		return
+	}
+
 	filePath := filepath.Join(s.dataDir, experimentName, fileName)
 
 	// Security check: ensure the path is within dataDir
@@ -198,6 +210,12 @@ func (s *Server) handleGetResult(w http.ResponseWriter, r *http.Request, experim
 func (s *Server) handleGetImage(w http.ResponseWriter, r *http.Request, experimentName, fileName string) {
 	if r.Method != http.MethodGet {
 		s.writeError(w, http.StatusMethodNotAllowed, "Method not allowed")
+		return
+	}
+
+	// Validate inputs to prevent path traversal
+	if !isValidName(experimentName) || !isValidName(fileName) {
+		s.writeError(w, http.StatusBadRequest, "Invalid experiment name or filename")
 		return
 	}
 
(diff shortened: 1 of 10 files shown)


cmd/cli/cmd/llm/benchmark/dashboard/handlers.go
```diff
@@ -467,17 +507,42 @@ func (s *Server) getResultSummaries(experimentName string) ([]ResultSummary, err
...
+		if !((r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') ||
+			(r >= '0' && r <= '9') || r == '-' || r == '_' || r == '.') {
+			return false
+		}
+	}
+	return true
+}
+
 // isPathSafe checks if the given path is within the data directory.
 func (s *Server) isPathSafe(path string) bool {
```
cheyang, line 556: `isPathSafe` normalizes the string path with `Clean`/`Abs`, but it never resolves symlinks. That still allows a file or directory inside `dataDir` to point outside the tree while passing the prefix check here, and the later `os.ReadFile` / `os.Stat` / `os.ReadDir` calls will follow the symlink. As a result, the current fix closes straightforward `../` traversal but not symlink-based escape. Please resolve symlinks (or explicitly reject symlinked targets/components) before doing the containment check, and add a regression test for that case.

cmd/cli/cmd/llm/pull.go
```diff
@@ -355,11 +355,21 @@ func injectMetadataSave(podTemplate *corev1.PodTemplateSpec, modelID, revision,
 	// Build the metadata JSON using proper JSON encoding to prevent injection
 	// Note: downloadedAt is generated inside the container after download completes
 	// to ensure accurate timing, not at CLI execution time
-	// modelID and revision are JSON-escaped to prevent injection
-	modelIDEscaped := jsonSafeString(modelID)
-	revisionEscaped := jsonSafeString(revision)
-	metadataJSON := fmt.Sprintf(`{"modelID":%s,"revision":%s,"downloadedAt":"%s"}`,
-		modelIDEscaped, revisionEscaped, timestampCmd)
+	metadata := struct {
```
cheyang, line 355: This switches the metadata construction to `json.Marshal`, which is the right direction, but the old escaping-focused tests were removed without adding a behavior-level regression test for the new path. It would be worth adding one that extracts the generated metadata payload and verifies it is valid JSON and still correctly handles quotes, backslashes, and newlines in `modelID` / `revision`, since this code is here specifically to address a security finding.

cmd/cli/cmd/llm/benchmark/dashboard/handlers.go
```diff
@@ -467,17 +507,42 @@ func (s *Server) getResultSummaries(experimentName string) ([]ResultSummary, err
...
+	for _, r := range name {
+		if !((r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') ||
+			(r >= '0' && r <= '9') || r == '-' || r == '_' || r == '.') {
+			return false
+		}
+	}
+	return true
+}
+
 // isPathSafe checks if the given path is within the data directory.
```
cheyang, line 554: `isPathSafe` currently does `Clean`/`Abs` plus a prefix check, but that still validates the pre-dereference path only. A symlink under `dataDir` can still point outside the tree and pass this check, while `os.ReadFile` / `os.Stat` / `os.ReadDir` will follow it.

I think this should resolve symlinks before the containment check. For example:
- `absDataDir -> EvalSymlinks(absDataDir)`
- `absPath -> EvalSymlinks(absPath)`
- compare the resolved target against the resolved data dir with the separator-aware prefix check

Since this code only serves existing files/directories, failing closed when `EvalSymlinks` fails would also be fine here. Something along these lines would address the remaining gap:

```go
func (s *Server) isPathSafe(path string) bool {
    absDataDir, err := filepath.Abs(s.dataDir)
    if err != nil {
        return false
    }
    resolvedDataDir, err := filepath.EvalSymlinks(absDataDir)
    if err != nil {
        return false
    }

    absPath, err := filepath.Abs(filepath.Clean(path))
    if err != nil {
        return false
    }
    resolvedPath, err := filepath.EvalSymlinks(absPath)
    if err != nil {
        return false
    }

    prefix := resolvedDataDir + string(filepath.Separator)
    return strings.HasPrefix(resolvedPath+string(filepath.Separator), prefix)
}
```

Please also add a regression test that creates a symlink inside `dataDir` pointing to a file outside the tree and verifies the request is rejected.

cheyang (approved):
/lgtm
/approve

PR #262, opened 2026-04-09
fix(rbgsa-controller): use Warning event type for failure events

sebest:
## Summary
- Change `FailedGetRBGRole` and `FailedScale` events from `EventTypeNormal` to `EventTypeWarning` — failure events should use Warning per K8s conventions
- Fix `logger.Error` calls that used printf-style `%s` format verbs instead of structured key-value pairs, which produced malformed log entries with literal `%s` in the message

## Test plan
- [ ] `go test ./internal/controller/workloads/` passes
- [ ] `go vet ./internal/controller/workloads/` clean


Syspretor (approved):
/lgtm

PR #264, opened 2026-04-09
fix(kustomize): fix stale references and missing CRDs in kustomization

sebest:
## Summary
- **config/rbac/kustomization.yaml**: update `rbgs_controller_role.yaml` → `role.yaml` (renamed in #209 but the kustomization reference was not updated, breaking `kustomize build config/default`)
- **config/crd/kustomization.yaml**: add 5 missing CRD resources that were generated in `config/crd/bases/` but not listed:
  - `coordinatedpolicies.workloads.x-k8s.io`
  - `instances.workloads.x-k8s.io`
  - `instancesets.workloads.x-k8s.io`
  - `roleinstances.workloads.x-k8s.io`
  - `roleinstancesets.workloads.x-k8s.io`
- **deploy/kubectl/manifests.yaml** and **config/manager/kustomization.yaml**: regenerated via `make build-installer`

## Test plan
- [x] `kustomize build config/default` succeeds
- [x] All 9 CRDs present in regenerated `deploy/kubectl/manifests.yaml`

### DIFF
--- a/config/rbac/kustomization.yaml
+++ b/config/rbac/kustomization.yaml
@@ -1,5 +1,5 @@
 resources:
-  - rbgs_controller_role.yaml
+  - role.yaml
   - role_binding.yaml
   - service_account.yaml
   - secret_role.yaml
--- a/config/crd/kustomization.yaml
+++ b/config/crd/kustomization.yaml
@@ -2,13 +2,15 @@
 # since it depends on service name and namespace that are out of this kustomize package.
 # It should be run by config/default
 resources:
-- bases/workloads.x-k8s.io_rolebasedgroups.yaml
-- bases/workloads.x-k8s.io_rolebasedgroupsets.yaml
 - bases/workloads.x-k8s.io_clusterengineruntimeprofiles.yaml
+- bases/workloads.x-k8s.io_coordinatedpolicies.yaml
+- bases/workloads.x-k8s.io_instances.yaml
+- bases/workloads.x-k8s.io_instancesets.yaml
+- bases/workloads.x-k8s.io_rolebasedgroups.yaml
 - bases/workloads.x-k8s.io_rolebasedgroupscalingadapters.yaml
+- bases/workloads.x-k8s.io_rolebasedgroupsets.yaml
 - bases/workloads.x-k8s.io_roleinstances.yaml
 - bases/workloads.x-k8s.io_roleinstancesets.yaml
-- bases/workloads.x-k8s.io_coordinatedpolicies.yaml
 # +kubebuilder:scaffold:crdkustomizeresource
 
 patches:


deploy/kubectl/manifests.yaml
```diff
@@ -2,11 +2,11 @@ apiVersion: v1
 kind: Namespace
 metadata:
   labels:
+    app.kubernetes.io/managed-by: kustomize
```
Syspretor: The manifest.yaml file is generated by `tools/release/update-release.sh`, so it shouldn’t be modified directly. Please make the change in the script instead.

cheyang (commented):
The kustomize config fixes look correct — the stale `rbgs_controller_role.yaml` reference in `config/rbac/kustomization.yaml` clearly should be `role.yaml` (missed during #209), and adding the missing `instances`/`instancesets` CRDs to `config/crd/kustomization.yaml` is needed.

However, this PR currently has **merge conflicts** that need to be resolved.

Additionally, the regenerated `deploy/kubectl/manifests.yaml` is based on an outdated base. On current main the image tag is `v0.7.0-888ade7` (via `config/manager/kustomization.yaml`), but your regenerated output shows `v0.3.2-4c0f843`. After rebasing onto the latest main, you'll need to re-run the manifests generation so it picks up the current state.

**To unblock this PR:**
1. Rebase onto the latest `main`
2. Re-run `kustomize build` (or whatever generates `deploy/kubectl/manifests.yaml`) against the rebased tree
3. Force-push the updated branch

The two kustomization.yaml fixes themselves are straightforward and correct — just need the rebase + regeneration.

cheyang:
Thanks for catching the missing CRDs in `config/crd/kustomization.yaml` -- the `instances` and `instancesets` resources are indeed present in `config/crd/bases/` but missing from the kustomization on main, so that fix is still needed and valuable.

However, this PR currently has merge conflicts. Main has moved significantly since the PR's base (8+ commits touching `deploy/kubectl/manifests.yaml`, version bumps to v0.7.0, etc.). The RBAC fix (`rbgs_controller_role.yaml` -> `role.yaml`) has also already been resolved on main.

Could you rebase onto the current main branch and re-run `make manifests` to regenerate `deploy/kubectl/manifests.yaml`? After rebasing, the PR scope should narrow to just the CRD kustomization fix plus the regenerated manifests.

config/rbac/kustomization.yaml
```diff
@@ -1,5 +1,5 @@
 resources:
-  - rbgs_controller_role.yaml
+  - role.yaml
   - role_binding.yaml
```
cheyang, line 3: The fix renaming `rbgs_controller_role.yaml` to `role.yaml` has already been applied on `main`. After rebasing, this change will be a no-op. No action needed, but the original issue was real.

cheyang:
Hi @sebest — this PR conflicts with `main`. Please rebase onto the latest `main` and regenerate `manifests.yaml` after the rebase so the kustomize output stays in sync. Once conflicts are resolved I'll pick the review back up. Thanks!

PR #270, opened 2026-04-12
feat(rbgsa-controller): rollout-aware scale-down and spec.replicas drift detection

### LINKED ISSUE #269: [Feature] Rollout-aware scaling for RBGSA on partition-based workloads
## Summary

This proposal adds rollout-awareness to the `RoleBasedGroupScalingAdapter` (RBGSA) controller for **partition-based rolling updates** on StatefulSet and LeaderWorkerSet workloads. Today, when an HPA triggers a scale-down during an in-progress partition-based rollout, the rollout progress is destroyed and must restart from zero. This proposal gates scale-down at the RBGSA level while a partition-based rolling update is in progress, and exposes the gating state via standard Kubernetes conditions and events.

### Background: Partition-Based Rolling Updates

RBG uses **partition-based rolling updates** for StatefulSet and LeaderWorkerSet workloads. The partition value divides pods into two groups:
- Pods with ordinal **>= partition** run the **new template** (already updated)
- Pods with ordinal **< partition** run the **old template** (not yet updated)

The rollout progresses by decrementing partition from `replicas` toward `0`, updating pods from the highest ordinal down. This is controlled by the STS reconciler (`sts_reconciler.go`, Cases 3-6 in `rollingUpdateParameters`) and the coordination layer for multi-role rollouts.

### Current Architecture

The scaling data flow is **unidirectional** with no feedback loop from rollout state:

```
HPA → RBGSA.Spec.Replicas → RBGSA controller → RBG.Spec.Roles[].Replicas → RBG controller → Workload
```

The RBGSA controller watches only RBGSA objects (`SetupWithManager` at `rolebasedgroupscalingadapter_controller.go:302-308`). It does not watch RBG objects, workloads, or their status. When HPA writes a new `spec.replicas` via the scale subresource, the RBGSA controller unconditionally propagates it to `rbg.Spec.Roles[].Replicas` via `updateRoleReplicas()` — without checking whether a rolling update is in progress.

The RBG controller watches RBGSA via `Owns()` (`rolebasedgroup_controller.go:905`), so RBGSA spec changes trigger RBG reconciliation. But the reverse does not exist — there is no mechanism for RBG rollout state changes to reach the RBGSA controller. A [TODO comment](https://github.com/sgl-project/rbg/blob/main/internal/controller/workloads/rolebasedgroupscalingadapter_controller.go#L139) in the existing code acknowledges this gap:

```go
// TODO: currently reconcile unbound adapter by a default reconcile interval, need to implement a rbg event-driven manager
```

### The Problem: Scale-Down Kills Updated Pods

StatefulSet and LeaderWorkerSet remove pods from the **highest ordinals** on scale-down. Partition-based rollout updates pods from the **highest ordinals** down. These two mechanisms directly conflict: **scale-down kills the already-updated pods first**, resetting rollout progress to zero.

**Concrete example**:

```
Before scale-down:  replicas=10, partition=7
                    Pods:  [0  1  2  3  4  5  6 | 7  8  9]
                            ── old template ──   ─ new ─

HPA scales down to 6:
  StatefulSet deletes pods 9, 8, 7 (highest ordinals — the updated ones)
  Partition clamps to 6 (STS reconciler Case 5: partition = min(partition, burstReplicas))

After scale-down:   replicas=6, partition=6
                    Pods:  [0  1  2  3  4  5]
                            ── all old template ──
                    Rollout progress: 3 → 0 updated pods. Restarts from scratch.
```

This is a **fundamental constraint of Kubernetes StatefulSet ordering** — no partition value adjustment can prevent the loss of already-updated pods during scale-down. The only defense is to defer the scale-down until the rollout completes.

**Note on other workload types**: Deployments handle rollout via ReplicaSets and the Kubernetes Deployment controller proportionally scales both old and new ReplicaSets during concurrent scaling — scale-down does not preferentially kill new-revision pods. This proposal focuses on the partition-based workloads (StatefulSet, LWS) where the problem is critical.

### Secondary Problem: Coordination Partition Recalculation

When a `CoordinatedPolicy` uses percentage-based partitions (e.g., `"70%"`), the partition value is recalculated relative to `role.Replicas` (`rolebasedgroup_controller.go:1153`). A scale event changes `role.Replicas`, causing the partition floor to shift:

```
Before:  "70%" of 10 replicas = partition 7  (3 pods updated)
After:   "70%" of 6 replicas  = partition 5  (1 pod updated — but those pods were deleted)
```

This compounds the ordinal-deletion problem by also breaking the coordination partition math across coordinated roles.

### Impact Analysis

| Scenario | Partition behavior | Updated pods | Severity |
|---|---|---|---|
| Scale UP, no coordination | Partition unchanged | New pods get new template (ordinal >= partition) | **Low** |
| Scale UP, with coordination | Percentage partition recalculates upward | Floor partition may stall rollout temporarily | **Medium** |
| Scale DOWN, no coordination | Partition clamped to new replicas | Already-updated pods deleted, rollout restarts | **Critical** |
| Scale DOWN, with coordination | Partition recalculated + cross-role skew breaks | Updated pods lost, coordination integrity lost | **Critical** |

#### Goals

1. Gate scale-down operations at the RBGSA controller level while a partition-based rolling update is in progress for the target role
2. Allow scale-up operations to proceed normally during rollout (new pods land on the new template since their ordinal >= partition)
3. Expose gating state via standard Kubernetes conditions on RBGSA status and Warning events
4. Make this the unconditional default behavior for partition-based workloads
5. Add a `Watches()` on RBG to the RBGSA controller for event-driven rollout completion detection (resolving the existing TODO)

#### Non-Goals

1. Fixing scale-up partition regression with coordination percentage-based partitions
2. Gating for Deployment-backed roles (Deployments handle concurrent scaling + rollout natively)
3. Configurable opt-out annotation (can be added later if needed)
4. Auto-release timeout for stuck rollouts (a Warning event after sustained deferral is included, but not an automatic bypass)

---

#### Gate at the RBGSA controller level, not at the workload reconciler

The RBGSA controller is the right place to gate because:

- It sits at the **entry point** of the scaling data flow — gating here prevents the replica change from ever reaching `rbg.Spec.Roles[].Replicas`
- Since `rbg.Spec.Roles[].Replicas` stays unchanged during gating, the **CoordinatedPolicy sees no disruption** — `handleCoordinationStrategies()` reads `rbg.Spec.Roles[].Replicas` and `RoleStatus` which are both stable. No coordination recalculation, no skew violation.
- The STS/LWS reconcilers don't need changes — they never see the scale-down
- The RBG controller's reconcile loop (Step 4→5→7: status → coordination → roles) continues undisturbed

Fixing the problem at the workload reconciler level (e.g., adjusting partition math in STS reconciler Case 5) cannot work because StatefulSet always deletes from the highest ordinal — no partition value prevents the deletion of updated pods.

#### Scope to partition-based workloads only

The scale-down ordinal problem is specific to workloads that use partition-based rolling updates (StatefulSet, LeaderWorkerSet). Kubernetes Deployments handle concurrent rollout and scaling natively — the Deployment controller proportionally adjusts both old and new ReplicaSets during scale-down.

Gating is applied only when the target role uses a partition-based workload (StatefulSet, LWS, or RoleInstanceSet in stateful mode). Deployment-backed roles are excluded.

This requires the RBGSA controller to determine the workload type. The role's `Workload` field in `RoleSpec` provides this information, and is already accessible since the RBGSA controller fetches the full RBG object.

#### Condition only present when gating

Per Kubernetes API conventions, we use the condition only in the abnormal state. The `ScaleDownDeferred` condition is **added** when gating is active, and **removed** when scaling proceeds normally. This avoids cluttering every RBGSA with a `ScaleDownDeferred=False` condition during normal operation.

### Interaction with CoordinatedPolicy

This is an important architectural property of the proposed solution.

**When RBGSA gates a scale-down:**

1. RBGSA does NOT call `updateRoleReplicas()` — so `rbg.Spec.Roles[role].Replicas` stays at the current value
2. RBG controller reconciles normally:
   - `constructAndUpdateRoleStatuses()` (Step 4) reads unchanged workload state
   - `handleCoordinationStrategies()` (Step 5) reads unchanged `role.Replicas` and `RoleStatus` → produces unchanged `scalingTargets` and `rollingUpdateStrategies`
   - `reconcileRoles()` (Step 7) applies unchanged values
3. Coordination maxSkew constraints are **not affected** because no spec values changed

**When RBGSA releases the gate (rollout completes):**

1. RBGSA calls `updateRoleReplicas()` — changes `rbg.Spec.Roles[role].Replicas` to HPA's desired value
2. RBG controller reconciles with the new spec replicas
3. Coordination recalculates with the new desired values — this is the normal coordination path, same as any user-initiated replica change

**Cross-role scenario** (known limitation):

If roles A and B are coordinated with maxSkew, and HPA wants to scale down both:
- Role A is mid-rollout → RBGSA gates its scale-down → spec stays at current
- Role B is not rolling → RBGSA allows its scale-down → spec changes

Now coordination sees A at old (higher) replicas and B at new (lower) replicas. The coordination scaler (`scaler.go:70-172`) computes progress as `CurrentReplicas/DesiredReplicas`. Since A's desired is still high (unchanged spec), its progress is normal. B's progress may spike (current > desired momentarily). The coordination system handles this gracefully — it would hold B from scaling further until A catches up. But the intent divergence (HPA wanted both to scale down, only one did) may require operator awareness. The `ScaleDownDeferred` condition on the gated RBGSA makes this visible.

This is documented as a known limitation. The alternative — letting both scale down and destroying A's rollout — is worse.

---

#### Motivation

The RBGSA controller currently relies on polling (`RequeueAfter`) to detect state changes in the RBG. This is acknowledged as a gap by a [TODO in the existing code](https://github.com/sgl-project/rbg/blob/main/internal/controller/workloads/rolebasedgroupscalingadapter_controller.go#L139). Adding a `Watches()` on RBG objects enables event-driven reconciliation — the RBGSA controller will be notified immediately when the RBG's rollout state changes, removing the polling delay.

#### Implementation

In `SetupWithManager()`, add a watch on RBG that enqueues associated RBGSA(s):

```go
func (r *RoleBasedGroupScalingAdapterReconciler) SetupWithManager(mgr ctrl.Manager, options controller.Options) error {
    return ctrl.NewControllerManagedBy(mgr).
        WithOptions(options).
        For(&workloadsv1alpha2.RoleBasedGroupScalingAdapter{},
            builder.WithPredicates(RBGScalingAdapterPredicate())).
        Watches(&workloadsv1alpha2.RoleBasedGroup{},
            handler.EnqueueRequestsFromMapFunc(r.findScalingAdaptersByRBG),
            builder.WithPredicates(rbgStatusChangedPredicate())).
        Named("workloads-rolebasedgroup-scalingadapter").
        Complete(r)
}
```

The mapper function finds RBGSA objects whose `ScaleTargetRef.Name` matches the changed RBG. The predicate filters to only `status.roleStatuses` changes (not all RBG changes).

A `RequeueAfter: 30s` is retained as a **safety net** for cases where the watch event is missed, not as the primary trigger.

#### Acceptance Criteria

- [ ] RBGSA controller watches RBG objects via `Watches()` with a mapper function
- [ ] Predicate filters to status changes only (not spec changes, to avoid reconcile storms)
- [ ] Mapper finds RBGSA objects by `ScaleTargetRef.Name` match
- [ ] Existing TODO comment at line 139 is resolved

---

#### Rollout Detection

The RBGSA controller already fetches the full RBG object during reconciliation (line 106). Rollout state for partition-based workloads is detected per-role using:

```go
roleStatus, found := rbg.GetRoleStatus(targetRoleName)
if found && isPartitionBasedWorkload(role) && roleStatus.UpdatedReplicas < roleStatus.Replicas {
    // Partition-based rolling update in progress for this role
}
```

`RoleStatus` is populated by each workload reconciler:

| Workload | `Replicas` source | `UpdatedReplicas` source | Partition-based? |
|---|---|---|---|
| StatefulSet | `sts.Spec.Replicas` | `sts.Status.UpdatedReplicas` | **Yes** |
| LeaderWorkerSet | `lws.Status.Replicas` | `lws.Status.UpdatedReplicas` | **Yes** |
| RoleInstanceSet | `ris.Spec.Replicas` | `ris.Status.UpdatedReplicas` | **Yes** |
| Deployment | `deploy.Spec.Replicas` | `deploy.Status.UpdatedReplicas` | No — excluded from gating |

**Note on maxSurge**: For StatefulSet with maxSurge > 0, `sts.Spec.Replicas` includes surge pods (e.g., 12 for a role with 10 replicas and maxSurge=2). This means `UpdatedReplicas < Replicas` may be true even at the end of the rollout while surge replicas are being reclaimed. This is safe-side over-gating: the gate holds slightly longer than strictly necessary.

#### Gating Logic

In the `Reconcile` method, between lines 209 and 211:

```go
isScaleDown := *desiredReplicas < *currentReplicas
if isScaleDown && isPartitionBasedWorkload(targetRole) {
    roleStatus, found := rbg.GetRoleStatus(targetRoleName)
    if found && roleStatus.UpdatedReplicas < roleStatus.Replicas {
        // 1. Emit Warning event
        // 2. Set ScaleDownDeferred condition (with ObservedGeneration)
        // 3. Patch RBGSA status
        // 4. Return RequeueAfter: 30s (safety net; primary trigger is RBG watch)
    }
}
// Scale-up, Deployment-backed roles, and non-rollout scale-down proceed normally
```

When scaling proceeds (scale-up, or scale-down after rollout completes), the `ScaleDownDeferred` condition is **removed** from RBGSA status.

#### HPA Interaction

When RBGSA defers a scale-down:
- HPA has already written `spec.replicas` via the scale subresource (200 OK)
- RBGSA does not propagate the change to RBG, so `status.replicas` reflects the actual count
- HPA sees `status.replicas > spec.replicas` — it recognizes there are "too many" replicas but its desired value is already set
- Once rollout completes, the next reconcile propagates the pending scale-down

**Edge case**: If HPA adjusts to a value between current and original desired (e.g., from `spec.replicas=6` to `spec.replicas=8` while current is 10), the RBGSA controller sees `8 < 10` — still a scale-down — and continues to defer. Event messages include both `spec.replicas` and actual count to make this clear.

#### Edge Cases

| Edge case | Behavior | Rationale |
|-----------|----------|-----------|
| Role status not found (`found=false`) | Scale-down proceeds | Don't block on missing status; fresh RBG may not have status yet |
| Rollout stuck (bad image, crash loop) | Scale-down blocked; condition makes state visible | User must fix/revert rollout; escalated warning is future work |
| Initial pod creation (`UpdatedReplicas=0, Replicas=N`) | Scale-down deferred | Conservative; unlikely scenario (HPA wouldn't scale down during creation) |
| HPA adjusts UP during deferral (e.g., 6 → 12, current=10) | Scale-up proceeds; condition removed | `12 > 10` is not a scale-down |
| HPA adjusts to mid-value during deferral (e.g., 6 → 8, current=10) | Still deferred (`8 < 10`) | Event message clarifies the situation |
| Multiple deferred adjustments (6 → 4 → 3) | Latest `spec.replicas` wins when gate releases | Natural behavior — last write to spec is what's applied |
| Scale-down to zero during rollout | Deferred | Intentional teardown should be done by deleting the RBG or disabling the scaling adapter |
| Deployment-backed role during rollout | Scale-down proceeds (no gating) | Deployments handle concurrent scaling + rollout natively via ReplicaSet proportioning |

#### Acceptance Criteria

- [ ] Scale-down is deferred for partition-based workloads when `RoleStatus.UpdatedReplicas < RoleStatus.Replicas`
- [ ] Scale-down proceeds normally for Deployment-backed roles regardless of rollout state
- [ ] Scale-up proceeds normally regardless of rollout state or workload type
- [ ] Scale-down proceeds after rollout completes (condition removed)
- [ ] Warning event emitted on first deferral, with both `spec.replicas` and actual count
- [ ] Event not re-emitted on subsequent requeues (rate-limited to first occurrence)
- [ ] `RequeueAfter: 30s` as safety net

---

#### Type Changes

Add `Conditions` field to `RoleBasedGroupScalingAdapterStatus`:

```go
type RoleBasedGroupScalingAdapterStatus struct {
    Phase         constants.AdapterPhase `json:"phase,omitempty"`
    Replicas      *int32                 `json:"replicas,omitempty"`
    Selector      string                 `json:"selector,omitempty"`
    LastScaleTime *metav1.Time           `json:"lastScaleTime,omitempty"`

    // Conditions represent the latest available observations of the adapter's state.
    // +optional
    Conditions []metav1.Condition `json:"conditions,omitempty"`
}
```

#### Condition Type

Constant defined in the API types package (`api/workloads/v1alpha2/rolebasedgroupscalingadapter_types.go`):

```go
const (
    // AdapterConditionScaleDownDeferred indicates that a scale-down request
    // is being held because a partition-based rolling update is in progress
    // for the target role. This condition is only present when gating is
    // active; it is removed when scaling proceeds normally.
    AdapterConditionScaleDownDeferred = "ScaleDownDeferred"
)
```

Condition includes `ObservedGeneration` set to `rbgScalingAdapter.Generation`.

| Condition | Status | Reason | Present when |
|-----------|--------|--------|------|
| `ScaleDownDeferred` | `True` | `RollingUpdateInProgress` | Scale-down requested but partition-based rollout active |
| *(removed)* | | | Scaling proceeded or no scale-down pending |

#### Event Reasons

New event reasons in `event.go`:

```go
const (
    ScaleDownDeferred = "ScaleDownDeferred"
)
```

Example event messages:
```
Warning  ScaleDownDeferred  Scale-down to 6 replicas deferred (current: 10):
                            partition-based rolling update in progress for role
                            worker (updated 3/10)
Normal   SuccessfulScale    Succeed to scale target role [worker] of rbg [my-rbg]
                            from 10 to 6 replicas
```

#### Status Patching

The existing `ToRoleBasedGroupScalingAdapterStatusApplyConfiguration()` helper must be extended to include `Conditions`. Since the RBGSA controller is the sole field manager for RBGSA status, SSA with conditions is straightforward — include all conditions in every status patch.

Use `apimeta.SetStatusCondition` for condition manipulation on the API type, then convert to the apply configuration for the SSA patch.

#### Codegen

After type changes:
```bash
make generate   # deepcopy
make manifests  # CRD manifests
```

Verify:
- `api/workloads/v1alpha2/zz_generated.deepcopy.go` includes `Conditions` copy
- `client-go/applyconfiguration/workloads/v1alpha2/rolebasedgroupscalingadapterstatus.go` has `Conditions` field and `WithConditions()` method
- CRD manifest includes `conditions` in the status schema

#### Acceptance Criteria

- [ ] `Conditions` field added to RBGSA status type with `+optional` marker
- [ ] Condition type constant defined in API types package
- [ ] Apply configuration regenerated with `Conditions` support
- [ ] CRD manifests regenerated
- [ ] `ObservedGeneration` set on all conditions
- [ ] Condition added on gating, removed on scaling or no-op
- [ ] Warning events emitted on state transitions
- [ ] Event rate-limiting during repeated requeue cycles

---

#### Unit Tests

Table-driven tests in `rolebasedgroupscalingadapter_controller_test.go`:

| Test case | Setup | Expected |
|-----------|-------|----------|
| Scale-down during rollout (STS) | STS role, `UpdatedReplicas=3, Replicas=10, desired=6` | Deferred, Warning event, condition set, requeue |
| Scale-down during rollout (LWS) | LWS role, `UpdatedReplicas=3, Replicas=10, desired=6` | Deferred, Warning event, condition set, requeue |
| Scale-down during rollout (Deployment) | Deployment role, `UpdatedReplicas=3, Replicas=10, desired=6` | **Proceeds normally** (no gating for Deployments) |
| Scale-down no rollout | STS role, `UpdatedReplicas=10, Replicas=10, desired=6` | Proceeds normally, no condition |
| Scale-up during rollout | STS role, `UpdatedReplicas=3, Replicas=10, desired=15` | Proceeds normally, no condition |
| Scale-down after rollout completes | Previously deferred, `UpdatedReplicas==Replicas` | Proceeds, condition removed |
| Role status not found | `GetRoleStatus` returns `found=false` | Proceeds (don't block) |
| Scale-down during initial creation | `UpdatedReplicas=0, Replicas=10, desired=6` | Deferred (conservative) |
| Scale-down with maxSurge active | `UpdatedReplicas=8, Replicas=12 (10+2 surge), desired=6` | Deferred |
| HPA adjusts UP during deferral | Deferred at 6, HPA writes 12 (>current 10) | Scale-up proceeds, condition removed |
| HPA adjusts to mid-value | Deferred at 6, HPA writes 8 (still <10) | Still deferred |
| Repeated requeue during deferral | Multiple reconciles while deferred | Condition stays, event not re-emitted |
| Condition has ObservedGeneration | Any condition update | `ObservedGeneration` matches adapter's `Generation` |

#### Acceptance Criteria

- [ ] All unit test cases pass
- [ ] Existing RBGSA controller tests continue to pass
- [ ] Test coverage for condition lifecycle (added → removed)
- [ ] Test coverage for event emission and rate-limiting
- [ ] Test coverage for workload-type-aware gating (Deployment vs STS/LWS)

---

## Files to Modify

| File | Change |
|------|--------|
| `internal/controller/workloads/rolebasedgroupscalingadapter_controller.go` | Add RBG watch, gate logic, condition management, workload type check |
| `api/workloads/v1alpha2/rolebasedgroupscalingadapter_types.go` | Add `Conditions` field + condition constant |
| `internal/controller/workloads/event.go` | Add `ScaleDownDeferred` |
| `client-go/applyconfiguration/workloads/v1alpha2/rolebasedgroupscalingadapterstatus.go` | Add `Conditions` support |
| `api/workloads/v1alpha2/zz_generated.deepcopy.go` | Regenerate |
| CRD manifests | Regenerate |
| `internal/controller/workloads/rolebasedgroupscalingadapter_controller_test.go` | New test cases |

---

### 1. Cross-role coordination divergence

When two coordinated roles both need scale-down, one may be gated (mid-rollout) while the other proceeds. The coordination scaler handles this gracefully in steady state, but the intent divergence (HPA wanted both to scale down) may require operator awareness. The `ScaleDownDeferred` condition on the gated RBGSA makes this visible.

### 2. Stuck rollout blocks scale-down indefinitely

If a rollout never completes (bad image, crashlooping), scale-down is blocked until the rollout is fixed or reverted. The `ScaleDownDeferred` condition on RBGSA status makes the state visible. A future enhancement could add a `ScaleDownDeferredTooLong` escalated warning event after sustained deferral (~5 minutes).

### 3. MaxSurge over-gating

During StatefulSet rollout with maxSurge, `sts.Spec.Replicas` includes surge pods. The gate may hold slightly longer than necessary while surge replicas are being reclaimed. This is safe-side behavior.

---

## Future Work

1. **Scale-up partition regression**: With coordination percentage-based partitions, scale-up recalculates the partition floor relative to the new replica count, potentially stalling rollout. A ratio-preservation formula in the STS reconciler could fix this.

2. **Configurable opt-out**: An annotation like `rbg.workloads.x-k8s.io/rollout-aware-scaling: disabled` on the RBGSA to bypass gating for specific use cases (e.g., intentional scale-to-zero during rollout).

3. **Stuck rollout escalation event**: Emit a `ScaleDownDeferredTooLong` Warning event after the deferred state persists for ~5 minutes, using the condition's `LastTransitionTime`. Purely informational for operators.

4. **Auto-release timeout**: A configurable duration after which scale-down is allowed even during rollout, to handle permanently stuck rollouts without manual intervention.

---

## Related Resources

- STS reconciler rollout logic: `pkg/reconciler/sts_reconciler.go` (Cases 3-6 in `rollingUpdateParameters`)
- RBGSA controller: `internal/controller/workloads/rolebasedgroupscalingadapter_controller.go`
- RBG coordination: `internal/controller/workloads/rolebasedgroup_controller.go` (`calculateRollingUpdateForPolicyRule`)
- Coordination scaling: `pkg/coordination/coordinationscaling/scaler.go`
- Existing TODO for RBG event-driven watch: `rolebasedgroupscalingadapter_controller.go:139`

sebest:
This PR adds two related RBGSA-controller behaviors. They are bundled because the second reuses the watch infrastructure introduced for the first.

## Summary

**1. Rollout-aware scale-down gating (original scope)**

- Gate scale-down at the RBGSA controller level for partition-based workloads (StatefulSet, LWS, RoleInstanceSet) while a rolling update is in progress.
- StatefulSet deletes highest-ordinal pods on scale-down, which are the already-updated pods — this destroys rollout progress.
- Scale-up and Deployment-backed roles are unaffected.
- Opt-in via `scalingAdapter.scaleDownPolicy: DeferDuringRollout` (default is `Unrestricted` for backward compatibility).

**2. Wake on per-role `spec.replicas` drift (new — second commit)**

- Extend the RBG-watch predicate (renamed `RBGRoleStatusPredicate` → `RBGRoleSpecOrStatusPredicate`) to also fire when any role's `spec.replicas` changes between old and new.
- Without this, the RBGSA controller's writeback path only wakes on status changes, so an external mutator that stomps `spec.roles[].replicas` (e.g. `helm upgrade` whose template stops emitting the field, `kubectl edit` from a manifest that omits replicas, non-SSA scripts) waits for the next status-change event before the apiserver-stored spec is corrected.
- Pairs with the in-memory override added in #306: that override prevents the workload from being sized to the stale value on the next reconcile, and this predicate change closes the latency gap by waking the writeback immediately on the stomp instead of when status next changes.
- New `rolesReplicasDiffer` helper matches roles by name; added, removed, and renamed roles all count as a difference. Other spec fields (image, env, command) deliberately do not fire the predicate — the adapter reconcile is irrelevant to them.

## Changes

**Rollout-aware scale-down:**

- Add `Conditions` field and `ScaleDownDeferred` condition to RBGSA status
- Add `ScaleDownPolicy` field to `ScalingAdapter` spec (`DeferDuringRollout` / `Unrestricted`)
- Detect rollout via `RoleStatus.UpdatedReplicas < Replicas` (workload-agnostic)
- Gate only partition-based workloads using existing `IsStatefulRole()` helper
- Add `Watches()` on RBG for event-driven rollout completion detection (resolves existing TODO)
- Emit Warning event on first deferral only (rate-limited, not on every requeue)
- Use SSA for all status writes (consistent field manager ownership)
- DeepCopy after Get to prevent informer cache corruption
- Expand `doc/features/autoscaler.md` with full scaling adapter documentation
- Update API reference and variables docs

**Spec drift detection:**

- Rename `RBGRoleStatusPredicate` → `RBGRoleSpecOrStatusPredicate`; restructure `UpdateFunc` to OR-in a per-role `spec.replicas` check
- Add `rolesReplicasDiffer` helper (matches by role name; nil/non-nil pointer transitions and added/removed/renamed roles all count as differences)

## Design proposal

Closes #269

## Companion PR

- #306 — adds `applyRBGSAReplicasOverride` at the top of the RBG `Reconcile` so the workload is never sized from a stale spec value. Together with the predicate extension here, this provides defense-in-depth against external mutators that drop `spec.roles[].replicas`.

## Test plan

- [x] All RBGSA controller unit tests pass, including:
  - `TestScaleDownDeferredDuringRollout`, `TestScaleDownDeferredConditionLifecycle`, `TestFindScalingAdaptersByRBG` (rollout gating)
  - `TestRBGRoleSpecOrStatusPredicate` (renamed; extended with 6 new sub-tests covering spec.replicas nil↔set, equal-but-other-fields-differ, added/removed/renamed roles)
  - `TestRBGStatusChangedPredicate` (envtest-side, also updated to use the renamed predicate)
- [x] Full controller test suite passes (`go test ./internal/controller/workloads/...`)
- [x] `make generate && make manifests` — codegen clean
- [x] `make lint` — passes (cyclomatic complexity under 30)
- [ ] Manual: create RBG with STS role + RBGSA + HPA, trigger rollout, observe scale-down deferred
- [ ] Manual: simulate `helm upgrade` that drops `spec.roles[].replicas`, observe RBGSA writeback fires immediately (paired with #306)


Syspretor:
@sebest I think the motivation behind this change makes a lot of sense. Please give us a little time to think through the API and workflow design, and I’ll follow up with feedback afterward.

Syspretor:
@sebest We reviewed this PR in our community weekly meeting, and overall everyone really appreciated both the motivation and the design.

The only feedback we had was around the naming/semantics of `DeferDuringRollout`, which feels a bit ambiguous to us. Would it make sense to consider a clearer name, something along the lines of `DeferUntilRolloutCompleted`, to make the intended behavior more explicit?

Syspretor:
@sebest the e2e failure is not caused by this PR. It was due to an issue on the main branch, which I’ve already fixed. Could you please rebase onto the latest main branch and push again to rerun the e2e tests? Thanks!

Syspretor:
@sebest btw, We’d also be glad to have you join the community discussions. If you’re interested, please feel free to reach out to me at tongyuguo@outlook.com.

internal/controller/workloads/rolebasedgroupscalingadapter_controller.go
```diff
@@ -213,37 +218,155 @@ func (r *RoleBasedGroupScalingAdapterReconciler) Reconcile(ctx context.Context,
...
+		return false, ctrl.Result{}, nil
+	}
+
+	roleStatus, found := rbg.GetRoleStatus(targetRoleName)
+	if !found || roleStatus.UpdatedReplicas >= roleStatus.Replicas {
+		return false, ctrl.Result{}, nil
+	}
+
+	logger := log.FromContext(ctx)
+	msg := fmt.Sprintf(
```
cheyang, line 303: Structured logger (klog/logr) doesn't interpret `%s` format verbs — this ends up logging the literal `%s` in the message string. Should be key-value style:

```go
logger.Error(err, "Failed to update status", "rbgScalingAdapterName", rbgScalingAdapter.Name)
```

cheyang (commented):
## Review: rollout-aware scale-down & spec.replicas drift detection

**Status**: This PR currently has merge conflicts — needs rebase onto latest `main` (as previously requested).

### Architecture & Design

The approach is well-considered:
- Gating at the RBGSA controller level rather than the workload controller is the right call — it keeps the scaling adapter self-contained and avoids entangling rollout logic with the core RBG reconciler.
- The distinction between partition-based workloads (STS, LWS, RoleInstanceSet) and Deployments is correctly scoped via `IsStatefulRole()`.
- Default of `Unrestricted` preserves backward compatibility without surprise behavioral changes for existing users.
- The `rolesReplicasDiffer` predicate for spec.replicas drift detection is narrowly scoped (only fires on replica count changes, not other spec mutations), which avoids excessive reconcile churn.

### Correctness

- `deferScaleDownIfNeeded` correctly short-circuits for scale-up (`desiredReplicas >= currentReplicas`).
- Rollout detection via `UpdatedReplicas < Replicas` from RBG status is reasonable — it reflects the workload controller's view of partition progress.
- Event emission only on first deferral (checking condition presence before emitting) avoids log/event noise on every 30s requeue.
- The DeepCopy after Get prevents informer cache mutation — important since conditions are modified in-place.
- SSA for status writes is consistent with the existing pattern in this controller.

### Minor Issue

One structured-logger misuse (see inline comment): `logger.Error(err, "Failed to update status for %s", ...)` — `%s` isn't interpolated by logr.

### Tests

Coverage is thorough:
- 13 table-driven unit test cases covering all permutations of workload type, policy, rollout state, and scaling direction
- Condition lifecycle tests (set/clear/idempotent re-set)
- 6 envtest integration tests covering the full controller interaction including scale subresource writes and RBG status watch
- Predicate tests for the new `rolesReplicasDiffer` logic

### Summary

The design and implementation are solid. The blocking issue is the merge conflict. Once rebased and CI passes, this looks ready to merge pending the one-line logger fix.

cheyang:
Nicely designed feature with strong test coverage (810 lines of unit tests + 350 lines of envtest). The rollout-gating logic, condition lifecycle, and SSA status writes all look correct. The DeepCopy-after-Get fix for informer cache safety is a good catch.

I left one concrete comment on a malformed `logger.Error` call in the scale-down error path, and a minor test coverage suggestion.

The PR currently has merge conflicts with `main` -- a rebase is needed before this can be merged.

internal/controller/workloads/rolebasedgroupscalingadapter_controller.go
```diff
@@ -213,37 +218,155 @@ func (r *RoleBasedGroupScalingAdapterReconciler) Reconcile(ctx context.Context,
...
 			"Failed to scale target role [%s] of rbg [%s] from %v to %v replicas: %v",
-			targetRoleName, rbgName, *currentReplicas, *desiredReplicas, err,
+			targetRoleName, rbgName, currentReplicas, desiredReplicas, err,
 		)
 		return ctrl.Result{}, err
 	}
 	rbgScalingAdapterApplyConfig := ToRoleBasedGroupScalingAdapterApplyConfiguration(rbgScalingAdapter).
 		WithStatus(
-			ToRoleBasedGroupScalingAdapterStatusApplyConfiguration(rbgScalingAdapter.Status, true).WithReplicas(*desiredReplicas),
+			ToRoleBasedGroupScalingAdapterStatusApplyConfiguration(rbgScalingAdapter.Status, true).WithReplicas(desiredReplicas),
```
cheyang, line 356: This `logger.Error` call uses `%s` in the message string and passes `rbgScalingAdapter.Name` as a lone positional arg:
```go
logger.Error(err, "Failed to update status for %s", rbgScalingAdapter.Name)
```
In logr, `Error(err, msg, keysAndValues...)` does not interpolate format verbs -- `%s` appears literally in the output, and the adapter name becomes an orphaned key (odd-numbered arg) that triggers a `MISSING` value warning in the structured log.

Consider:
```go
logger.Error(err, "Failed to update status", "rbgScalingAdapterName", rbgScalingAdapter.Name)
```

internal/controller/workloads/rolebasedgroupscalingadapter_rollout_test.go
```diff
@@ -0,0 +1,810 @@
...
+			rbgsa:              buildBoundRBGSA(0),
+			workloadAPIVersion: "apps/v1",
+			workloadKind:       "StatefulSet",
+			scaleDownPolicy:    ptr.To(workloadsv1alpha2.ScaleDownPolicyDeferDuringRollout),
+			roleReplicas:       10,
+			roleStatuses: []workloadsv1alpha2.RoleStatus{
+				{Name: "worker", Replicas: 10, ReadyReplicas: 7, UpdatedReplicas: 3},
+			},
+			expectDeferred:     true,
+			expectRequeueAfter: 30 * time.Second,
```
cheyang, line 345: The condition lifecycle tests cover deferral, rollout completion, and HPA adjusting upward, but there is no test for the scenario where the HPA adjusts desired replicas to exactly match current replicas while a `ScaleDownDeferred` condition is active. In that case `clearScaleDownDeferredCondition` on the "nothing to do" path should remove the stale condition. A small table-driven case would close this gap.

This is a minor gap -- the path is exercised implicitly, but an explicit test documents the expected behavior.

cheyang:
Hi @sebest — this PR now conflicts with `main`. Please rebase onto the latest `main` and force-push so we can validate the changes with green CI. The previously noted `logger.Error` format-string fix is still outstanding as well. Thanks!

Issue #260, opened 2026-04-08
[Enhancement] Add headless Service for leaders


JasonHe-WQ:
### Checklist

- [ ] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [ ] 2. Please use English, otherwise it will be closed.

### Motivation

Currently, sglang only support `headless` mode for DP accross nodes. And in large EP distribution, cross nodes DP is critical to avoid cross nodes attention operation and save GPU memory to allocate more kv cache.

In this circumstance, only the leader node would luanch fully operational API server and serve requests while others would only luanch a dummy API server. Therefore, it would be nice to have an independent svc resource for leaders of each `roleinstanceset` in leader worker pattern.

In `rbg`, only one service would be created for each `roleinstanceset`, which means it may be hard to operate large EP without external gateway when using `sglang` to serve model across nodes.

https://github.com/sgl-project/rbg/blob/18d454d61f4337691838c4a6fce9b2255a382c91/pkg/reconciler/roleinstanceset_reconciler.go#L68-L76

https://github.com/sgl-project/rbg/blob/6114db1855cdc495e44dd8d79bbf66f12f438f0d/pkg/reconciler/svc_reconciler.go#L106-L121

#### sglang source code:
For followers, only dummy api server would be luanched

https://github.com/sgl-project/sglang/blob/46bf19cdab3b3d64f8040cb63eadf19054e64f4f/python/sglang/srt/entrypoints/engine.py#L681-L698

##### Case 1: headless

https://github.com/vllm-project/vllm/blob/72cad44d3c6eaac69d8353981b9f2b4439d12ecd/vllm/entrypoints/cli/serve.py#L115-L122

##### Case 2: hybrid
on this occasion, each node __CAN__ serve fully operational API server, and we can use `service` for the `roleinstanceset` as a gateway

PR #273, opened 2026-04-13
feat(rbg): add leader only service

JasonHe-WQ:
### Ⅰ. Motivation
#260

fixes #XXXX

## Checklist


keps/260-leaderonly-service/README.md
```diff
@@ -0,0 +1,297 @@
...
+
+## Design Details
+
+### API
+
+```go
+
+type RoleSpec struct {
+    // +optional
+    NetworkConfig *NetworkConfig `json:"networkConfig,omitempty"`
```
cheyang: I’m not sure `RoleSpec.NetworkConfig.TargetPolicy` is the right API boundary here. Since `LeaderOnly` appears to be meaningful only for `leaderWorkerPattern` (where a leader/worker topology actually exists), placing it under a generic role-level `NetworkConfig` feels a bit too broad. I’d suggest either scoping this under `leaderWorkerPattern` directly, or at least making the KEP explicitly state that `LeaderOnly` is only supported for `leaderWorkerPattern` and must be rejected for other patterns. A more targeted shape could look like:
```go
type LeaderWorkerPattern struct {
    // +optional
SharedServiceSelection *SharedServiceSelectionPolicy `json:"sharedServiceSelection,omitempty"`
}

type SharedServiceSelectionPolicy string

const (
    SharedServiceSelectionAll        SharedServiceSelectionPolicy = "All"
    SharedServiceSelectionLeaderOnly SharedServiceSelectionPolicy = "LeaderOnly"
)
```

cheyang (changes requested):

keps/260-leaderonly-service/README.md
JasonHe-WQ: OK, let's move it under `LeaderWorkerPattern` field

api/workloads/v1alpha2/rolebasedgroup_types.go
```diff
@@ -162,6 +162,17 @@ const (
 	RecreateRoleInstanceOnPodRestart RestartPolicyType = "RecreateRoleInstanceOnPodRestart"
 )
 
+// SharedServiceSelectionPolicy defines the service policy of service per role
```
cheyang, line 165: The KEP and the actual API implementation are not aligned.

The KEP says `RoleSpec.NetworkConfig.TargetPolicy = Full|LeaderOnly`, but the code implements `leaderWorkerPattern.sharedServiceSelection = All|LeaderOnly`. This is not a minor drift — it is a fundamental API design discrepancy.

Before merging, please decide: is this a **leader-worker-specific capability** or a **generic shared-service endpoint selection capability**? If it's the former, the current field placement is acceptable; if it's the latter, putting it under `LeaderWorkerPattern` is too narrow.

Either update the KEP to match the current code, or update the code to match the KEP's more general design. Do not merge in an inconsistent state.

api/workloads/v1alpha2/rolebasedgroup_types.go
```diff
@@ -162,6 +162,17 @@ const (
...
+type SharedServiceSelectionPolicy string
+
+const (
+	// SharedServiceSelectionAll - All pods would be routed to
+	SharedServiceSelectionAll SharedServiceSelectionPolicy = "All"
```
cheyang, line 170: The enum value `All` contradicts the KEP documentation which says the default is `Full`. This creates confusion for users.

Additionally, the CRD definition has no enum constraint — `sharedServiceSelection` is just a plain `string`, so users can pass arbitrary values like `"abc"` and apiserver won't reject them.

Please:
1. Unify the naming (`Full` or `All`, pick one and stick with it in both code and docs)
2. Add `+kubebuilder:validation:Enum=All;LeaderOnly` (or `Full;LeaderOnly`) to the type definition
3. Clarify the relationship between unset/nil and the explicit default value

pkg/reconciler/svc_reconciler.go
```diff
@@ -107,6 +107,10 @@ func (r *ServiceReconciler) constructServiceApplyConfiguration(
 		constants.GroupNameLabelKey: rbg.Name,
 		constants.RoleNameLabelKey:  role.Name,
 	}
+	if role.IsLeaderWorkerPattern() && role.LeaderWorkerPattern.SharedServiceSelection != nil &&
+		*role.LeaderWorkerPattern.SharedServiceSelection == workloadsv1alpha2.SharedServiceSelectionLeaderOnly {
+		selectMap[constants.ComponentNameLabelKey] = "leader"
```
cheyang: Hardcoded `"leader"` magic string instead of using a constant.

The project already defines `LeaderComponentType = "Leader"` in constants, and the Pod label value in `roleinstanceset_reconciler.go:390` is also `"leader"`. But the casing is inconsistent — the constant uses uppercase `"Leader"` while both the reconciler and this selector use lowercase `"leader"`.

If someone refactors the Pod label to use the constant (uppercase), this selector will silently stop matching, and the Service will have zero endpoints.

Suggestion: define a lowercase constant (e.g., `LeaderComponentValue = "leader"`) or unify casing, and use it in both the Service selector and Pod label construction.

pkg/reconciler/svc_reconciler.go
```diff
@@ -107,6 +107,10 @@ func (r *ServiceReconciler) constructServiceApplyConfiguration(
 		constants.GroupNameLabelKey: rbg.Name,
 		constants.RoleNameLabelKey:  role.Name,
 	}
+	if role.IsLeaderWorkerPattern() && role.LeaderWorkerPattern.SharedServiceSelection != nil &&
```
cheyang, line 110: Missing validation for unsupported combinations — KEP says "should reject" but code silently falls back.

The KEP explicitly states: "Unsupported combinations should reject `LeaderOnly` instead of silently falling back to `Full`."

But the current implementation only checks `IsLeaderWorkerPattern()` at reconcile time, and when the condition is not met, it silently falls back to the default selector. This contradicts the KEP's stated behavior and makes misconfiguration hard for users to detect.

Please add at least one of:
- kubebuilder CEL/XValidation on the CRD to reject `LeaderOnly` for non-leaderWorkerPattern roles
- Admission webhook validation
- Controller status condition or event to signal the invalid configuration

Silent fallback is not acceptable when the design document explicitly promises rejection.

config/crd/bases/workloads.x-k8s.io_rolebasedgroups.yaml
```diff
@@ -25335,6 +25335,10 @@ spec:
                           description: LeaderTemplatePatch indicates patching for
                             the leader template.
                           x-kubernetes-preserve-unknown-fields: true
+                        sharedServiceSelection:
```
cheyang, line 25338: The CRD defines `sharedServiceSelection` as a plain `string` with no enum constraint.

Without an enum, apiserver accepts arbitrary values. This is a weak API that will be hard to tighten later once users start relying on it.

Please add an enum constraint matching the valid values (e.g., `All`/`Full` and `LeaderOnly`).

keps/260-leaderonly-service/README.md
```diff
@@ -0,0 +1,290 @@
...
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+This KEP proposes a new role-level networking field,`NetworkConfig.TargetPolicy`, to control which Pods are selected by the existing shared headless Service of a role.
```
cheyang: KEP terminology drift — this section says `NetworkConfig.TargetPolicy` with values `Full`/`LeaderOnly`, but the actual implementation uses `LeaderWorkerPattern.SharedServiceSelection` with values `All`/`LeaderOnly`.

Within the same document, readers encounter `NetworkConfig.TargetPolicy`, `TargetPolicy`, `SharedServiceSelection`, `Full`, and `All` — it's unclear which is the final API.

Please unify the KEP to match the final API decision, or mark clearly which is provisional vs. final.

keps/260-leaderonly-service/README.md
```diff
@@ -0,0 +1,290 @@
...
+
+`Full` keeps the current behavior and remains the default. `LeaderOnly` preserves the existing shared Service name, 
+but changes the Service selector so that only leader Pods are targeted by that Service.
+
+
+## Motivation
+
+RBG currently creates one shared headless Service per role, and the Service selector includes all Pods of that role.
+
+That behavior is acceptable when every Pod behind the Service is a real serving endpoint. But it would make requests for large EP inference engine
```
cheyang: This sentence is grammatically broken: "But it would make requests for large EP inference engine which only leader Pods are expected to accept requests abnormal"

Please rewrite into complete sentences, e.g.: "But it causes problems for large inference engines where only leader Pods should accept requests — routing traffic to worker Pods running dummy servers leads to abnormal behavior."

test/e2e/testcase/v1alpha2/shared_service_selection.go
```diff
@@ -0,0 +1,322 @@
+/*
```
cheyang, line 1: The e2e and unit tests cover the happy path well (selector generation, in-place update, EndpointSlice verification). However, there are no tests for **invalid configurations** that the KEP promises to reject:

- A standalone role with `LeaderOnly` should be rejected
- A non-RoleInstanceSet workload with `LeaderOnly` should be rejected
- An invalid enum value should be rejected
- Explicit `All` → `LeaderOnly` → `All` transitions should also be covered if `All` is retained

Please add negative tests to prove that misconfiguration is actually caught, not silently ignored.

cheyang (changes requested):
Thanks for this PR — the controller implementation is clean and the test coverage for the happy path is solid (selector change + in-place update + EndpointSlice verification).

However, I'm requesting changes because of three blockers that need to be addressed before merge:

**1. KEP / API / implementation are misaligned**
The KEP describes `RoleSpec.NetworkConfig.TargetPolicy = Full|LeaderOnly`, but the code implements `leaderWorkerPattern.sharedServiceSelection = All|LeaderOnly`. This isn't a naming drift — it's a fundamental design discrepancy. Please decide on the final API design and unify both the KEP and the code.

**2. No validation for unsupported combinations**
The KEP explicitly says "Unsupported combinations should reject `LeaderOnly` instead of silently falling back to `Full`." The current implementation only checks `IsLeaderWorkerPattern()` at reconcile time and silently falls back when the condition isn't met. This contradicts the stated design. Please add CRD validation, webhook validation, or at least a controller status condition to signal invalid configuration.

**3. CRD lacks enum constraint**
`sharedServiceSelection` is a plain string with no enum. Users can pass arbitrary values. Please add `+kubebuilder:validation:Enum` and clarify the unset/nil vs. explicit default relationship.

Additional suggestions (non-blocking):
- Use a constant instead of the hardcoded `"leader"` magic string in the Service selector
- Add negative tests for invalid configurations
- Clean up KEP terminology drift and grammar issues
- Decide between "minimal API" (only `LeaderOnly`, nil=default) vs. "explicit policy" (enum with defaulting) and be consistent

pkg/reconciler/svc_reconciler.go
JasonHe-WQ: May I also normalize these in `pkg/reconciler/roleinstanceset_reconciler.go` ?
```
	roleInstanceTemplateConfig.
		WithRestartPolicy(workloadsv1alpha2.RoleInstanceRestartPolicyRecreateOnPodRestart).
		WithComponents(
			workloadsv1alpha2client.RoleInstanceComponent().
				WithName("leader").
				WithServiceName(svcName).
				WithSize(1).
				WithTemplate(leaderTemplateApplyCfg.WithLabels(map[string]string{
					constants.ComponentNameLabelKey: "leader",
					constants.ComponentSizeLabelKey: fmt.Sprintf("%d", *lwp.Size),
				})),
			workloadsv1alpha2client.RoleInstanceComponent().
				WithName("worker").
				WithSize(workerSize).
				WithTemplate(workerTemplateApplyCfg.WithLabels(map[string]string{
					constants.ComponentNameLabelKey: "worker",
					constants.ComponentSizeLabelKey: fmt.Sprintf("%d", *lwp.Size),
				})))
```

keps/260-leaderonly-service/README.md
JasonHe-WQ:
```plaintext
That behavior is acceptable when every Pod behind the Service is a real serving endpoint. But it would make requests for large EP inference engine
which only leader Pods are expected to accept requests abnormal
```
Maybe we need a comma

api/workloads/v1alpha2/rolebasedgroup_types.go
```diff
@@ -323,6 +334,11 @@ type LeaderWorkerPattern struct {
 	// +kubebuilder:pruning:PreserveUnknownFields
 	// +kubebuilder:validation:Schemaless
 	WorkerTemplatePatch *runtime.RawExtension `json:"workerTemplatePatch,omitempty"`
+
+	// SharedServiceSelection indicates the service policy of the role
+	// +optional
+	// +kubebuilder:validation:Enum={All;LeaderOnly}
```
cheyang: The Enum marker uses a semicolon instead of a comma:

```go
// +kubebuilder:validation:Enum={All;LeaderOnly}
```

Should be:

```go
// +kubebuilder:validation:Enum={All,LeaderOnly}
```

The rest of the codebase uses commas (e.g. `{RollingUpdate}`, `{None,RecreateRBGOnPodRestart,...}`). The semicolon causes `controller-gen` to fail with `expected comma, got ";"`, which cascades to every CI job.

After fixing this, please re-run `make manifests` and commit the regenerated CRDs.

keps/260-leaderonly-service/README.md
cheyang: This still references the old API surface:

> This KEP proposes a new role-level networking field, `NetworkConfig.TargetPolicy`...

The actual implementation (and the Design Details section below) uses `LeaderWorkerPattern.SharedServiceSelection`. Please update the Summary to match.

config/crd/bases/workloads.x-k8s.io_rolebasedgroups.yaml
cheyang, line 25338: The generated CRD is missing the `enum` constraint — `sharedServiceSelection` is just `type: string` here. This is a side effect of the broken Enum marker in the Go types. Once the marker syntax is fixed and `make manifests` is re-run, this should get an `enum: [All, LeaderOnly]` entry automatically.

cheyang (changes requested):
Thanks for the update — the reconciler logic and tests are in good shape now, and previous blocker #2 (unsupported-combination validation) is structurally resolved by placing the field under `LeaderWorkerPattern`.

However, the Enum marker syntax error at `rolebasedgroup_types.go:340` is breaking all CI (lint, unit-test, envtest, e2e) — see inline comments.

**Remaining blockers before merge:**

1. Fix the Enum marker: `{All;LeaderOnly}` → `{All,LeaderOnly}` (semicolon → comma)， such as https://github.com/openshift/sriov-network-operator/blob/main/api/v1/sriovnetwork_types.go#L45
2. Re-run `make manifests` to regenerate CRDs with the enum constraint
3. Update the KEP Summary — it still references `NetworkConfig.TargetPolicy` instead of `LeaderWorkerPattern.SharedServiceSelection`

Once these are addressed and CI goes green, this should be ready for another pass.

cheyang:
Implementation is clean — reconciler logic, CEL validation, SSA transitions, and constant alignment are all correct. No blocking issues. Main feedback: KEP doc has a few terminology and grammar issues from the last review round, and the unit test for selector updates only covers one transition direction.

keps/260-leaderonly-service/README.md
```diff
@@ -0,0 +1,290 @@
...
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+This KEP proposes a new role-level networking optional field `SharedServiceSelectionPolicy``, under `RoleSpec.LeaderWorkerPattern`, to control which Pods are selected by the existing shared headless Service of a role.
```
cheyang: The Summary says `SharedServiceSelectionPolicy` but that is the **type** name. The **field** name is `SharedServiceSelection` (or fully qualified: `LeaderWorkerPattern.SharedServiceSelection`). The previous review round specifically asked for the field name. There is also a stray double-backtick at the end of the identifier.

Same issue appears in the Proposal section (~line 143).

keps/260-leaderonly-service/README.md
```diff
@@ -0,0 +1,290 @@
...
+followers that do not run a fully functional serving endpoint do not receive external traffic.
+
+#### Story 2
+
+As an operator building a gateway, I want one stable shared Service per role, but I only want the real serving Pods to appear in that Service's endpoints. 
+This lets the gateway keep using Service-level discovery without routing requests to worker Pods that should only participate in internal execution.
+
+
+#### Story 3
+As a platform engineer, although we do support a pod-level model gateway, e.g. `sgalng model gateway`, we still need a fallback in case the gateway is absent.
```
cheyang: A few doc issues worth a quick pass:
- Line 170: `sgalng` should be `sglang`.
- Lines 113-114: "But it would make requests ... requests abnormal" is grammatically broken.
- Line 117: missing "to be" — "causes requests routed to" should read "causes requests **to be** routed to".
- Line 243: "targeted by in" — stray preposition.

pkg/reconciler/svc_reconciler_test.go
```diff
@@ -94,14 +112,70 @@ func TestServiceReconciler_reconcileHeadlessService(t *testing.T) {
...
+	require.NoError(t, cl.Get(context.Background(), types.NamespacedName{Name: svcName, Namespace: rbg.Namespace}, svc))
+	assert.NotContains(t, svc.Spec.Selector, constants.ComponentNameLabelKey)
+
+	roleRef.LeaderWorkerPattern.SharedServiceSelection = ptr.To(
+		workloadsv1alpha2.SharedServiceSelectionLeaderOnly,
+	)
+	require.NoError(t, reconciler.reconcileHeadlessService(context.Background(), rbg, roleRef))
+
+	require.NoError(t, cl.Get(context.Background(), types.NamespacedName{Name: svcName, Namespace: rbg.Namespace}, svc))
+	assert.Equal(t, "leader", svc.Spec.Selector[constants.ComponentNameLabelKey])
```
cheyang, line 176: This test covers `nil -> LeaderOnly` (adding `component-name` to the selector) but not the reverse direction. Consider adding a third step that sets `SharedServiceSelection` back to nil and asserts `component-name` is removed. The e2e test covers this direction, so both directions are tested across the suite — but having it in the unit test too would catch regressions faster.

Also a nit: prefer `string(constants.LeaderComponentType)` over the hardcoded `"leader"` for consistency with the reconciler.

api/workloads/v1alpha2/rolebasedgroup_types.go
```diff
@@ -162,8 +162,20 @@ const (
...
+	// SharedServiceSelectionAll - All pods would be routed to
+	SharedServiceSelectionAll SharedServiceSelectionPolicy = "All"
+
+	// SharedServiceSelectionLeaderOnly - The headless service would only target at the leaders
+	SharedServiceSelectionLeaderOnly SharedServiceSelectionPolicy = "LeaderOnly"
+)
+
 // RoleSpec defines the specification for a role in the group
 // +kubebuilder:validation:XValidation:rule="!(has(self.standalonePattern) && has(self.leaderWorkerPattern))",message="standalonePattern and leaderWorkerPattern are mutually exclusive"
+// +kubebuilder:validation:XValidation:rule="!has(self.leaderWorkerPattern) || !has(self.leaderWorkerPattern.sharedServiceSelection) || self.leaderWorkerPattern.sharedServiceSelection != 'LeaderOnly' || !has(self.annotations) || !('rbg.workloads.x-k8s.io/role-workload-type' in self.annotations) || self.annotations['rbg.workloads.x-k8s.io/role-workload-type'] == 'workloads.x-k8s.io/v1alpha2/RoleInstanceSet'",message="leaderWorkerPattern.sharedServiceSelection=LeaderOnly is only supported for RoleInstanceSet + leaderWorkerPattern"
```
cheyang, line 178: The six-clause OR chain is correct but hard to parse at a glance. A short comment above the marker explaining the intent would help future readers, e.g.:
```
// SharedServiceSelection=LeaderOnly is only valid when the effective workload type is RoleInstanceSet.
```

Syspretor (approved):
/lgtm

PR #280, opened 2026-04-21
fix(rbgsa-controller): preserve and initialize readyReplicas

sebest:
## Summary
- bug: `status.readyReplicas`  is reset to nil when status is updated
- preserve RBGSA status fields by building SSA status patches from the latest live adapter
- prevent later status applies from pruning `status.readyReplicas` back to nil
- initialize `status.readyReplicas` to `0` when a bound adapter is created before the parent RBG has role status

## Testing
- go test ./internal/controller/workloads -count=1


sebest:
Addressed the open review comment in `3af25e88`.

The RBGSA status patch helper now builds a status-only apply configuration instead of reusing the full object apply configuration, so `/status` SSA patches no longer carry `spec`. I also added a unit test to assert the status-patch helper excludes `spec`.

Syspretor (approved):
/lgtm

PR #12, opened 2025-09-02
bugfix: Fix the permission issue that rbgs controller cannot create rbg

bcfre:
The rgbs controller needs to create/delete rbg objects according to the rgbs declaration. Currently, the controller lacks the create/delete permission for rgb. The specific error is as follows
`{"level":"ERROR","time":"2025-09-02T01:12:47.892Z","caller":"workloads/rolebasedgroupset_controller.go:97","message":"create rbg failed.","controller":"workloads-rolebasedgroupset","controllerGroup":"workloads.x-k8s.io","controllerKind":"RoleBasedGroupSet","RoleBasedGroupSet":{"name":"rbgs-test","namespace":"default"},"namespace":"default","name":"rbgs-test","reconcileID":"7fbf093e-4a74-494e-a182-0f1efef5ffb7","rbgset":{"name":"rbgs-test","namespace":"default"},"error":"create rbg error: rolebasedgroups.workloads.x-k8s.io is forbidden: User \"system:serviceaccount:rbgs-system:rbgs-rbgs-controller-sa\" cannot create resource \"rolebasedgroups\" in API group \"workloads.x-k8s.io\" in the namespace \"default\"","stacktrace":"sigs.k8s.io/rbgs/internal/controller/workloads.(*RoleBasedGroupSetReconciler).Reconcile.func1\n\t/workspace/internal/controller/workloads/rolebasedgroupset_controller.go:97"}`

Reproduction method: 
1. In the project root directory, execute `make deploy` to deploy rbg. 
2. Execute `kubectl apply -f examples/rbgs/rbgs-base.yaml` to deploy the test case. 
3. Use `kubectl logs -n rbgs-system <controller-pod>` to view the error log. 
4. Run `make undeploy` to clean up the test resources.

PR #281, opened 2026-04-21
refactor: remove workload field in v1alpha2

Syspretor:

fixes #XXXX

## Checklist


cheyang (approved):
The KEP doc examples use `strings.Split(workloadType, "/")` to parse the annotation value, but the actual implementation uses `strings.LastIndex`. The two approaches are logically equivalent for well-formed values, but the KEP should reflect the actual code to avoid confusion for future readers. Also worth noting in the KEP: the `apiVersion/kind` format (e.g. `apps/v1/StatefulSet`) is inherently ambiguous since `/` is used as both the group/version separator and the annotation delimiter — `LastIndex` correctly handles multi-`/` apiVersions, but the format choice and its parsing strategy should be explicitly documented.

Consider updating the KEP code examples to use `strings.LastIndex` consistently, and adding a brief note on why this format was chosen over alternatives (e.g. JSON `{apiVersion, kind}`) and how edge cases (malformed values, empty values) are handled.

keps/workload-compatibility-mode/README.md
```diff
@@ -0,0 +1,260 @@
...
+func convertRoleV2ToV1alpha1(src *v2.RoleSpec, dst *RoleSpec) error {
+    dst.Name = src.Name
+    dst.Labels = src.Labels
+    dst.Annotations = src.Annotations
+    
+    // Restore workload from annotation
+    if src.Annotations != nil {
+        workloadType := src.Annotations[constants.RoleWorkloadTypeAnnotationKey]
+        if workloadType != "" {
+            parts := strings.Split(workloadType, "/")
```
cheyang, line 159: The KEP doc examples use `strings.Split(workloadType, "/")` to parse the annotation value, but the actual implementation uses `strings.LastIndex`. The two approaches are logically equivalent for well-formed values, but the KEP should reflect the actual code to avoid confusion for future readers.

Also worth noting: the `apiVersion/kind` format is inherently ambiguous since `/` is used as both the group/version separator and the annotation delimiter. `LastIndex` handles this correctly, but the format choice and its parsing strategy should be explicitly documented in the KEP.

api/workloads/v1alpha2/rolebasedgroup_types.go
```diff
@@ -228,6 +222,36 @@ type RoleSpec struct {
...
+	// Parse "apiVersion/kind" format
+	idx := strings.LastIndex(wt, "/")
+	if idx > 0 {
+		return WorkloadSpec{
+			APIVersion: wt[:idx],
+			Kind:       wt[idx+1:],
+		}
+	}
+	// Return default if parsing fails
+	return WorkloadSpec{
```
cheyang, line 249: The `WorkloadSpec` struct is no longer serialized to CRD but still lives here as an in-memory helper. Consider adding a comment clarifying its role (e.g. `// WorkloadSpec is an internal helper; not serialized to CRD schema`), or moving it to a separate file so it does not appear alongside CRD-backed type definitions.

api/workloads/v1alpha2/roletemplate_validation.go
```diff
@@ -105,7 +105,8 @@ func validateRoleTemplateFields(
 		}
 
 		// Defense-in-depth: CRD validates this, but controller validates as well.
-		if role.Workload.Kind == "InstanceSet" {
+		workloadSpec := role.GetWorkloadSpec()
+		if workloadSpec.Kind == "InstanceSet" {
```
cheyang, line 109: Good defensive check — but since the annotation is meant for conversion compatibility only, could someone still set `InstanceSet` via annotation on a new v1alpha2 RBG? If so, would this validation correctly block it? Worth adding a brief comment here to clarify the intent.

pkg/dependency/dependency_test.go
```diff
@@ -300,18 +300,13 @@ func TestDefaultDependencyManager_CheckDependencyReady(t *testing.T) {
 				{
```
cheyang, line 300: role1 has `Workload` removed but no annotation set — it will default to RoleInstanceSet. Was the original intent for role1 to use a different workload type (e.g. StatefulSet)? If so, the annotation should be added here. If RoleInstanceSet is the intended default, this is fine, but worth noting for clarity.

cheyang:
## PR #281 Code Review Summary

Overall direction is sound: using a role-level annotation to carry v1alpha1 workload compatibility info keeps the v1alpha2 schema clean, and the default `RoleInstanceSet` is appropriate for new objects. Below are findings organized by priority.

---

## 🔴 Blocking Issues

### 1. Annotation may leak to Pod / downstream workloads

`RoleSpec.Annotations` is typically propagated to downstream STS/Deployment/Pod metadata during rendering. `rbg.workloads.x-k8s.io/role-workload-type` is **control-plane internal semantics** and should not be inherited by Pods — it would appear as a misleading annotation on user workloads.

**Recommendation**:
- Explicitly filter out this key before rendering downstream resources;
- State in the KEP: "This annotation is not propagated to downstream resources";
- Add a corresponding unit test.

### 2. `dst.Annotations` may share underlying map with `src`

In `convertRoleV1alpha1ToV2`, if upstream code does `dst.Annotations = src.Annotations` (shallow copy), then `dst.Annotations[key]=` **mutates the source object passed from apiserver**. Writing to an apiserver-provided object is a dangerous anti-pattern. Always deep-copy or copy-on-write when mutating annotations.

### 3. `GetWorkloadType` / `GetWorkloadSpec` lack nil-receiver protection

These are public methods called directly at multiple sites (`dependency.go`, `pod_controller.go`, `svc_reconciler.go`, etc.) as `role.GetWorkloadSpec()`. A nil `RoleSpec` pointer would cause a panic. Add:

```go
if r == nil {
    return constants.RoleInstanceSetWorkloadType
}
```

---

## 🟡 Important Issues

### 4. `convertRoleV2ToV1alpha1` leaves `dst.Workload` empty when annotation is absent

For pure v1alpha2 objects (no annotation), `dst.Workload` remains empty, producing an invalid v1alpha1 object. While v1alpha2 is the Hub in normal lifecycle, edge cases can trigger this (old v1alpha1 clients reading a v1alpha2-originated object, `kubectl convert`, or test code). Two-line fallback is cheap:

```go
if dst.Workload.APIVersion == "" || dst.Workload.Kind == "" {
    ws := src.GetWorkloadSpec()
    dst.Workload.APIVersion = ws.APIVersion
    dst.Workload.Kind = ws.Kind
}
```

Alternatively, explicitly document in the KEP: "v1alpha1 API does not guarantee correct reading of pure v1alpha2 objects."

### 5. Annotation format `apiVersion/kind` ambiguity

`apiVersion` itself contains `/` (e.g. `leaderworkerset.x-k8s.io/v1/LeaderWorkerSet`), so `/` as delimiter requires `LastIndex`. Current implementation is correct, but:
- **KEP doc uses `strings.Split`** which would split `leaderworkerset.x-k8s.io/v1` into fragments — inconsistent with actual code. Please unify to `LastIndex`.
- Add a comment at the constant declaration: "apiVersion may contain `/`; parsing must use LastIndex."

### 6. Malformed annotation value silently falls back to default

A user writing a wrong annotation value gets silently treated as `RoleInstanceSet` with no error or warning — behavior may diverge far from intent with zero signal. Recommend:
- Validate format in admission webhook or `validateRoleTemplateFields`;
- Or at minimum, log a warning via `logger.Error` + `recorder.Eventf(Warning, ...)`.

### 7. Missing conversion round-trip unit tests

The KEP "Test Plan" lists conversion tests, but this PR has no incremental changes to `rolebasedgroup_conversion_test.go`. Please add table-driven tests for:
- v1alpha1 Workload → v2 annotation;
- v2 annotation → v1alpha1 Workload;
- v2 no annotation → v1alpha1 Workload (relates to #4 fallback behavior);
- v2 malformed annotation → defined behavior.

### 8. Incomplete CRD breaking change upgrade strategy

CRD yaml deletes the `workload` field outright. For existing clusters:
- v1alpha2 objects persisted in etcd with `workload` will have it pruned on next write, losing information;
- User YAML / helm charts that still write `spec.roles[*].workload` get silently pruned with no warning.

KEP "Upgrade/Downgrade Strategy" only covers v1alpha1↔v1alpha2. Please supplement:
- Migration plan for existing v1alpha2 objects (conversion webhook or script to map old `workload` to annotation);
- Release note explicitly marking **breaking change**;
- Update docs/examples to remove `workload:` usage.

---

## 🟢 Nits / Minor

### 9. `GetWorkloadSpec()` default values hardcoded

`"workloads.x-k8s.io/v1alpha2"` and `"RoleInstanceSet"` duplicate `constants.RoleInstanceSetWorkloadType`. Parse the constant instead to avoid drift.

### 10. Wrapper comments outdated

- `BuildLeaderWorkerRole` comment still says "Sets the Workload to LeaderWorkerSet" — update.
- `BuildStandaloneRole` same issue.

### 11. KEP vs implementation discrepancies

- KEP Goals: "Remove `Workload` field **and `WorkloadSpec` struct**" — but `WorkloadSpec` is retained as an internal helper. Change to "Remove `Workload` field only."
- KEP code examples use `strings.Split`, actual code uses `strings.LastIndex` (see #5).
- Add comment to `WorkloadSpec`: `// WorkloadSpec is an internal helper; not serialized to CRD schema.`

### 12. `InstanceSet` defensive check policy

KEP says "annotation is for conversion only, users should not set it," but code does not prevent users from manually writing the annotation. Clarify policy:
- Admission webhook blocks user-written annotation (enforce KEP intent); OR
- Acknowledge annotation as a supported override mechanism with full validation.

### 13. Minor refactoring

- `LeaderWorkerRoleWrapper.WithWorkload` and `StandaloneRoleWrapper.WithWorkload` are identical — extract shared function or embed common struct.
- `cmd/cli/cmd/llm/run.go` removed `// TODO: Remove workload field after PR #261` — mention #261 in PR description.

---

## ✅ What Looks Good

- `LastIndex` for parsing `apiVersion/kind` is correct (vs `Split`);
- `GetWorkloadType()` / `GetWorkloadSpec()` abstraction keeps controller changes minimal (one-line `role.Workload → role.GetWorkloadSpec()`);
- Table-driven tests in `helper_test.go` cover nil/empty/each kind/malformed;
- CRD yaml, applyconfiguration, CLI, E2E wrappers all updated in one pass — no "delete Go but forget CRD" gaps.

---

## Recommended Fix Order

1. **#3 annotation leak → #2 map sharing → #1 nil receiver** — correctness risks, #3 most urgent (affects every reconcile cycle).
2. Add conversion round-trip tests (**#7**), decide on **#4** fallback vs KEP clarification.
3. Fix **#5** (unify KEP to `LastIndex`) and **#6** (malformed value strategy).
4. Update KEP and comments (**#10**, **#11**, **#12**).
5. Release note + migration guide (**#8**).

cheyang (approved):
/lgtm
/approve

PR #285, opened 2026-04-22
chore: add e2e cases for inplace update

Syspretor:

fixes #XXXX

## Checklist

### DIFF
--- a/test/e2e/testcase/v1alpha2/inplace_update.go
+++ b/test/e2e/testcase/v1alpha2/inplace_update.go
@@ -1,0 +1,788 @@
+/*
+Copyright 2026 The RBG Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha2
+
+import (
+	"time"
+
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	workloadsv1alpha2 "sigs.k8s.io/rbgs/api/workloads/v1alpha2"
+	"sigs.k8s.io/rbgs/test/e2e/framework"
+	testutils "sigs.k8s.io/rbgs/test/utils"
+	wrappersv2 "sigs.k8s.io/rbgs/test/wrappers/v1alpha2"
+)
+
+// newTestImage returns a different image for testing updates
+const updatedImage = "nginx:stable-alpine"
+
+func RunInPlaceUpdateTestCases(f *framework.Framework) {
+	ginkgo.Describe("inplace update", func() {
+		registerBasicInplaceUpdateTestCases(f)
+		registerLeaderWorkerInplaceUpdateTestCases(f)
+		registerGracePeriodInplaceUpdateTestCases(f)
+		registerMultiReplicaInplaceUpdateTestCases(f)
+		registerRoleTemplateInplaceUpdateTestCases(f)
+	})
+}
+
+// nolint:gocyclo
+func registerBasicInplaceUpdateTestCases(f *framework.Framework) {
+	ginkgo.It("standalone pattern with InPlaceIfPossible should perform inplace update when only image changes", func() {
+		rbg := wrappersv2.BuildBasicRoleBasedGroup("e2e-inplace-basic", f.Namespace).
+			WithRoles([]workloadsv1alpha2.RoleSpec{
+				wrappersv2.BuildStandaloneRole("role-1").
+					WithRollingUpdate(workloadsv1alpha2.RollingUpdate{
+						Type: workloadsv1alpha2.InPlaceIfPossibleUpdateStrategyType,
+					}).
+					Obj(),
+			}).Obj()
+
+		f.RegisterDebugFn(func() { dumpDebugInfo(f, rbg) })
+
+		gomega.Expect(f.Client.Create(f.Ctx, rbg)).Should(gomega.Succeed())
+		f.ExpectRbgV2Equal(rbg)
+
+		// Get initial pod and record UID
+		podList := &corev1.PodList{}
+		gomega.Eventually(func() bool {
+			err := f.Client.List(f.Ctx, podList, client.InNamespace(f.Namespace),
+				client.MatchingLabels{
+					"rbg.workloads.x-k8s.io/group-name": rbg.Name,
+					"rbg.workloads.x-k8s.io/role-name":  "role-1",
+				})
+			return err == nil && len(podList.Items) > 0
+		}, testutils.Timeout, testutils.Interval).Should(gomega.BeTrue())
+
+		initialPodUID := podList.Items[0].UID
+
+		// Update image
+		updateRbgV2(f, rbg, func(rbg *workloadsv1alpha2.RoleBasedGroup) {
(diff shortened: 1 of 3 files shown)


cheyang (commented):
Thanks for adding comprehensive e2e coverage for in-place update. The test structure is solid and follows existing patterns well.

Two things need attention before this can merge:

**1. Merge conflict** — the PR currently conflicts with the base branch. Please rebase.

**2. CI failure (not flaky)** — the `leaderWorker pattern with multiple replicas should perform inplace update` test times out at 150s (line ~567 in `inplace_update.go`). The in-place update for a 2-replica LeaderWorkerSet with size=2 (4 pods total) does not converge. This looks like a real bug in the test expectation or controller behavior rather than a timing fluke. Worth investigating whether the controller correctly handles in-place updates across multiple replicas of a LeaderWorkerSet.

A few minor observations below via inline comments.

test/e2e/testcase/v1alpha2/inplace_update.go
```diff
@@ -0,0 +1,788 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	workloadsv1alpha2 "sigs.k8s.io/rbgs/api/workloads/v1alpha2"
+	"sigs.k8s.io/rbgs/test/e2e/framework"
+	testutils "sigs.k8s.io/rbgs/test/utils"
+	wrappersv2 "sigs.k8s.io/rbgs/test/wrappers/v1alpha2"
+)
+
```
cheyang, line 33: Stale comment: this says `newTestImage returns a different image` but it's a const, not a function. Trivial nit — just remove or fix the comment.

test/e2e/testcase/v1alpha2/inplace_update.go
```diff
@@ -0,0 +1,788 @@
...
+			for _, pod := range podList.Items {
+				if !initialPodUIDs[pod.UID] {
+					return false
+				}
+				if pod.Spec.Containers[0].Image != updatedImage {
+					return false
+				}
+			}
+			return true
+		}, testutils.Timeout, testutils.Interval).Should(gomega.BeTrue())
```
cheyang, line 567: This is the test that times out in CI. With 2 replicas x size 2 = 4 pods, the Eventually block waits for all pods to retain UIDs and update images. Please check whether the controller actually completes the in-place update for all pods in this scenario, or if there's a reconciliation issue that leaves some pods stuck.

test/e2e/testcase/v1alpha2/inplace_update_template.go
```diff
@@ -0,0 +1,681 @@
...
+	type metadata struct {
+		Labels map[string]string `json:"labels"`
+	}
+	type patch struct {
+		Metadata metadata `json:"metadata"`
+	}
+	b, err := json.Marshal(patch{Metadata: metadata{Labels: labels}})
+	if err != nil {
+		return &runtime.RawExtension{}
+	}
```
cheyang, line 679: Silently returning an empty RawExtension on marshal failure will produce confusing test results. Consider using `gomega.Expect(err).ShouldNot(gomega.HaveOccurred())` or `ginkgo.GinkgoT().Fatal(err)` to surface the error.

test/e2e/testcase/v1alpha2/inplace_update.go
```diff
@@ -0,0 +1,788 @@
...
+
+		f.RegisterDebugFn(func() { dumpDebugInfo(f, rbg) })
+
+		gomega.Expect(f.Client.Create(f.Ctx, rbg)).Should(gomega.Succeed())
+		f.ExpectRbgV2Equal(rbg)
+
+		// Get initial pods (2 groups * 2 pods each = 4 pods)
+		podList := &corev1.PodList{}
+		gomega.Eventually(func() bool {
+			err := f.Client.List(f.Ctx, podList, client.InNamespace(f.Namespace),
```
cheyang, line 528: This test timed out in CI after 150s. The CI logs show that the in-place image update succeeds (`SuccessfulUpdatePodInPlace`), but then the default `RecreateInstanceOnPodRestart` restart policy triggers pod deletion and recreation with new UIDs. The assertion that all pod UIDs remain unchanged then never passes.

You likely need to either:
1. Set a restart policy on the LeaderWorker role that does not recreate instances on container restart, or
2. Adjust the assertion to account for the fact that in-place updates combined with `RecreateInstanceOnPodRestart` will produce new pod UIDs.

The same issue may affect the single-replica leaderWorker inplace tests that currently pass - they might be passing only due to timing differences with fewer pods.

test/e2e/testcase/v1alpha2/inplace_update.go
```diff
@@ -0,0 +1,788 @@
...
+		// Get initial pods
+		podList := &corev1.PodList{}
+		gomega.Eventually(func() bool {
+			err := f.Client.List(f.Ctx, podList, client.InNamespace(f.Namespace),
+				client.MatchingLabels{
+					"rbg.workloads.x-k8s.io/group-name": rbg.Name,
+					"rbg.workloads.x-k8s.io/role-name":  "role-1",
+				})
+			return err == nil && len(podList.Items) == 3
+		}, testutils.Timeout, testutils.Interval).Should(gomega.BeTrue())
```
cheyang, line 468: The test is titled `leaderWorker pattern with InPlaceIfPossible should recreate pods when updating leader/worker patches`, but the test body only updates the base template image and then asserts that all pod UIDs remain the same (lines 505-514) - i.e., it verifies *in-place* update, not recreation.

If the intent is to test recreation when patches change, the test should modify `LeaderTemplatePatch` or `WorkerTemplatePatch` with non-image fields and assert UIDs change. If the intent is to verify that in-place update still works when leader/worker patches are present, rename the test to match that behavior.

test/e2e/testcase/v1alpha2/inplace_update.go
```diff
@@ -0,0 +1,788 @@
...
+
+		initialPodUID := podList.Items[0].UID
+		initialImage := podList.Items[0].Spec.Containers[0].Image
+
+		// Update image + env (should not work with InPlaceOnly)
+		updateRbgV2(f, rbg, func(rbg *workloadsv1alpha2.RoleBasedGroup) {
+			rbg.Spec.Roles[0].StandalonePattern.Template.Spec.Containers[0].Image = updatedImage
+			rbg.Spec.Roles[0].StandalonePattern.Template.Spec.Containers[0].Env = []corev1.EnvVar{
+				{Name: "TEST_ENV", Value: "test"},
+			}
```
cheyang, line 289: Both grace period tests (standalone at line 289 and leaderWorker at line 359) set `GracePeriodSeconds: 5` but only verify the final state - image updated and pod UID unchanged. They never assert that the update was *delayed* by the grace period.

Consider adding a `Consistently` check right after the update to confirm the pod image has NOT changed yet during the grace window, before the `Eventually` check that confirms the update eventually completes. Without this, these tests are functionally identical to the basic inplace update tests.

test/e2e/testcase/v1alpha2/inplace_update_template.go
```diff
@@ -0,0 +1,681 @@
...
+					}).
+					Obj(),
+			}).Obj()
+
+		f.RegisterDebugFn(func() { dumpDebugInfo(f, rbg) })
+
+		gomega.Expect(f.Client.Create(f.Ctx, rbg)).Should(gomega.Succeed())
+		f.ExpectRbgV2Equal(rbg)
+
+		// Get initial pod
```
cheyang, line 623: `buildRawExtensionPatch` here is essentially a duplicate of `buildRawPatch` in `test/wrappers/v1alpha2/role_wrapper.go`. Consider reusing the existing helper or extracting both into a shared test utility package.

test/e2e/testcase/v1alpha2/inplace_update.go
cheyang, line 33: The comment `// newTestImage returns a different image for testing updates` reads like a function doc comment, but decorates a `const` declaration. Something like `// updatedImage is a different image tag used for in-place update test scenarios` would be clearer.

cheyang:
Hi @Syspretor — this PR has fallen into merge conflict against `main`. Could you rebase onto the latest `main` and force-push? Once conflicts are resolved and CI is green, we can pick the review back up. Thanks!

PR #286, opened 2026-04-22
fix: deal with injectMetadataSave properly

diw-zw:
## Summary

Fixes two issues in the `injectMetadataSave` helper within `rbgctl llm pull`:

1. **Incomplete `/bin/sh -c` detection** — the previous code only recognized `Command: ["/bin/sh", "-c"], Args: [...]`. It now also handles the variant `Command: ["/bin/sh"], Args: ["-c", ...]` and normalizes it to the standard form so downstream logic stays uniform.
2. **Inconsistent metadata save command between Case 1 and Case 2** — Case 2 (wrap a direct binary) was using a simplified `printf '%s\n'` without generating the `downloadedAt` timestamp via `date` and `sed`. It now reuses the same `metadataCmd` as Case 1, ensuring the `.rbg-metadata.json` file always contains the correct `downloadedAt` value.

## Changes

- `cmd/cli/cmd/llm/pull.go`
  - Expands `isShellC` detection to cover `/bin/sh` with `-c` in `Args`
  - Normalizes the split form into `Command: ["/bin/sh", "-c"]` before appending the metadata command
  - Handles the edge case where `container.Args` is empty
  - Unifies Case 2 to use the same `metadataCmd` (with `date` + `sed`) instead of a shorter variant
- `cmd/cli/cmd/llm/pull_test.go`
  - Updates test comments to reflect that both cases now use `printf '%s\n'` with `date` and `sed`


Syspretor (approved):
/lgtm

PR #293, opened 2026-04-23
feat(cli): allow custom all fields without pre-build model config

diw-zw:
### Ⅰ. Motivation

Previously, `kubectl rbg llm svc run` required a pre-built model config for every model ID. Users deploying custom/private models without a built-in config entry would get an error and could not proceed, even if they knew all the deployment parameters. This PR enables deploying any model via CLI flags alone, without depending on a pre-built model config.

### Ⅱ. Modifications

**`cmd/cli/cmd/llm/svc/run.go`** (+145/-65)

- **Flag-only deployment**: When no model config is found but `--engine` is specified, a minimal config is constructed from flags (`custom` mode) instead of returning an error.
- **New CLI flags**:
  - `--resource` (stringArray, `key=value` e.g. `nvidia.com/gpu=1`) — specify resource requirements directly
  - `--distributed-size` (int32) — multi-node deployment size
  - `--shm-size` (string, e.g. `8Gi`) — shared memory size
- **Unified flag override layer** (`applyFlagOverrides`): All flag values (image, resources, distributed-size, shm-size, env, args) are applied onto `ModeConfig` via a single function, replacing scattered ad-hoc overrides in `buildGenerateOptions`. Flag values always win over config on conflict.

**`cmd/cli/cmd/llm/svc/run_test.go`** (+164/-1)

### DIFF
--- a/cmd/cli/README.md
+++ b/cmd/cli/README.md
@@ -33,7 +33,11 @@ kubectl rbg llm model pull Qwen/Qwen3.5-0.8B
 Deploy a model to your Kubernetes cluster:
 
 ```bash
+# Deploy a model with a pre-built model config
 kubectl rbg llm svc run my-qwen Qwen/Qwen3.5-0.8B
+
+# Or deploy a custom model without a pre-built model config
+kubectl rbg llm svc run my-model org/new-model --engine vllm --resource nvidia.com/gpu=1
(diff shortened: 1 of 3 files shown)


Syspretor (approved):
/lgtm

PR #306, opened 2026-04-25
fix(rbg-controller): override role.Replicas from bound RBGSA

sebest:
## Summary

When a chart that gates `spec.roles[i].replicas` on `Release.IsInstall` is upgraded with the default client-side merge, the rendered manifest no longer carries the field, the apiserver fills it with the CRD default of `1` (`+kubebuilder:default=1` on `RoleSpec.Replicas`), the RBG controller scales the workload to 1, and pods churn. The same class of bug applies to any external mutator that drops the field (`kubectl edit` from a manifest that omits replicas, non-SSA scripts, ArgoCD without `ServerSideApply=true`, etc.).

When a `RoleBasedGroupScalingAdapter` (RBGSA) owns a role's replicas, its `spec.replicas` is the runtime source of truth. This change adds `applyRBGSAReplicasOverride` at the top of `RoleBasedGroupReconciler.Reconcile`:

- For each role with `scalingAdapter.enable=true` whose bound RBGSA has a `spec.replicas` that differs from `role.Replicas`, `role.Replicas` is overridden in-memory to the adapter's value **before any reconcile step reads it**.
- The on-disk RBG field becomes advisory when an autoscaler is bound; the workload is never sized from a stale spec value.
- A Warning Event `ReplicasOverriddenByAdapter` is emitted on each override so operators see external-mutator drift in `kubectl describe`.

The override is gated so callers that intentionally manage replicas without an autoscaler are unaffected: when `scalingAdapter.enable` is false, when the adapter is missing or unbound, or when `adapter.spec.replicas` is nil, the role's own value is kept.

The companion deploy-side recommendation — `helm install/upgrade --server-side` (or ArgoCD `ServerSideApply=true`) — remains the right fix at the protocol layer. This operator change is the safety net for clients that don't use server-side apply.

## Test plan

- [x] `make test` — all unit tests pass, including the two new ones.
  - `TestRoleBasedGroupReconciler_applyRBGSAReplicasOverride` — 7-case table-driven suite covering every gating branch (disabled adapter, missing adapter, unbound adapter, nil adapter spec, equal/different replicas, `role.Replicas` nil), with `record.FakeRecorder` assertions that the Warning event fires exactly when the override does.
  - `TestRoleBasedGroupReconciler_applyRBGSAReplicasOverride_GetError` — interceptor-based test for the apiserver `Get` error path; verifies the wrapped error preserves the original via `%w`.
- [x] `go test -run applyRBGSAReplicasOverride -count=10 ./internal/controller/workloads/...` — stable across 10 runs.
- [x] `make fmt vet` — clean.

## Notes

- No CRD or API changes. No new metrics (consistent with the rest of the rbg operator's observability conventions: Events + structured logs + conditions).
- Reuses existing helpers: `scale.IsScalingAdapterEnable`, `scale.GenerateScalingAdapterName`, `constants.AdapterPhaseBound`. No new abstractions.
- A latency-optimization follow-up is possible (waking the RBGSA controller's writeback immediately on a `spec.replicas` stomp via a new `Watches(&RoleBasedGroup{}, ..., predicate)` on its `SetupWithManager`). It is intentionally not bundled here since it is a separate design discussion (new watch infrastructure) and the in-memory override on the next reconcile already prevents the workload from being sized to the stale value.

### DIFF
--- a/internal/controller/workloads/event.go
+++ b/internal/controller/workloads/event.go
@@ -37,6 +37,14 @@ const (
 	// InvalidGangSchedulingAnnotations is emitted when group-gang-scheduling and
 	// role-instance-gang-scheduling annotations are set simultaneously on the same RBG.
 	InvalidGangSchedulingAnnotations = "InvalidGangSchedulingAnnotations"
+	// ReplicasOverriddenByAdapter is emitted when an external mutator (e.g. a
+	// `helm upgrade` whose chart no longer carries spec.roles[].replicas) sets
+	// spec.roles[i].replicas to a value that disagrees with a bound
+	// RoleBasedGroupScalingAdapter. The controller treats the adapter's value
+	// as authoritative and emits this event each reconcile that performs an
+	// override (so it fires once per drift episode, then stops as the adapter
+	// controller writes the value back).
+	ReplicasOverriddenByAdapter = "ReplicasOverriddenByAdapter"
 )
 
 // rbg-scaling-adapter events
--- a/internal/controller/workloads/rolebasedgroup_controller.go
+++ b/internal/controller/workloads/rolebasedgroup_controller.go
@@ -158,6 +158,10 @@ func (r *RoleBasedGroupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		logger.Info("Finished reconciling", "duration", time.Since(start))
 	}()
 
+	if err := r.applyRBGSAReplicasOverride(ctx, rbg); err != nil {
+		return ctrl.Result{}, err
+	}
+
 	// Step 0: Pre-check validations
 	if err := r.preCheck(ctx, rbg); err != nil {
 		return ctrl.Result{}, err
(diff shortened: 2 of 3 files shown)


cheyang (approved):
Solid defensive fix. The override placement at the top of `Reconcile()` — before `preCheck` and all downstream steps — ensures no reconcile path can accidentally read stale replicas from a mutated spec.

**What I verified:**

- Guard chain (`!enable` → `NotFound` → `!Bound` → `nil replicas` → equality) is correct and ordered defensively. Each exit point is independently safe.
- `ptr.To(*adapter.Spec.Replicas)` avoids pointer aliasing between the adapter object and the role spec.
- Error wrapping uses `%w`, so `errors.Is`/`errors.As` callers won't break.
- The in-memory-only approach is the right call: the controller doesn't own the RBG spec, so writing back would fight the GitOps source. Let the adapter controller reconcile the spec field on its own schedule.
- Test coverage hits all branches (7-case table + error path with interceptor). Event assertions are explicit.

**One minor observation** (non-blocking): the Warning event fires on every reconcile that detects drift, not just the first. If the adapter controller is slow to write back (or is unhealthy), operators may see repeated events. This is arguably fine for observability, but if it becomes noisy in practice, rate-limiting with `EventFilterFunc` or a last-overridden annotation could tame it later.

All CI green. LGTM.

Syspretor (approved):

PR #308, opened 2026-04-26
fix: coordinatedPolicy rollingUpdate maxSkew stall RoleInstanceSet

### LINKED ISSUE #307: [Bug]  CoordinatedPolicy rollingUpdate maxSkew can stall RoleInstanceSet
### Checklist

- [ ] 1. I have searched related issues but cannot get the expected help.
- [ ] 2. The bug has not been fixed in the latest version.
- [ ] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [ ] 5. Please use English, otherwise it will be closed.

### Describe the bug

When using `CoordinatedPolicy` with `rollingUpdate.maxSkew`, a `RoleInstanceSet` rollout can get stuck if the coordinated roles have only one replica.

In this case the RBG controller calculates `partition: 1` for a role with `replicas: 1`. The RoleInstanceSet controller then selects instances to update with `ordinal >= partition`, so no RoleInstance is able for update. The workload stays forever with `currentRevision != updateRevision`.

This is confusing because the `RoleInstanceSet.spec.roleInstanceTemplate` is updated, but the existing `RoleInstance.spec` remains on the old revision. For example, probes and container command are different between RoleInstanceSet and RoleInstance.

### Reproduction

  Create an RBG with two roles, e.g. `prefill` and `decode`, each with `replicas: 1`, and a CoordinatedPolicy:

  ```yaml
  kind: CoordinatedPolicy
  metadata:
    name: <rbg-name>
    namespace: <namespace>
  spec:
    policies:
    - name: auto-all-predictors
      roles:
      - prefill
      - decode
      strategy:
        rollingUpdate:
          maxSkew: 10%
```

Then update the role template, for example change command, model path, probe delay, or any field that creates a new RoleInstanceSet revision.

### Environment

latest 0.7.0 alpha3

JasonHe-WQ:
### Ⅰ. Motivation
fix #307 

### Ⅱ. Modifications
Add another special path for roleinsatnceset with 1 replica and all pods not ready

fixes #XXXX

## Checklist

### DIFF
--- a/internal/controller/workloads/rolebasedgroup_controller.go
+++ b/internal/controller/workloads/rolebasedgroup_controller.go
@@ -1290,6 +1290,9 @@ func calculateNextRollingTarget(
 	// Due to the synchronization between rbg and workloads, the rbg may not get the least updatedReplicas correctly here.
 	// We add maxSkew/2 to the next rolling step to make sure the skew is not too large.
 	nextRollingStep := max(distance2Balance, maxSkew>>1)
+	if nextRollingStep == 0 && updatedReplicas[slowestRole] < desiredReplicas[slowestRole] {
+		nextRollingStep = 1
+	}
 
 	// If balance reached and the fastest role is ready, we add 1 to the next rolling step to make sure the slowest role is updated.
 	if readyReplicas[fastestRole] == desiredReplicas[fastestRole] {
--- a/internal/controller/workloads/rolebasedgroup_controller_test.go
+++ b/internal/controller/workloads/rolebasedgroup_controller_test.go
@@ -331,6 +331,33 @@ func Test_CalculateNextRollingTarget_WithNormalCases(t *testing.T) {
 	}
 }
 
+func Test_CalculateNextRollingTarget_MinProgressWhenNoRoleReady(t *testing.T) {
+	maxSkew := "10%"
+	roles := sets.New[string]("prefill", "decode")
+	desiredReplicas := map[string]int32{
+		"prefill": 1,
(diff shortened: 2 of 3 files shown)


cheyang (approved):
Solid fix for a real stall scenario. The root cause is clear: when `maxSkew` is percentage-based on single-replica roles (e.g. "10%" of 1 replica yields integer maxSkew=1), `maxSkew>>1` evaluates to 0. Combined with `distance2Balance == 0` (both roles at same update ratio), the rolling step computes to 0 and the update never starts.

The existing fallback (`readyReplicas[fastestRole] == desiredReplicas[fastestRole]`) only triggers once the fastest role is fully ready, which creates a chicken-and-egg problem at the beginning of the update when no role has progressed yet.

The 3-line fix is minimal and correct:
- Condition is precise: only fires when the step is 0 AND there are still replicas to update.
- No risk of over-advancing since `min(..., upperBound+1)` clamps the final value.
- No double-counting with the existing `max(nextRollingStep, 1)` guard downstream (uses `max`, not addition).

Test coverage is thorough -- both a unit test exercising the exact stall condition and an E2E test verifying single-replica coordinated rolling updates make forward progress.

All CI green. LGTM.

cheyang (approved):
LGTM. Precise fix for rolling update stall on single-replica roles with good test coverage. All CI green.

PR #312, opened 2026-04-27
ci: add manifest-based E2E test for deploy changes

Syspretor:
Add a separate E2E test job that uses kubectl manifest installation instead of helm. This job only runs when files in deploy/ directory change to avoid redundant test runs.

Key changes:
- Add deploy_changed output to check-changes job
- Add e2e-test-manifest job that installs via deploy/kubectl/manifests.yaml
- Use sed to override image tags for local testing

fixes #XXXX

## Checklist


cheyang (approved):
/lgtm
/approve

PR #316, opened 2026-04-29
feat(rbg): enhanced customComponentsPatten lifcycle management

Syspretor:

fixes #XXXX

## Checklist

### DIFF
--- a/keps/173-enhanced-custom-components-pattern/README.md
+++ b/keps/173-enhanced-custom-components-pattern/README.md
@@ -291,7 +291,7 @@ must receive those endpoints at startup. `CustomComponentsPattern` with the new
 | `startAfter` gate may cause components to wait longer than expected | The gate is evaluated against `RoleInstance.Status.ComponentStatuses.ReadyReplicas >= Size`, which is set only when pods pass readiness probes. This is the intended semantics — if a dependency is slow to become ready, the controller waits correctly without creating unready dependents. |
 | Discovery annotation references a component that doesn't exist | Controller validation (webhook) rejects the RBG at admission time. |
 | Port reference targets a port not defined in port-allocator annotation | Controller returns error in reconcile and surfaces via RBG status condition. |
-| Circular dependency in `startAfter` / `deleteAfter` (A refs B, B refs A) | `DetectCycle` runs DFS on the dependency graph at each reconcile. If a cycle is found, the controller logs an error and falls back to the default parallel mode to avoid a deadlock. |
+| Circular dependency in `startAfter` / `deleteAfter` (A refs B, B refs A) | `DetectCycle` runs DFS on the dependency graph at each reconcile. If a cycle is found, the controller logs an error and falls back to the default parallel mode to avoid a deadlock. Note: cycles involving different components (A→B and B→A) are possible even when self-reference is disallowed; the cycle detection handles this case. |
 
 ## Design Details
 
@@ -341,6 +341,24 @@ ReadyReplicas >= Size  &&  Size > 0
 
 This is evaluated on the status written by the previous reconcile loop, so the gate becomes true
 only after the dependency's pods are genuinely ready (not merely scheduled).
+
+**Why `ReadyReplicas >= Size` instead of `Scheduled`?**
+
+The motivation section describes scenarios where workers crash because the leader's gRPC server
+is not yet ready when they attempt to connect. The `Scheduled` condition only indicates that a pod
+has been assigned to a node — the container may not even be running yet. Using `ReadyReplicas >= Size`
+ensures that all pods of the dependency component have passed their readiness probes, meaning the
+application is actually serving traffic.
+
+**Recommended user-side configuration:**
+
+When using `startAfter` ordering, users should:
+- Define proper `readinessProbe` in the dependency component's pod spec (e.g., TCP socket check on
+  the gRPC port, or an HTTP health endpoint).
+- Ensure the dependent component's startup logic handles transient connection failures gracefully
+  (the dependency may become unavailable after initial readiness due to restarts).
+- Consider using `restartPolicy: OnFailure` at the RBG level if transient failures should trigger
+  pod restart rather than role instance recreation.
 
 **Example: Router starts after Leader and Worker**
 
@@ -518,7 +536,20 @@ constructed as:
 ```
 
 Where `<service-name>` is the headless service name for the role, which already exists as created
-by the `ServiceReconciler`.
+by the `ServiceReconciler`. The relationship between `serviceName` and the role-level headless
+service is:
+- The `InstanceComponent.serviceName` field specifies the service that backs the component's pods.
+- If not explicitly set, the controller uses the role's default headless service (named
+  `s-<rbg-name>-<role-name>` by convention).
+- The FQDN uses this service name as the subdomain, enabling DNS resolution via Kubernetes
+  headless service DNS.
+
+**Cluster domain assumption**: The FQDN format uses `.svc.cluster.local` as the default cluster
+domain suffix. This matches the vast majority of Kubernetes clusters. If your cluster uses a
+custom domain (e.g., `.svc.mycluster.local`), the injected address may not resolve correctly.
+A future enhancement may support configurable cluster domains or use the shorter
+`<pod>.<svc>.<namespace>` format which is domain-agnostic. For now, users on non-default cluster
+domains should consider this limitation or use the shorter address format via init containers.
 
 **Example**: For RBG `llm-prefill`, role `prefill`, instance ordinal `0`, component `leader`, pod
 index `0`, service `s-llm-prefill-prefill`:
@@ -542,6 +573,14 @@ feature (`from` field in `PortReference`). The resolution reads the allocated po
 | `RoleScoped` | `<component-name>.<port-name>` | `leader.leader-grpc` |
 | `PodScoped` | `<pod-name>.<port-name>` | `llm-prefill-prefill-0-leader-0.leader-grpc` |
 
+**`Index` field semantics for different scopes:**
+
+- For `RoleScoped` ports: The `Index` field in `ComponentPortRef` is **ignored** because
+  `RoleScoped` ports have the same value for all pods in the component. The key does not include
+  pod-specific identifiers, so specifying an index is unnecessary.
+- For `PodScoped` ports: The `Index` field is used to construct the pod name portion of the key.
+  It must be within bounds `[0, size-1]` of the referenced component.
+
 When a `ComponentPortRef` is processed, the controller:
 1. Determines whether the target port is `RoleScoped` or `PodScoped` by inspecting the
    port-allocator annotation on the referenced component template.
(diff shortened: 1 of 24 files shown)


cheyang:
Thanks for the KEP — the motivation is solid and the design is well-structured overall. A few things to sort out:

**Blocker:** `kep.yaml` has `authors: "@TBD"` — needs the real handle before merge. Also the PR description template is completely empty (Motivation, Modifications, linked issue are all placeholders); please fill it in so reviewers have context at a glance.

**Design questions worth discussing:**
1. The ordering gate waits for `Scheduled`, but the stated problem is crash loops from services not being ready. Why not `Ready`? What's the recommended user-side pattern?
2. Self-reference is banned as a no-op — but `size > 1` components may legitimately need intra-component pod discovery (allreduce rings, etc.).
3. Env var collision behavior is unspecified (override/append/reject?).
4. `Index` semantics for `RoleScoped` ports — is it just ignored?
5. Port stability across pod recreation — are allocated ports guaranteed stable?

**Minor:** commit message has a typo (`Patten` → `Pattern`), and the FQDN section could clarify the relationship between `serviceName` and the role-level headless service.

keps/173-enhanced-custom-components-pattern/kep.yaml
```diff
@@ -0,0 +1,16 @@
+title: Enhanced CustomComponentsPattern - Component Lifecycle Ordering and Intra-Role Service Discovery
+kep-number: 173
+authors:
```
cheyang, line 3: `authors: "@TBD"` — please replace with the actual GitHub handle before merging. Placeholder metadata shouldn't land on main.

keps/173-enhanced-custom-components-pattern/README.md
```diff
@@ -0,0 +1,537 @@
...
+## Design Details
+
+### Feature 1: Component Lifecycle Ordering
+
+**Creation Order**
+
+Components within `CustomComponentsPattern.components[]` are created in array declaration order.
+The controller iterates the components slice in forward order when constructing `RoleInstance` specs.
+The underlying `RoleInstanceSet` controller then creates component pods in the same order, waiting
+for each component's pods to reach the `Scheduled` phase before proceeding to the next component.
```
cheyang: The ordering gate waits for pods to reach `Scheduled`, but the motivation describes workers crashing because the leader's gRPC server isn't ready yet. `Scheduled` only means the pod landed on a node — the container may not even be running. The risk table mentions readiness probes as a mitigation, but there's no recommended end-to-end pattern here. Could you add a paragraph explaining why `Scheduled` was chosen over `Ready` and what complementary user-side configuration (probes, init containers, restart policy) you expect?

keps/173-enhanced-custom-components-pattern/README.md
```diff
@@ -0,0 +1,537 @@
...
+   b. Append env var {Name: ref.Env, Value: <port>} to all containers
+4. Remove component-discovery annotation from the generated Pod
+   (it is a controller directive, not a pod-level annotation)
+```
+
+**Validation (admission webhook):**
+
+- All `component` names in `ComponentAddressRef` and `ComponentPortRef` must reference a component
+  name that exists in the same `CustomComponentsPattern.components[]` array.
+- A component cannot reference itself (self-reference is a no-op and is rejected to avoid
```
cheyang, line 625: Self-reference is rejected as a no-op, but for components with `size > 1` (e.g. 4 workers), worker-0 discovering worker-1's address is a real use case (ring allreduce, mesh topologies). Is this intentionally out of scope? If so, it would help to state the reasoning — otherwise users will wonder why they can't do intra-component discovery.

keps/173-enhanced-custom-components-pattern/README.md
```diff
@@ -0,0 +1,537 @@
...
+(`PodReconciler.ConstructPodTemplateSpecApplyConfiguration`), after port allocation and before
+writing the final Pod spec. This mirrors the existing `port-allocator` injection flow.
+
+**Processing steps in the Pod reconciler:**
+
+```
+1. Parse component-discovery annotation from pod template
+2. For each AddressRef:
+   a. Compute target pod FQDN from (component, index, role context)
+   b. Append env var {Name: ref.Env, Value: <fqdn>} to all containers
```
cheyang: The processing steps say "Append env var" to all containers, but what happens when the injected env var name already exists in the container spec? Override, append, or reject? This should be defined explicitly to avoid hard-to-debug issues.

keps/173-enhanced-custom-components-pattern/README.md
```diff
@@ -0,0 +1,537 @@
...
+
+Port references are resolved using the same mechanism as the existing port-allocator reference
+feature (`from` field in `PortReference`). The resolution reads the allocated port value from the
+`RoleInstance` annotation.
+
+**Port key formats** (consistent with KEP-171):
+
+| Port scope | Key format | Example |
+|---|---|---|
+| `RoleScoped` | `<component-name>.<port-name>` | `leader.leader-grpc` |
```
cheyang, line 573: The port key format table shows `RoleScoped` keys don't include a pod name, but `ComponentPortRef` still has an `Index` field. Is `Index` ignored when the target port is `RoleScoped`? Clarifying this for both scopes would prevent user confusion.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -159,6 +214,11 @@ func (c *realControl) createPods(ctx context.Context, updateInstance *workloadsv
 			}
 			componentPortConfigs[component.Name] = config
 		}
+
+		// Check if component-discovery injection is needed for this component
+		if componentdiscovery.HasComponentDiscoveryConfig(&component.Template) {
```
cheyang, line 219: This activation check reads only `component.Template.Annotations`, while the lifecycle code uses `componentAnnotation()` which prefers `comp.Annotations` and falls back to the template. The two paths agree today only because `roleinstanceset_reconciler.go` mirrors `component.Annotations` into `tpl.Annotations` before constructing the apply config.

If that mirroring is ever removed, or if a `RoleInstance` is constructed by another path (tests, future reconciler, manual), discovery would silently no-op while lifecycle keeps working. Worth a single helper used by both directives so the source of truth is explicit.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -120,14 +121,67 @@ func (c *realControl) calculateDiffsWithExpectation(ctx context.Context, updateI
...
+						toScaleRoleIDS[rg.Name] = rg.ToScaleIDs
+					}
+				}
+			}
+
+			// Scale in: delete pods for a component only when every component in its deletion
+			// gates (reverse-of-start ∪ explicit deleteAfter) has no remaining excess pods.
+			for _, rg := range prt.Topologies {
+				if rg.ToDeleteIDs.Len() > 0 {
+					if allDependentsDeleted(deletionGates[rg.Name], topoMap) {
```
cheyang, line 165: The deletion gate is exercised inside `calculateDiffsWithExpectation`, which only runs for spec-driven scale-in (user reduces `Spec.Components[].Size`). When the user issues `kubectl delete rbg <name>`, owner-ref garbage collection deletes pods directly and never goes through this code path, so `deleteAfter` is effectively ignored at RBG-level deletion.

If this is intentional, please make the scope explicit in the KEP and `patterns.md`. If `deleteAfter` is meant to apply to RBG deletion as well, a finalizer-driven shutdown path would be needed.

pkg/component-discovery/component_discovery.go
```diff
@@ -0,0 +1,204 @@
...
+	targetPodName := fmt.Sprintf("%s-%s-%d", instance.Name, ref.Component, ref.Index)
+	podScopedKey := fmt.Sprintf("%s.%s", targetPodName, ref.PortName)
+
+	// RoleScoped key format: <component-name>.<port-name>
+	roleScopedKey := fmt.Sprintf("%s.%s", ref.Component, ref.PortName)
+
+	if v, ok := instance.Annotations[podScopedKey]; ok {
+		return v, nil
+	}
+	if v, ok := instance.Annotations[roleScopedKey]; ok {
```
cheyang, line 178: `appendOrReplaceEnv` overwrites any existing env entry with the same name. That matches the intent (controller injects authoritative values), but a user who happens to set `LEADER_ADDR` or `LEADER_GRPC_PORT` in their pod spec gets it overwritten with no signal. Consider an info-level klog when a replace happens — debugging "my env is empty" without it is painful.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -120,14 +121,67 @@ func (c *realControl) calculateDiffsWithExpectation(ctx context.Context, updateI
 	if err != nil {
 		return nil, err
 	}
-	for _, rg := range prt.Topologies {
-		if rg.ToDeleteIDs.Len() > 0 {
-			toDeleteNum += rg.ToDeleteIDs.Len()
-			toDeletePods = append(toDeletePods, rg.ToDeletePod...)
+
+	// Build the per-component dependency graphs from template annotations.
+	startDeps, deleteDeps, err := componentdiscovery.ParseAllComponentDependencies(updateInstance.Spec.Components)
```
cheyang, line 126: `ParseAllComponentDependencies` returns an error on invalid JSON, which fails the whole reconcile. A cycle, on the other hand, is logged and recovered with a parallel-mode fallback. Both are user-config errors of similar severity. If invalid JSON is intentionally fatal, please call that out in the KEP; otherwise consider symmetric "log and fall back to parallel" handling for resilience.

keps/173-enhanced-custom-components-pattern/README.md
```diff
@@ -0,0 +1,702 @@
...
+              spec:
+                containers:
+                  - name: router
+                    image: prefill-router:v1
+```
+
+When an instance `prefill-0` is created, the resulting Pods will have:
+- `worker-{0..3}` Pods env: `LEADER_GRPC_PORT=<allocated-port>`, `LEADER_ADDR=<leader-0-fqdn>`
+
+The `router` pod is created only after both `leader` and `worker` pods have `ReadyReplicas == Size`
```
cheyang, line 272: This sentence (and the Go comment at line 317) say `ReadyReplicas == Size`, while line 339 and the actual implementation use `ReadyReplicas >= Size && Size > 0`. Functionally equivalent for normal scale-up, but the inconsistency is a maintenance trap. Could you align the KEP and the type comment to use `>=` to match `allNamedComponentsReady`?

pkg/reconciler/roleinstance/sync/instance_scale_test.go
```diff
@@ -539,3 +540,90 @@ func TestFailedPodDeletion(t *testing.T) {
 		})
```
cheyang, line 540: The new helpers (`allNamedComponentsReady`, etc.) are covered well, but the gated scale-out / scale-in branches and the cycle-fallback path inside `calculateDiffsWithExpectation` itself are only exercised end-to-end. A focused, table-driven unit test over a few `prt.Topologies` shapes plus a cycle case would catch regressions much earlier than e2e signals.

test/e2e/testcase/v1alpha2/component_ordering.go
```diff
@@ -0,0 +1,254 @@
...
+			ginkgo.By("Verifying router pod creation timestamp is after leader and worker pods")
+			verifyRouterCreatedAfterLeaderWorker(f, rbg)
+
+			// --- Verify componentStatuses reflect correct readyReplicas ---
+			ginkgo.By("Verifying componentStatuses show correct readyReplicas for all components")
+			verifyComponentStatuses(f, rbg)
+		})
+
+		// Case 2: deleteAfter — verify router is cleaned up and full RBG deletion completes
+		ginkgo.It("RBG with deleteAfter should be fully deleted in correct order", func() {
```
cheyang, line 65: The case is named "RBG with deleteAfter should be fully deleted in correct order", but the body only waits for full deletion; there is no per-pod timestamp ordering check. Either tighten the assertion to verify deletion order (analogous to `verifyRouterCreatedAfterLeaderWorker`), or rename the case to something like "RBG with deleteAfter completes deletion cleanly" to avoid overpromising.

pkg/reconciler/roleinstanceset_reconciler.go
```diff
@@ -270,8 +270,30 @@ func (r *RoleInstanceSetReconciler) constructRoleInstanceTemplateByCustomCompone
...
+		// Deep-copy the pod template and merge component-level labels/annotations into it.
+		// This allows users to set controller-directive annotations (e.g. component-depends-on,
+		// port-allocator, component-discovery) at the component level instead of inside
+		// template.metadata, keeping the pod spec clean.
+		tpl := *component.Template.DeepCopy()
+		if len(component.Annotations) > 0 {
+			if tpl.Annotations == nil {
+				tpl.Annotations = make(map[string]string, len(component.Annotations))
+			}
+			for k, v := range component.Annotations {
```
cheyang, line 282: Merging `component.Annotations` straight into `tpl.Annotations` makes every pod carry purely controller-side directives like `component-depends-on` in its metadata. `component-discovery` actively strips itself from the live pod, but `component-depends-on` does not, so the live pod ends up with metadata it does not need. Filtering out known directive keys here (or moving their consumption to a path that reads from `comp.Annotations` directly) would keep pod metadata clean.

cheyang (approved):
/lgtm
/approve

PR #324, opened 2026-05-06
fix(controller): prevent RestartInProgress condition loss due to stal…

Syspretor:
…e cache race

When the PodReconciler triggers a restart (RecreateRBGOnPodRestart), the RestartInProgress condition was being permanently lost from the RBG status after the restart completed. This caused e2e test "standalone role with restartPolicy" to time out waiting for RestartInProgress=False.

Root cause: a read-modify-write race between two controllers sharing the same RBG status subresource.

1) PodReconciler.setRestartCondition() used the caching client to read the
   RBG before Status().Update(). If the informer cache had not yet synced
   the RBG controller's latest status writes (Ready condition,
   RoleStatuses), the pod controller would overwrite those fields with
   stale data.

2) RoleBasedGroupReconciler.updateRBGStatus() patches status via SSA with
   Force=true. It only checked the API reader for RestartInProgress. If
   the pod controller had not yet committed its write, the condition would
   be absent from the patch and SSA Force=true would remove it from API.

Fixes:
- PodReconciler: add apiReader (non-caching) and use it in setRestartCondition to always read the authoritative API state before updating, avoiding informer cache staleness.
- updateRBGStatus: check both the cached rbg AND the API reader for RestartInProgress (defense-in-depth), only falling back to the API reader when the cache does not have it.

Ancillary reconcile error "workload generation not equal to observed generation" is expected and self-resolving: after the pod controller deletes and recreates a Deployment, the Deployment controller needs time to update ObservedGeneration to match Generation.

fixes #XXXX

## Checklist

### DIFF
--- a/internal/controller/workloads/pod_controller.go
+++ b/internal/controller/workloads/pod_controller.go
@@ -48,14 +48,16 @@ import (
 
 // PodReconciler reconciles a Pod object owned by RBG
 type PodReconciler struct {
-	client client.Client
-	scheme *runtime.Scheme
+	client    client.Client
+	apiReader client.Reader
+	scheme    *runtime.Scheme
 }
 
 func NewPodReconciler(mgr ctrl.Manager) *PodReconciler {
 	return &PodReconciler{
-		client: mgr.GetClient(),
-		scheme: mgr.GetScheme(),
+		client:    mgr.GetClient(),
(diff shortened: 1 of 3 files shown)


cheyang (approved):
/lgtm
/approve

PR #325, opened 2026-05-06
fix: prevent RestartInProgress condition loss during RBG restart

Syspretor:
Fix the e2e test failure "standalone role with restartPolicy" where the RestartInProgress condition disappears from RBG status after a pod-triggered restart completes.

Root Cause
----------
The RestartInProgress condition is lost through a chain of events:

1. PodReconciler sets RestartInProgress=True on RBG status, then calls RecreateWorkload which deletes and recreates the Deployment.

2. RBG controller reconciles (triggered by Deployment change) and calls ConstructRoleStatus. The new Deployment has Generation=1 but ObservedGeneration=0. ConstructRoleStatus detects this mismatch and returns an error: "workload generation not equal to observed generation".

3. constructAndUpdateRoleStatuses receives the error and returns immediately WITHOUT calling updateRBGStatus. The defense-in-depth code that preserves RestartInProgress (from cache or API reader) never executes.

4. This error-retry cycle repeats until the Deployment controller observes the new generation. By the time the RBG controller finally succeeds, the RestartInProgress condition may already be gone from the informer cache.

Changes
-------

1. CRD: add x-kubernetes-list-type: map to all Conditions fields

   Added +listType=map and +listMapKey=type kubebuilder markers to Conditions fields across all 5 CRD types:
   - api/workloads/v1alpha1/rolebasedgroup_types.go
   - api/workloads/v1alpha1/rolebasedgroupset_types.go
   - api/workloads/v1alpha2/rolebasedgroup_types.go
   - api/workloads/v1alpha2/rolebasedgroupset_types.go
   - api/workloads/v1alpha2/coordinatedpolicy_types.go

   Regenerated CRDs with controller-gen v0.17.2. This changes SSA
   behavior from atomic (whole-list replacement) to map (per-entry
   merge by type key), preventing the RBG controller SSA patch
   from overwriting conditions it does not own.

2. ConstructRoleStatus: treat generation mismatch as non-fatal

   pkg/reconciler/deploy_reconciler.go
   pkg/reconciler/sts_reconciler.go
   pkg/reconciler/roleinstanceset_reconciler.go

   Generation mismatch (Status.ObservedGeneration < Generation) is a
   transient state that resolves automatically when the workload
   controller observes the latest spec. Previously it was returned
   as an error, causing constructAndUpdateRoleStatuses to bail out
   before calling updateRBGStatus. Changed to log a V(1) warning and
   return (status, false, nil).

3. constructAndUpdateRoleStatuses: always call updateRBGStatus

   internal/controller/workloads/rolebasedgroup_controller.go

   Removed the "if updateStatus || readyConditionMissing" guard.
   updateRBGStatus is now called on every reconcile, ensuring the
   defense-in-depth RestartInProgress preservation code runs every
   time. The SSA patch is efficient, with map-type conditions it only
   touches fields that have actually changed.

Defense-in-Depth (previous commit)
----------------------------------
The PodReconciler uses apiReader (non-caching) + RetryOnConflict + Status().Update() to set RestartInProgress. updateRBGStatus checks both the cached RBG and the API server to find and preserve the condition before applying the SSA patch.

fixes #XXXX

## Checklist


cheyang (approved):
/lgtm
/approve

PR #329, opened 2026-05-08
fix(rbg): handle inactive pod states (Evicted/Failed) for RBG

Syspretor:
…recreation

Previously, RBG controller only triggered recreation when container restarted or pod was deleted. This missed cases where pods directly entered Failed/Succeeded phase (e.g., Evicted, UnexpectedAdmissionError, Unschedulable).

Changes:
- Add pod inactive detection functions using native Kubernetes IsPodActive:
  - PodBecameInactive: detect state transition from active to inactive
  - IsPodEvicted: detect eviction (supports DisruptionTarget condition)
  - IsPodUnexpectedAdmissionError: detect admission errors
  - IsPodFailedSchedule: detect scheduling failures
  - GetPodInactiveReason: return human-readable inactive reason

- Extend Pod controller trigger conditions:
  - podToRBG now triggers on pod becoming inactive (!IsPodActive)
  - predicate UpdateFunc only captures active→inactive transitions

- Fix RoleInstance controller shouldRecreateInstance:
  - Use kubecontroller.IsPodActive instead of only checking DeletionTimestamp
  - Correctly detect Failed/Evicted pods as inactive

- Add unit tests for all new detection functions (27 test cases)

- Add e2e test cases for inactive pod handling:
  - Evicted pod with RecreateRBGOnPodRestart
  - Failed pod with RecreateRoleInstanceOnPodRestart
  - Inactive pod with RestartPolicy=None (replacement pod creation)
  - Predicate state transition verification

KEP-173: inactive-pod-handling

fixes #XXXX

## Checklist

### DIFF
--- a/internal/controller/workloads/pod_controller.go
+++ b/internal/controller/workloads/pod_controller.go
@@ -256,22 +256,18 @@ func (r *PodReconciler) podToRBG(ctx context.Context, obj client.Object) []recon
 		return []reconcile.Request{}
 	}
 
-	// Original trigger conditions
+	// Dimension 1 triggers ONLY: container restart or pod deletion
+	// Pod Failed (Dimension 2) is handled by RoleInstance Controller through normal reconciliation
 	containerRestarted := utils.ContainerRestarted(pod)
 	podDeleted := utils.PodDeleted(pod)
 
-	// New trigger condition: Pod became Failed (excluding Succeeded per KEP Non-Goals)
-	// Uses native Kubernetes IsPodActive for consistency with ReplicaSet/StatefulSet behavior
-	// Succeeded pods are explicitly excluded as they represent normal completion
-	podBecameFailed := pod.Status.Phase == corev1.PodFailed && pod.DeletionTimestamp == nil
-
-	// Only trigger if any condition is met
-	if !containerRestarted && !podDeleted && !podBecameFailed {
+	// Only trigger if any Dimension 1 condition is met
+	if !containerRestarted && !podDeleted {
 		return []reconcile.Request{}
 	}
 
 	logger := log.FromContext(ctx).WithValues("Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
-	logger.V(1).Info("Processing Pod event for reconciliation", "containerRestarted", containerRestarted, "podDeleted", podDeleted, "podBecameFailed", podBecameFailed)
+	logger.V(1).Info("Processing Pod event for reconciliation", "containerRestarted", containerRestarted, "podDeleted", podDeleted)
 
 	var rbg workloadsv1alpha2.RoleBasedGroup
 	// Use apiReader (non-caching) to get the latest RBG status for RestartInProgress check.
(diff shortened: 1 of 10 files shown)


internal/controller/workloads/pod_controller.go
```diff
@@ -314,8 +327,14 @@ func (r *PodReconciler) SetupWithManager(mgr ctrl.Manager, options controller.Op
...
 				_, newExist := newPod.Labels[constants.GroupNameLabelKey]
-				return oldExist && newExist
 
+				if !oldExist || !newExist {
+					return false
+				}
+
+				// Detect Pod state transition from active to inactive
+				// Only trigger on state transition to avoid duplicate triggers on every status update
+				return utils.PodBecameInactive(oldPod, newPod)
```
cheyang: This predicate now only passes `active → inactive` transitions. However, a container restart event keeps the pod in `Running` phase — `IsPodActive` returns true for both old and new pod, so `PodBecameInactive` returns false and the event is silently dropped.

This means the existing `RecreateRBGOnPodRestart` + container restart path is broken: the restart count increment never reaches `podToRBG`.

Consider allowing container restart events through as well:
```go
return utils.PodBecameInactive(oldPod, newPod) || containerRestartCountChanged(oldPod, newPod)
```
Or revert to the broader predicate (`return oldExist && newExist`) and let `podToRBG` do the filtering — the downside is more events enqueued, but it preserves correctness.

pkg/utils/pod_utils.go
```diff
@@ -88,3 +94,100 @@ func ContainerRestarted(pod *corev1.Pod) bool {
...
+	if !podutil.IsPodTerminal(pod) {
+		return false
+	}
+	if pod.Status.Phase != corev1.PodFailed {
+		return false
+	}
+	// Check DisruptionTarget condition (K8s 1.22+)
+	// Native reasons: PreemptionByScheduler, TerminationByKubelet
+	for _, cond := range pod.Status.Conditions {
+		if cond.Type == corev1.DisruptionTarget {
```
cheyang: This only checks condition type presence but not `cond.Status == corev1.ConditionTrue`. In practice DisruptionTarget is usually only added when disruption occurs, but a defensive check on Status would be more robust.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -239,21 +241,23 @@ func shouldRecreateInstance(instance *workloadsv1alpha2.RoleInstance, pods []*v1
...
 		expectedPodCount := getExpectedPodCount(instance)
-		// Calculate active pods (exclude those being deleted).
-		// Pods with DeletionTimestamp != nil are in Terminating state and should not
-		// be counted as active, otherwise the controller may fail to detect pod deletion.
+		// Calculate active pods using native IsPodActive.
+		// Pods with Phase=Failed/Succeeded or DeletionTimestamp!=nil are inactive.
 		activeCount := 0
 		for _, p := range pods {
-			if p.DeletionTimestamp == nil {
+			if kubecontroller.IsPodActive(p) {
```
cheyang: `IsPodActive` treats both `Failed` and `Succeeded` pods as inactive. The KEP Non-Goals state "Do not handle Pod Succeeded state (normal completion)", but this change will trigger recreate when a pod succeeds normally. Is that intentional for RBG workloads, or should Succeeded pods be excluded here?

cheyang (approved):
The refactor cleanly separates the two dimensions of Pod restart handling. The original blocking issue (predicate dropping container restart events) is properly fixed — `ContainerRestartCountChanged` in the predicate solves it elegantly while also reducing unnecessary enqueues.

A few minor notes on the latest iteration:

1. `ContainerRestartCountChanged` uses index-based comparison of `ContainerStatuses` between old and new Pod. In practice Kubernetes guarantees stable ordering for a given pod spec, so this is safe — but it's worth noting in a comment that this relies on stable ordering from kubelet.

2. `ContainerRestartCountChanged` doesn't have a dedicated unit test in `pod_utils_test.go`. It's covered indirectly by the predicate test in `pod_controller_test.go`, but a direct unit test with edge cases (mismatched lengths, patio-runtime skip, etc.) would be good for completeness.

3. `PodBecameInactive` and `PodBecameFailed` are both defined but `PodBecameFailed` doesn't appear to be called anywhere in the current code. If it's reserved for future use, consider marking it with a comment; otherwise it could be removed to reduce dead code.

Overall the Dimension 1/2 separation is well-reasoned and the tests are thorough. LGTM.

PR #331, opened 2026-05-08
feat: change default restartPolicy to RecreateRoleInstanceOnPodRestart

Syspretor:
…t for all patterns

Previously, the default restartPolicy was different for each pattern:
- StandalonePattern: None
- CustomComponentsPattern: None
- LeaderWorkerPattern: RecreateRoleInstanceOnPodRestart

This change unifies the default to RecreateRoleInstanceOnPodRestart for all patterns, providing consistent failure handling behavior across deployment types.

Updates:
- pkg/reconciler/roleinstanceset_reconciler.go: modify default values for StandalonePattern and CustomComponentsPattern
- doc/reference/api.md: add default value documentation
- doc/features/failure-handling.md: document default policy
- examples/basic/rbg/patterns/*.yaml: add comments for default values

fixes #XXXX

## Checklist

### DIFF
--- a/pkg/reconciler/roleinstanceset_reconciler.go
+++ b/pkg/reconciler/roleinstanceset_reconciler.go
@@ -151,11 +151,10 @@ func (r *RoleInstanceSetReconciler) constructRoleInstanceSetApplyConfiguration(
 	}
 
 	// 1. construct role instance configuration
-	var restartPolicy workloadsv1alpha2.RoleInstanceRestartPolicyType
-	if role.RestartPolicy == workloadsv1alpha2.RecreateRoleInstanceOnPodRestart {
-		restartPolicy = workloadsv1alpha2.RoleInstanceRestartPolicyType(workloadsv1alpha2.RecreateRoleInstanceOnPodRestart)
-	} else {
-		restartPolicy = workloadsv1alpha2.NoneRoleInstanceRestartPolicy
+	// Default to RecreateRoleInstanceOnPodRestart for all patterns when not explicitly set.
+	restartPolicy := role.RestartPolicy
+	if restartPolicy == "" {
+		restartPolicy = workloadsv1alpha2.RecreateRoleInstanceOnPodRestart
 	}
 	roleInstanceTemplateConfig := workloadsv1alpha2client.RoleInstanceTemplate().
 		WithRestartPolicy(restartPolicy)
@@ -285,7 +284,6 @@ func (r *RoleInstanceSetReconciler) constructRoleInstanceTemplateByCustomCompone
 			}
 		}
 		roleInstanceTemplateConfig.
-			WithRestartPolicy(workloadsv1alpha2.NoneRoleInstanceRestartPolicy).
 			WithComponents(workloadsv1alpha2client.RoleInstanceComponent().
 				WithName(component.Name).
 				WithServiceName(svcName).
@@ -369,7 +367,6 @@ func (r *RoleInstanceSetReconciler) constructRoleInstanceTemplateByLeaderWorkerP
 
 	workerSize := utils.NonZeroValue(*lwp.Size - 1)
 	roleInstanceTemplateConfig.
-		WithRestartPolicy(workloadsv1alpha2.RoleInstanceRestartPolicyRecreateOnPodRestart).
 		WithComponents(
 			workloadsv1alpha2client.RoleInstanceComponent().
 				WithName("leader").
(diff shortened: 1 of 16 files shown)


cheyang (commented):
Agree with the Gemini review: the hardcoded `WithRestartPolicy` calls in `constructRoleInstanceTemplateByCustomComponentsPattern` (line 293) and `constructRoleInstanceTemplateFromStandalonePattern` (line 425) overwrite the restartPolicy that `constructRoleInstanceSetApplyConfiguration` already set based on `RoleSpec`. This means a user explicitly setting `restartPolicy: None` gets silently ignored. The fix is to remove these `WithRestartPolicy` calls and let the caller-set value take effect — which still achieves the PR goal because the default path in the caller (lines 154-166) falls through to `RecreateRoleInstanceOnPodRestart` when `role.RestartPolicy` is unset.

One additional concern: this is a behavioral change for existing deployments. If a cluster has running RoleInstanceSets under standalone or customComponents patterns that never specified a restartPolicy, they were implicitly `None` (no auto-restart). After this change, the next reconciliation will switch them to `RecreateRoleInstanceOnPodRestart`. This should at least be documented as an upgrade note so operators know existing workloads may start auto-recreating role instances on pod restarts where they previously did not.

pkg/reconciler/roleinstanceset_reconciler.go
```diff
@@ -290,7 +290,7 @@ func (r *RoleInstanceSetReconciler) constructRoleInstanceTemplateByCustomCompone
 			}
 		}
 		roleInstanceTemplateConfig.
-			WithRestartPolicy(workloadsv1alpha2.NoneRoleInstanceRestartPolicy).
+			WithRestartPolicy(workloadsv1alpha2.RoleInstanceRestartPolicyRecreateOnPodRestart).
```
cheyang: This `WithRestartPolicy(...)` call unconditionally overrides whatever the dispatch logic (lines 155-166) already set on `roleInstanceTemplateConfig`.

The dispatch logic correctly handles all cases:
- User omits `restartPolicy` → defaults to `RecreateRoleInstanceOnPodRestart`
- User sets `restartPolicy: None` → sets `NoneRoleInstanceRestartPolicy`
- User sets `RecreateRBGOnPodRestart` or `RecreateRoleInstanceOnPodRestart` → sets `RecreateRoleInstanceOnPodRestart`

By overriding here, if a user explicitly specifies `restartPolicy: None` on a CustomComponentsPattern role, their intent is silently ignored.

Suggested fix: remove the `WithRestartPolicy(...)` call from all three pattern functions and let the dispatch logic be the single source of truth. This achieves the PR's goal (unified default) without breaking user-specified overrides:

```go
roleInstanceTemplateConfig.
    WithComponents(workloadsv1alpha2client.RoleInstanceComponent().
        WithName(component.Name).
        ...
```

The same issue exists in `constructRoleInstanceTemplateByLeaderWorkerPattern` (line 377) and should be fixed together.

pkg/reconciler/roleinstanceset_reconciler.go
```diff
@@ -416,7 +416,7 @@ func (r *RoleInstanceSetReconciler) constructRoleInstanceTemplateFromStandaloneP
 	}
 
 	roleInstanceTemplateConfig.
-		WithRestartPolicy(workloadsv1alpha2.NoneRoleInstanceRestartPolicy).
+		WithRestartPolicy(workloadsv1alpha2.RoleInstanceRestartPolicyRecreateOnPodRestart).
```
cheyang: Same override issue as in the CustomComponentsPattern function above. This is the direct cause of the e2e failure: the test creates a standalone role with `RecreateRBGOnPodRestart`, but this line forces the instance-level policy to `RecreateRoleInstanceOnPodRestart` regardless, creating a conflict between the two restart mechanisms that prevents the system from converging within the test timeout.

Removing this `WithRestartPolicy(...)` call (and relying on the dispatch logic) would fix both the override bug and the e2e test.

cheyang (changes requested):
The e2e test `[v1alpha2] standalone role with restartPolicy` is failing (timed out after 150s) due to this change. Main branch e2e tests pass, confirming this is a regression introduced by this PR.

The root cause is a design conflict: the dispatch logic at lines 155-166 already correctly sets the restart policy based on user input, but each pattern function then unconditionally overrides it. See inline comments for details and a suggested fix.

cheyang (approved):
/lgtm
/approve

PR #332, opened 2026-05-09
feat: add annotation to exclude components from triggering restart po…

Syspretor:
…licy

Add ComponentNoTriggerRestartPolicyAnnotationKey to allow specific components in customComponentsPattern to be excluded from triggering the role's restart policy. When this annotation is set to "true" on a component's pod template, pod restart/delete events from this component will NOT trigger RecreateRoleInstanceOnPodRestart or RecreateRBGOnPodRestart actions.

This is useful for auxiliary components (e.g., monitoring, logging sidecars) whose failures should not affect the main workload.

Usage:
  template:
    metadata:
      annotations:
        rbg.workloads.x-k8s.io/component-no-trigger-restart: "true"

fixes #XXXX

## Checklist


internal/controller/workloads/pod_controller.go
```diff
@@ -263,6 +263,15 @@ func (r *PodReconciler) podToRBG(ctx context.Context, obj client.Object) []recon
 	logger := log.FromContext(ctx).WithValues("Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
 	logger.V(1).Info("Processing Pod event for reconciliation")
 
+	// Check if pod has the restart-trigger-policy annotation set to "Ignore"
+	// If so, this pod's restart/delete events should not trigger the role's restart policy
+	if pod.Annotations != nil {
+		if pod.Annotations[constants.RestartTriggerPolicyAnnotationKey] == constants.RestartTriggerPolicyIgnore {
+			logger.V(1).Info("Pod has restart-trigger-policy annotation set to Ignore, skipping restart policy trigger")
+			return []reconcile.Request{}
```
cheyang: This check only prevents triggering `RecreateRBGOnPodRestart`. It does not cover `RecreateRoleInstanceOnPodRestart`.

The `podToRBG()` function already returns empty at line 299 when `RestartPolicy != RecreateRBGOnPodRestart`, so this annotation check is only effective for that one policy.

However, `RecreateRoleInstanceOnPodRestart` is enforced in a completely different path: `pkg/reconciler/roleinstance/sync/instance_scale.go` → `shouldRecreateInstance()`. That function iterates all pods in the instance checking `ContainerRestarted(pod)` and pod deletion count, with no annotation check.

Since `RecreateRoleInstanceOnPodRestart` is the default (and most common) policy, the annotation won't work for the majority of use cases as currently documented.

To fully implement the documented behavior, `shouldRecreateInstance()` also needs to skip pods with this annotation — both in the container-restart loop and in the `activeCount < expectedPodCount` deletion detection. Unit tests for both paths (`podToRBG` and `shouldRecreateInstance`) should be added as well.

cheyang (changes requested):
The annotation check only covers the `RecreateRBGOnPodRestart` path. The `RecreateRoleInstanceOnPodRestart` path (which is the default and more common policy) is not covered — see inline comment for details.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -258,16 +260,41 @@ func shouldRecreateInstance(instance *workloadsv1alpha2.RoleInstance, pods []*v1
...
 	for _, p := range pods {
+		if hasTriggerPolicyIgnore(p) {
+			continue
+		}
+		// Check if any Pod is in Failed phase (excluding pods being deleted)
 		if p.Status.Phase == v1.PodFailed && p.DeletionTimestamp == nil {
 			return true
 		}
+		// Check if any container has restarted
+		if containerRestarted(p) {
```
cheyang, line 372: `containerRestarted(p)` runs on all pods including Succeeded ones (via the `inactivePods` list) and pods being deleted. The sibling PodFailed check above has a `DeletionTimestamp == nil` guard but this doesn't.

A Succeeded pod that had a container restart during its lifetime (e.g. init crash-loop before eventual completion) would incorrectly trigger instance recreation. Same for a pod mid-termination.

Suggest:
```go
if p.DeletionTimestamp == nil && p.Status.Phase != v1.PodSucceeded && containerRestarted(p) {
    return true
}
```

doc/features/failure-handling.md
```diff
@@ -58,12 +45,59 @@ spec:
                 image: nginx:latest
 ```
 
+## Excluding Components from Restart Policy Trigger
+
+You can prevent specific pods from triggering the role's restart policy by setting an annotation on the pod template. This is useful for auxiliary components (e.g., monitoring, logging sidecars) whose failures should not affect the main workload. The annotation works with any deployment pattern (standalonePattern, leaderWorkerPattern, or customComponentsPattern).
```
cheyang, line 50: This claims the annotation works with "any deployment pattern" including LeaderWorkerPattern, but LeaderWorkerPattern maps `RecreateRoleInstanceOnPodRestart` to LWS `RecreateGroupOnPodRestart` and delegates entirely to the LWS controller (`lws_reconciler.go:286`). The LWS controller doesn't check this annotation.

Should be scoped to customComponentsPattern and standalonePattern.

cheyang (changes requested):
Two issues to fix before merge:

1. `containerRestarted()` needs DeletionTimestamp + phase guards to match the adjacent PodFailed check — without them, Succeeded pods with RestartCount > 0 and terminating pods incorrectly trigger instance recreation.

2. Docs overstate pattern scope — the annotation only takes effect for customComponentsPattern and standalonePattern. LeaderWorkerPattern delegates restart handling to the LWS controller (`lws_reconciler.go:286`) which doesn't check this annotation.

The annotation design is solid, test coverage is comprehensive, and the previous review feedback about RecreateRoleInstanceOnPodRestart coverage has been properly addressed.

cheyang (approved):
/lgtm
/approve

PR #333, opened 2026-05-09
fix(controller): fix stateful RoleInstanceSet surge rolling update

NoobDream2568:
### Ⅰ. Motivation

`RoleInstanceSet` (stateful mode) supports `maxUnavailable` and `partition` in its rolling-update strategy, but `maxSurge` is silently ignored — every rollout proceeds in lock-step with `maxUnavailable` and never provisions an availability buffer. This PR implements `maxSurge` for stateful rolling updates, aligned with K8s native semantics.

It also closes three correctness gaps in the existing rolling-update loop:

- A rollout where every base instance is unhealthy still allocated surge slots, wasting expensive GPU capacity to "buffer" instances that are not contributing to availability.
- Pushing a new template while a previous rollout was still in flight left stale-revision surge slots that were never recycled.
- Triggering a rolling update very shortly after creation could cascade into all base instances being deleted and recreated, momentarily breaking the `replicas` invariant.

### Ⅱ. Modifications

The stateful rolling-update loop is restructured into a four-phase pipeline that shares a single `topology` value as the source of truth for ordinal-range, surge sizing, partition, and budget decisions:

| Phase | Responsibility |
|---|---|
| A. observe | resolve `currentRevision` / `updateRevision`, `ApplyRevision`, build `topology`, refresh the per-UID unhealthy-since map |
| B. scale & identity | partition into in-range + condemned, fill empty slots, `processReplica` (create) and `processCondemned` (delete) |
| C. progress update | rolling update with a `free-vs-costly` budget, runs only when `topology.inRollout` |
| D. status & advance | recompute counts, run multi-layer guard before advancing `status.CurrentRevision` |

**Topology computation** (`stateful_instance_set_utils.go`):

```
healthyOldInBase            = healthy old-rev base ords [partition, replicas)
existingValidSurge          = surge ords [replicas, replicas+maxSurge) at the current updateRev
                              (stale-rev surge from a superseded rollout is excluded
                               so it falls outside endOrdinal and gets condemned)
allBaseAtUpdateRevHealthy   = every ord [partition, replicas) is at updateRev,
                              isCreated, isHealthy, !isTerminating, ObservedGen synced
surgeNeeded                 = max(0, healthyOldInBase - maxUnavailable)
activeSurge                 = surgeNeeded
                              if !allBaseAtUpdateRevHealthy && existingValidSurge > activeSurge:
                                  activeSurge = existingValidSurge   // sticky during rollout
                              clamp to maxSurge
endOrdinal                  = replicas + activeSurge
```

`surgeNeeded` only counts healthy old instances — instances that are not currently contributing to availability do not need a surge buffer to be replaced. `existingValidSurge` provides a stickiness floor so already-allocated surge does not thrash as healthy-old shrinks during a rollout. The stickiness is dropped once the partition range is fully at updateRev healthy, so surge ramps back down (this matters when `partition>0`, where `CurrentRevision` does not advance and the next-reconcile "inRollout=false → drop surge" cleanup path is otherwise never reached).

**Progress update** (`progressUpdate` in `stateful_instance_set_control.go`):

```
effectiveBudget    = maxUnavailable + availableSurge
                     (availableSurge = surge ords at updateRev that are healthy & non-terminating)
initialBaseUnavail = base ords currently counted as unavailable

target ord ∈ [partition, endOrdinal)   // includes surge ords so stale-rev
                                       // surge slots get recycled

isFree =  isSurgeSlot                  // surge is extra capacity, not part
                                       // of the maxUnavailable contract
       || isTerminating
       || isStablyUnhealthy(target)    // see "stable-unhealthy gate" below

if !isFree && initialBaseUnavail + newlyUnavail >= effectiveBudget: stop
```

Targets are processed in descending ordinal order, so surge ords are recycled before the controller chips at base. A healthy base instance is never deleted faster than `maxUnavailable + availableSurge` allows.

**Stable-unhealthy gate** (`observeInstanceHealth` / `isStablyUnhealthy` in `stateful_instance_set_control.go`):

The "free-delete unhealthy target" path (the `cleanupUnhealthyReplicas` analogue from K8s Deployment) is gated on a per-instance, per-UID timestamp recorded in an in-memory `sync.Map` keyed by `RoleInstance.UID`:

- Phase A calls `observeInstanceHealth` once per reconcile. For every observed instance: record the first time the controller sees it unhealthy; clear the entry the next time the same UID is observed healthy. UIDs no longer in the input slice are evicted.
- `isStablyUnhealthy(inst)` returns true only when `time.Since(firstUnhealthyObservation) >= stableUnhealthyDuration` (10s).

The map resets on controller restart; the worst case after a restart is that a genuinely-broken instance takes one extra `stableUnhealthyDuration` window before becoming eligible for free-delete. Because the timestamp resets on any healthy observation, transient cache flap (Ready toggling on the order of milliseconds because the informer briefly lags the API) cannot accumulate enough continuous unhealthy time to satisfy the gate, so it cannot trigger a free-delete on an actually-ready base instance. A genuinely broken instance accrues continuously and becomes eligible after the window. 10s is well above typical informer cache lag while keeping the recovery delay for genuinely broken pods small enough to avoid wasting GPU resources for too long.

**Advance guard** (`shouldAdvanceCurrentRevision`) — every layer is required:

1. We are in a rollout (`topology.inRollout`).
2. `partition == 0`. With `partition > 0` the ords below partition are pinned at the "old" revision, so advancing `CurrentRevision` to `updateRev` would lie about their state and break later partition decreases (which rely on `currentRev != updateRev` to be recognized as in-rollout).
3. `set.Status.UpdateRevision == updateRev` AND `set.Status.UpdatedReplicas >= replicas - partition` from the previously persisted status. This forces the "all base updated" observation to survive at least one full reconcile.
4. `updateExpectations` are satisfied (no in-flight in-place updates pending API ack).
5. Observed snapshot for every base ord `[partition, replicas)`: at `updateRev`, `isCreated`, `isHealthy`, `!isTerminating`, and `inst.Status.ObservedGeneration >= inst.Generation`.

`inPlaceUpdateInstance` deliberately does not mutate the in-memory revision label after a successful patch — keeping the snapshot honest is what lets the advance guard rely on it.

**Files touched**

- `pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_utils.go` — adds `topology`, `computeTopology`, `computeMaxSurge`, `computeMaxUnavailable`, `computePartition`, `countHealthyOldInBase`, `countExistingValidSurge`, `allBaseAtUpdateRevHealthy`.
- `pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control.go` — `updateStatefulInstanceSet` reshaped into Phase A–D; `progressUpdate`, `countAvailableSurge`, `collectBaseUnavailable`, `buildUpdateTargets`, `applyTargetUpdate`, `countUnhealthy`, `shouldAdvanceCurrentRevision`, `observeInstanceHealth`, `isStablyUnhealthy` introduced; `rollingUpdateInstances`, `updateInstancesInSequence`, `getMaxUnavailable` (method), `collectUnavailableInstances`, `sortInstancesToUpdate` removed; in-memory revision-label write in `inPlaceUpdateInstance` removed.
- `pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_controller.go` — adds the global `instanceUnhealthySince sync.Map` backing the stable-unhealthy gate.
- `test/wrappers/v1alpha2/role_wrapper.go` — adds `WithRollingUpdate` helper used by e2e tests.
- `test/e2e/testcase/v1alpha2/ris.go` and `test/e2e/e2e_test.go` — registers the new e2e specs.

fixes #XXXX

Unit tests in `stateful_instance_set_control_test.go`:

- `TestComputeTopology` — table-driven; covers no-rollout, paused, surge sizing under healthy-old base, all-unhealthy base (`activeSurge=0`), `maxUnavailable` partially / fully covering `surgeNeeded`, partition, stale-rev surge condemnation, surge stickiness across reconciles, partition rollout completion (sticky surge dropped so it can be condemned), and partition decrease (newly-exposed ord triggers surge allocation).
- `TestProgressUpdateBudget` — uses an in-package `fakeInstanceObjectManager` and `fakeInplaceControl` to assert which instances get deleted; covers free-vs-costly budgeting under each combination of `maxUnavailable`, `availableSurge`, and `partition`, plus chained rollout with stale-rev surge slots inside `[replicas, endOrdinal)` (Pending and Healthy variants), and the stable-unhealthy gate (a fresh-unhealthy instance is NOT free-deleted; one that has been seeded as stably-unhealthy IS).
- `TestShouldAdvanceCurrentRevision` — exercises each guard layer individually plus the happy path; covers `partition>0` (must not advance), `ObservedGeneration` lag, terminating instances, and prior-status concurrence.
- `TestObserveInstanceHealthAndIsStablyUnhealthy` and `TestIsStablyUnhealthyEdgeCases` — verify the per-UID timestamp lifecycle: unhealthy is recorded on first observation, cleared on healthy observation, evicted when the UID disappears, and the gate only fires once `stableUnhealthyDuration` has elapsed.

E2E tests in `test/e2e/testcase/v1alpha2/ris.go`:

- **`[RoleInstanceSet] surge maintains full readiness during rolling update`** — `replicas=4, maxSurge=2, maxUnavailable=0`; trigger a rolling update and assert ready count never drops below `replicas` during the entire rollout, surge instances are created and then cleaned up, final total returns to 4.
- **`[RoleInstanceSet] surge with maxUnavailable allows faster update with minimum readiness`** — `replicas=4, maxSurge=2, maxUnavailable=2`; observe the expected intermediate state (`total=6, ready=2`, only ords 0,1 ready), then progress to all-ready and surge cleanup.
- **`[RoleInstanceSet] large maxSurge is capped at replicas count`** — `replicas=2, maxSurge=4`; assert total instance count never exceeds `replicas + replicas = 4` (the sizing cap kicks in when `maxSurge > replicas`).
- **`[RoleInstanceSet] partition decrease progressively rolls more ordinals`** — `replicas=2, maxSurge=0, maxUnavailable=1, partition=1`; trigger a template change → only ord 1 rolls, ord 0 stays at the original revision throughout (`Consistently` for 20s); then lower partition to 0 → ord 0 picks up the new revision. Total instance count stays at 2 the entire time (no surge involved).

1. Deploy the latest RBG controller.
2. Create any RBG with a role that has `replicas=4`, `maxSurge=2`, `maxUnavailable=0`, and a pod startup probe with at least 5s `initialDelaySeconds` so that pods stay NotReady long enough to observe intermediate states.
3. Wait for all pods to become Ready.
4. Update the role (e.g. change the startup probe delay or any pod template field) to trigger a rolling update.
5. Observe that:
    - Surge instances are created during the rolling update (instance count = replicas + surge = 6).
    - The number of Ready instances never drops below `replicas` (stays = 4 when maxUnavailable=0).
    - After the update completes, surge instances are cleaned up and the instance count returns to replicas.
6. Verify partition behavior: set `partition=2` and observe that:
    - Only ordinals >= 2 are updated; ordinals 0,1 keep old revision and stay ready throughout.
    - Surge instances are created for the partition range and cleaned up once partition-range instances are stable.
    - Final state: ordinals 0,1 on old revision, ordinals 2,3 on new revision, no surge instances.

## Checklist

### DIFF
--- a/pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control.go
+++ b/pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control.go
@@ -20,14 +20,13 @@ package statefulmode
 
 import (
 	"context"
-	"fmt"
 	"math"
 	"sort"
 	"time"
 
 	apps "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
-	intstrutil "k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
@@ -46,7 +45,84 @@ const (
 
 	// Realistic value for maximum in-flight requests when processing in parallel mode.
 	MaxBatchSize = 500
+
+	// stableUnhealthyDuration is the minimum CONSECUTIVE time an instance must
+	// be observed unhealthy before progressUpdate is allowed to treat it as
+	// "free" (cleanup-unhealthy semantics — i.e. delete it without consuming
+	// the maxUnavailable budget).
+	//
+	// Without this gate, a transient cache view that briefly reports a ready
+	// base instance as not-ready could let progressUpdate wrongly free-delete
+	// it. Combined with surge that has just become available, this can
+	// cascade into wiping out healthy ready replicas mid-rollout.
+	//
+	// The gate is enforced by tracking, per RoleInstance UID, the first time
+	// THIS controller observed the instance as unhealthy (in
+	// instanceUnhealthySince). The timer resets whenever the same UID is
+	// subsequently observed as healthy — so flapping cache (Ready toggling on
+	// the order of milliseconds) can never accumulate enough continuous time
+	// to satisfy the gate. A genuinely broken instance is observed unhealthy
+	// continuously and becomes eligible after this window.
+	//
+	// 10s is well above the typical informer cache lag (sub-second) while
+	// keeping the recovery delay for genuinely broken pods small enough to
+	// avoid wasting GPU resources for too long.
+	stableUnhealthyDuration = 10 * time.Second
 )
+
+// observeInstanceHealth maintains the instanceUnhealthySince map: for each
+// instance, record the first observation time when unhealthy, and clear the
+// entry when observed healthy. Called once at the top of every reconcile in
+// Phase A so the timestamps reflect the current cache snapshot.
+//
+// Also opportunistically removes entries for UIDs that no longer appear in
+// the supplied instances slice, preventing the map from growing unbounded as
+// instances are recreated (each delete-and-recreate produces a new UID).
+func observeInstanceHealth(instances []*workloadsv1alpha2.RoleInstance) {
+	now := time.Now()
+	live := make(map[types.UID]struct{}, len(instances))
+	for _, inst := range instances {
+		if inst == nil || inst.UID == "" {
+			continue
+		}
+		live[inst.UID] = struct{}{}
+		if isHealthy(inst) {
+			instanceUnhealthySince.Delete(inst.UID)
+			continue
+		}
+		// Unhealthy: record first-observed time if not already recorded.
+		instanceUnhealthySince.LoadOrStore(inst.UID, now)
+	}
+	// Drop entries for instances that have disappeared (different UID after
+	// recreate, or fully deleted).
+	instanceUnhealthySince.Range(func(key, _ interface{}) bool {
+		uid, _ := key.(types.UID)
+		if _, alive := live[uid]; !alive {
+			instanceUnhealthySince.Delete(uid)
+		}
+		return true
+	})
+}
+
+// isStablyUnhealthy returns true when the controller has observed `inst` as
+// unhealthy for at least stableUnhealthyDuration of CONSECUTIVE time. This is
+// the only signal progressUpdate uses to decide whether a base instance is
+// "free" to update without consuming maxUnavailable budget. See
+// stableUnhealthyDuration for the rationale.
+func isStablyUnhealthy(inst *workloadsv1alpha2.RoleInstance) bool {
+	if inst == nil || inst.UID == "" {
+		return false
+	}
+	v, ok := instanceUnhealthySince.Load(inst.UID)
+	if !ok {
+		return false
+	}
+	first, ok := v.(time.Time)
+	if !ok {
+		return false
+	}
+	return time.Since(first) >= stableUnhealthyDuration
+}
 
 // StatefulInstanceSetControlInterface implements the control logic for updating InstanceSets managing Instances
 type StatefulInstanceSetControlInterface interface {
@@ -267,7 +343,22 @@ func (ssc *defaultStatefulInstanceSetControl) getInstanceSetRevisions(
 	return currentRevision, updateRevision, collisionCount, nil
 }
 
-// updateStatefulInstanceSet performs the update function for a InstanceSet managing instances
+// updateStatefulInstanceSet drives a single reconcile pass over the
+// RoleInstanceSet, in four ordered phases that each have a clear contract:
+//
+//	Phase A — observe         compute revisions & topology, build the
+//	                          desired status skeleton.
+//	Phase B — scale & identity ensure ord [start, end) is fully populated:
+//	                          condemn out-of-range, create missing slots,
+//	                          process replicas (creation) & condemned (delete).
+//	Phase C — progress update for in-rollout sets, drive rolling update with
+//	                          a free-vs-costly maxUnavailable budget.
+//	Phase D — status & advance recompute counts, decide whether to advance
+//	                          status.CurrentRevision (multi-layer guard).
+//
+// The four phases share the topology computed in Phase A as the single source
+// of truth for ordinal range, surge sizing, partition, and budget — so there
+// is no second place that recomputes "where surge ends".
 func (ssc *defaultStatefulInstanceSetControl) updateStatefulInstanceSet(
 	ctx context.Context,
 	set *workloadsv1alpha2.RoleInstanceSet,
@@ -276,12 +367,12 @@ func (ssc *defaultStatefulInstanceSetControl) updateStatefulInstanceSet(
 	collisionCount int32,
 	instances []*workloadsv1alpha2.RoleInstance,
 	revisions []*apps.ControllerRevision) (*workloadsv1alpha2.RoleInstanceSetStatus, error) {
+
+	// =====================  Phase A — observe  =====================
 	selector, err := utils.ValidatedLabelSelectorAsSelector(set.Spec.Selector)
 	if err != nil {
 		return set.Status.DeepCopy(), err
 	}
-
-	// get the current and update revisions of the set.
 	currentSet, err := ApplyRevision(set, currentRevision)
 	if err != nil {
 		return set.Status.DeepCopy(), err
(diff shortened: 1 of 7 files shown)


pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control.go
```diff
@@ -464,6 +543,23 @@ func (ssc *defaultStatefulInstanceSetControl) rollingUpdateInstances(
...
+	// When surge instances exist, we can allow more concurrent updates because
+	// surge instances provide an availability buffer. However, only AVAILABLE
+	// surge instances (healthy and not terminating) actually contribute to the
+	// buffer. Using maxSurge directly is too aggressive because surge instances
+	// that are not yet ready don't provide any availability guarantee.
+	// Counting only available surge instances ensures we never update base
+	// instances when the surge buffer is not actually available.
+	maxSurge := computeMaxSurge(set)
+	availableSurgeCount := 0
+	surgeStartIdx := int(*set.Spec.Replicas) // base replica count = surge start index in replicas array
```
cheyang: Potential nil pointer dereference: if `set.Spec.Replicas` is nil, this panics. `computeMaxSurge` handles nil safely and returns 0, but this line runs unconditionally afterward.

`getMaxUnavailable` only nil-checks `Replicas` when `MaxUnavailable` is explicitly set. With both fields unset (defaults), execution reaches here without any guard.

Suggested fix — derive from the ordinal range instead:
```go
_, endOrd, _ := getInstanceSetReplicasRange(set)
surgeStartIdx := endOrd - startOrdinal
```
This also removes the implicit assumption that `startOrdinal == 0`.

pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_utils.go
```diff
@@ -236,6 +236,34 @@ func newVersionedInstance(
 	return instance
 }
 
+// computeMaxSurge returns the maxSurge value for the RoleInstanceSet's update strategy.
+// Returns 0 if MaxSurge is not specified or Replicas is nil.
+func computeMaxSurge(set *workloadsv1alpha2.RoleInstanceSet) int {
+	if set.Spec.UpdateStrategy.MaxSurge == nil || set.Spec.Replicas == nil {
+		return 0
+	}
+	maxSurge, _ := intstrutil.GetScaledValueFromIntOrPercent(
```
cheyang: The error from `GetScaledValueFromIntOrPercent` is discarded here. If a user specifies a malformed `MaxSurge` value (e.g. `"abc"`), this silently returns 0 — no error propagated, no log, no event. Compare with `getMaxUnavailable` which properly returns the error.

Suggestion: either propagate the error (changing the signature to return `(int, error)`) or at minimum log a warning so misconfigurations are visible.

pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control.go
```diff
@@ -303,75 +303,29 @@ func (ssc *defaultStatefulInstanceSetControl) updateStatefulInstanceSet(
...
-	// slice that will contain all Instances such that getOrdinal(instance) < startOrdinal or getOrdinal(instance) >= endOrdinal
+	// Extend ordinal range to accommodate surge instances during rolling updates.
+	// When currentRevision != updateRevision and maxSurge > 0, we create extra
+	// instances beyond the base replica count to provide availability buffer.
+	// The surge count is capped at replicas to avoid creating excessive instances
+	// (e.g., maxSurge=10 with replicas=2 should create at most 2 surge instances).
+	effectiveEndOrdinal := endOrdinal
+	effectiveSurgeCount := computeEffectiveSurgeCount(set)
+	if currentRevision.Name != updateRevision.Name && effectiveSurgeCount > 0 {
+		effectiveEndOrdinal = endOrdinal + effectiveSurgeCount
```
cheyang: Nit: this surge expansion runs regardless of `set.Spec.UpdateStrategy.Paused`. If the rollout is paused, `processReplica` will still create these surge instances (since they appear as empty slots in the replicas array), even though `rollingUpdateInstances` returns immediately without doing any update work. Worth gating on `!Paused` here to avoid creating unnecessary instances during a deliberate pause.

cheyang (commented):
## Review Summary

The overall design is solid — the surge rolling update approach, the `ObservedGeneration` guard against stale informer cache, and removing the in-memory revision label update are all well-reasoned fixes for real race conditions. Unit test coverage is good.

Two items I'd want fixed before merge:

1. **Nil panic in `rollingUpdateInstances`** — `surgeStartIdx := int(*set.Spec.Replicas)` can panic when `Replicas` is nil (see inline comment).
2. **Swallowed error in `computeMaxSurge`** — invalid `MaxSurge` values silently become 0 instead of surfacing the misconfiguration (see inline comment).

Additional non-blocking observations:

- **Surge instances are created even when `Paused=true`**: The `effectiveEndOrdinal` expansion happens before the pause check (which is inside `rollingUpdateInstances`). If the update is paused, surge instances still get created by `processReplica` even though no actual rolling update occurs. Consider gating on `!set.Spec.UpdateStrategy.Paused`.
- **`advanceCurrentRevision` doesn't skip reserved ordinals**: Not an active problem today (reserve ordinals are always empty), but a nil slot from a reserved ordinal would permanently block revision advancement if reserve ordinals are implemented later.
- When `maxUnavailable=0, maxSurge>0` and all surge instances stay unhealthy indefinitely (e.g. broken image), the rolling update stalls completely. Is that the intended behavior, or should there be a fallback/timeout?

CI note: The e2e-test failure is in `inactive_pod.go` (RecreateRoleInstanceOnPodRestart timeout) — unrelated to this PR.

cheyang (approved):
/lgtm
/approve

Syspretor (approved):
/lgtm

PR #334, opened 2026-05-11
fix(controller): clean up stale RoleInstances on stateful/stateless pattern switch

NoobDream2568:
### Ⅰ. Motivation

When a RoleInstanceSet switches between stateful and stateless patterns (by changing the `rbg.workloads.x-k8s.io/role-instance-pattern` annotation), only the annotation on the RoleInstanceSet is updated while existing RoleInstances from the old pattern are not cleaned up. This results in stale instances (and their Pods) remaining running alongside newly created ones, effectively doubling the number of running Pods.

**Root cause:**
- **Stateful → Stateless**: Stateful instances lack the `RoleInstanceOwnerLabelKey` label that the stateless selector relies on, so the stateless reconciler cannot see them and creates new instances alongside.
- **Stateless → Stateful**: Stateless instances have random-ID names (e.g., `test-prefill-7jj4f`) that don't match the ordinal naming pattern (`test-prefill-{N}`), so the stateful reconciler's `isMemberOf` filter rejects them — releasing their OwnerReference but not deleting them.

Example of the issue:
```
kubectl get po
test-prefill-0-leader-0       1/1     Running   0          2m42s   # stale (stateful)
test-prefill-0-worker-0       1/1     Running   0          2m42s   # stale (stateful)
test-prefill-1-leader-0       1/1     Running   0          2m42s   # stale (stateful)
test-prefill-1-worker-0       1/1     Running   0          2m42s   # stale (stateful)
test-prefill-7jj4f-leader-0   1/1     Running   0          119s    # new (stateless)
test-prefill-7jj4f-worker-0   1/1     Running   0          119s    # new (stateless)
test-prefill-jdggk-leader-0   1/1     Running   0          119s    # new (stateless)
test-prefill-jdggk-worker-0   1/1     Running   0          119s    # new (stateless)
```

### Ⅱ. Modifications

- **`internal/controller/workloads/roleinstanceset_controller.go`**: Added `cleanupStaleInstances` and `instanceBelongsToPattern` functions in the dispatch reconciler. Before dispatching to the mode-specific reconciler, the dispatch layer now lists all RoleInstances owned by the set (via `spec.selector.matchLabels` which both stateful and stateless instances carry) and deletes any that don't belong to the current pattern:
  - **Stateless pattern**: instance must have `RoleInstanceOwnerLabelKey: UID` label
  - **Stateful pattern**: instance name must match `{setName}-{ordinal}` pattern
  - If any stale instances are deleted, the reconciler requeues to allow deletions to complete before the new mode reconciler creates fresh instances.
- **`internal/controller/workloads/roleinstanceset_controller_test.go`**: Added unit tests covering 16 scenarios for `instanceBelongsToPattern` (8 cases) and `cleanupStaleInstances` (8 cases), including both switch directions, mixed instances, foreign instances, and no-op cases.

No new annotations, API fields, or CRD changes are required.

**`TestInstanceBelongsToPattern`** — 8 cases:
- Stateless instance with matching `RoleInstanceOwnerLabelKey` label → true
- Stateless instance missing the label → false
- Stateless instance with wrong UID label → false
- Stateful instance with ordinal name matching set name → true
- Stateful instance with ordinal name under empty pattern (default stateful) → true
- Random-ID instance under StatefulPattern → false
- Stateful instance with wrong parent name → false
- Unknown pattern → true (safe default)

**`TestCleanupStaleInstances`** — 8 cases:
- Stateful set with matching instances → no deletion, no requeue
- Stateless set with matching instances → no deletion, no requeue
- Stateful → Stateless switch → stale stateful instances deleted, requeue
- Stateless → Stateful switch → stale stateless instances deleted, requeue
- Mixed instances under StatelessPattern → only stale stateful instances deleted
- Mixed instances under StatefulPattern → only stale stateless instances deleted
- Foreign instances (owned by another set) → not deleted
- No instances → no-op

1. Create an RBG with a role that defaults to Stateful pattern
2. Wait for RoleInstances to be created with ordinal names (e.g., `test-prefill-0`, `test-prefill-1`)
3. Change the `rbg.workloads.x-k8s.io/role-instance-pattern` annotation to `Stateless`
4. Verify that the old stateful RoleInstances are deleted and new stateless instances are created
5. Verify the total number of RoleInstances equals `spec.replicas` (no duplicates)
6. Repeat the test in the reverse direction (Stateless → Stateful)

## Checklist

### DIFF
--- a/internal/controller/workloads/roleinstanceset_controller.go
+++ b/internal/controller/workloads/roleinstanceset_controller.go
@@ -19,10 +19,13 @@ package workloads
 import (
 	"context"
 	"fmt"
+	"regexp"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -73,6 +76,13 @@ func (r *RoleInstanceSetReconciler) Reconcile(ctx context.Context, request recon
 		return reconcile.Result{}, err
 	}
 
+	// Clean up RoleInstances that belong to a different pattern than the current one.
+	// This handles the case where the pattern annotation is changed (stateful ↔ stateless),
+	// leaving behind instances from the old pattern that the new reconciler won't manage.
+	if requeue, err := r.cleanupStaleInstances(ctx, set); err != nil || requeue {
+		return reconcile.Result{Requeue: requeue}, err
+	}
+
 	// Dispatch based on the role instance pattern annotation
 	pattern := constants.InstancePatternType(set.Annotations[constants.RoleInstancePatternKey])
 	switch pattern {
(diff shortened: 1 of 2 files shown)


Syspretor (approved):
/lgtm

PR #337, opened 2026-05-18
fix(rbg): handle inactive pod recreation and clean up dead code

Syspretor:
Two-level Failed pod handling in RoleInstance Controller:
- Level 1: Delete Failed pods so replacements can be created (all policies)
- Level 2: Recreate entire Instance for RecreateRoleInstanceOnPodRestart policy

Pass inactivePods through reconciler chain so Failed pods are visible to shouldRecreateInstance. Remove dead utility functions (PodBecameInactive, PodBecameFailed, IsPodEvicted, etc.) and their tests. Update Pod Controller predicate test to match actual logic. Clean up "Dimension" terminology from code, tests, and KEP.

fixes #XXXX

## Checklist


cheyang (commented):
Thanks for this PR — the two-level Failed Pod handling design and the controller responsibility separation are a clear improvement. I have a few concerns that I think should be addressed before merge.

**Blocking: Predicate simplification causes performance regression and E2E failure**

The predicate change from `ContainerRestartCountChanged(oldPod, newPod)` to `oldExist && newExist` is the most impactful change in this PR, and I think it needs to be revised. Two problems stem from it:

1. **Performance**: Every status update for any RBG-labeled pod now triggers reconciliation, and `podToRBG` calls `r.apiReader.Get` (a non-caching API call) on each one. On clusters with many pods this will be a noticeable bottleneck. The predicate should filter at the source.

2. **E2E regression**: The test `standalone role with restartPolicy` is now failing — `RestartInProgress` never transitions to `False`. With the predicate passing all update events, the Pod Controller receives many more reconciles (including intermediate Terminating states), which likely causes race conditions during the restart flow.

Recommendation: restore the predicate filter with an additional DeletionTimestamp check:
```go
return oldExist && newExist &&
    (utils.ContainerRestartCountChanged(oldPod, newPod) ||
     (oldPod.DeletionTimestamp == nil && newPod.DeletionTimestamp != nil))
```
This also means `ContainerRestartCountChanged` in `pkg/utils/pod_utils.go` should not be removed.

---

**Comment inaccuracy in `shouldRecreateInstance`** (`instance_scale.go:236`)

The comment says "Container restart with RecreateRoleInstanceOnPodRestart is handled by Pod Controller, which triggers RBG-level recreation." But `podToRBG` explicitly returns early when `curRole.RestartPolicy != RecreateRBGOnPodRestart`, so Pod Controller does NOT handle `RecreateRoleInstanceOnPodRestart`. Container restart for that policy is handled by the underlying workload controller (e.g. LWS controller). Suggested correction:
```go
// Container restart with RecreateRoleInstanceOnPodRestart is handled by the
// underlying workload controller (e.g. LWS controller), not the RBG Pod Controller.
```

---

**Non-blocking: strict Phase=Failed check vs indirect activeCount** (`instance_scale.go:249`)

The old `shouldRecreateInstance` used `activeCount < expectedPodCount` as an indirect indicator — it could catch a pod failure even if Phase hadn't updated to Failed yet. The new version strictly checks `p.Status.Phase == v1.PodFailed`. In normal K8s flow this should be equivalent, but there's a subtle edge case if a pod is evicted and the Phase update is delayed. Low risk, just flagging.

---

**Missing end-to-end test** (`instance_scale_test.go`)

`TestFailedPodDeletion` tests the filter logic in isolation but doesn't verify the full `calculateDiffsWithExpectation` flow — specifically that after deleting Failed pods, the next reconcile creates replacements. An integration-level test would add confidence.

---

**Minor cleanups**

- `getExpectedPodCount` in `instance_scale.go:270` is now dead code — the new `shouldRecreateInstance` no longer calls it. Consider removing it.
- The KEP Non-Goals section removed "Do not actively clean up Failed Pods", which is consistent with the new behavior. Consider adding a corresponding entry in Goals to explicitly state the active cleanup rationale (unblocking `hasOrphanPod` detection).

cheyang (approved):

PR #344, opened 2026-05-20
fix(e2e): use RetryOnConflict for LWP update in revision test

Syspretor:
## Summary
- Fix flaky e2e test "semantic comparison can reconcile modifications on the workload" that fails with 409 Conflict
- The test races with the RBG controller: both try to update the LWS object concurrently
- Wrap the LWS drift-update in `retry.RetryOnConflict` to handle the race gracefully

## Test plan
- [ ] Run e2e tests and verify the revision test no longer fails with 409 Conflict
- [ ] Verify other revision tests still pass

### DIFF
--- a/test/e2e/testcase/v1alpha2/revision.go
+++ b/test/e2e/testcase/v1alpha2/revision.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"time"
 
+	"k8s.io/client-go/util/retry"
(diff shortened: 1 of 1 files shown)


cheyang (approved):
LGTM. Idiomatic RetryOnConflict wrapper around the LWS drift-update; matches the pattern used elsewhere in the codebase. CI green.

PR #346, opened 2026-05-21
feat: add --enable-pprof flag for runtime profiling

Syspretor:
## Summary
- Add `--enable-pprof` flag (default `false`) and `--pprof-bind-address` flag (default `:6060`) to the controller
- When enabled, starts a pprof HTTP server for CPU/memory profiling via `go tool pprof`
- Useful for performance debugging and stress testing scenarios

## Usage

```bash
# Run controller with pprof enabled
bin/manager --enable-webhooks=none --enable-pprof=true --pprof-bind-address=:6060

# Collect profiles
curl http://localhost:6060/debug/pprof/heap > heap.prof
curl http://localhost:6060/debug/pprof/profile?seconds=30 > cpu.prof
go tool pprof heap.prof
```

## Test plan
- [x] `make build` succeeds
- [x] `make lint` passes
- [ ] Run controller with `--enable-pprof=true`, verify `curl localhost:6060/debug/pprof/` returns pprof index
- [ ] Run controller without the flag, verify no pprof server starts

### DIFF
--- a/cmd/rbgs/main.go
+++ b/cmd/rbgs/main.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"net"
 	"net/http"
 	"net/http/pprof"
 	"os"
@@ -441,12 +442,17 @@ func main() {
 		os.Exit(1)
 	}
 
+	ctx := ctrl.SetupSignalHandler()
+
 	if enablePprof {
-		startPprofServer(pprofAddr)
+		if err := startPprofServer(ctx, pprofAddr); err != nil {
+			setupLog.Error(err, "unable to start pprof server")
+			os.Exit(1)
+		}
 	}
 
 	setupLog.Info("starting manager")
-	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
+	if err := mgr.Start(ctx); err != nil {
 		setupLog.Error(err, "problem running manager")
 		os.Exit(1)
 	}
@@ -550,9 +556,7 @@ func setupWebhookCertController(mgr ctrl.Manager, result *webhookBootstrapResult
 	return webhookCertReconciler.SetupWithManager(mgr, options)
 }
 
-func startPprofServer(pprofAddr string) {
-	setupLog.Info("Enabling pprof", "addr", pprofAddr)
-
+func startPprofServer(ctx context.Context, pprofAddr string) error {
 	mux := http.NewServeMux()
 	mux.HandleFunc("/debug/pprof/", pprof.Index)
 	mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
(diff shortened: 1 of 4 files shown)


cheyang (commented):
The feature is straightforward and the Helm integration looks correct. Main concern is the shutdown goroutine in `startPprofServer` — it's dead code as written since `context.Background()` never cancels. See inline comments for details and a suggested fix.

cmd/rbgs/main.go
```diff
@@ -519,6 +532,43 @@ func setupWebhookCertController(mgr ctrl.Manager, result *webhookBootstrapResult
...
+		Handler:           mux,
+		ReadHeaderTimeout: 5 * time.Second,
+	}
+
+	setupLog.Info("Starting pprof HTTP server", "addr", pprofServer.Addr)
+
+	go func() {
+		go func() {
+			ctx := context.Background()
+			<-ctx.Done()
```
cheyang: `context.Background()` is never cancelled, so `<-ctx.Done()` blocks forever and this entire shutdown branch is unreachable. The pprof server won't gracefully shut down on SIGTERM.

A clean fix: hoist `ctrl.SetupSignalHandler()` before this call and pass the context in:

```go
ctx := ctrl.SetupSignalHandler()
if enablePprof {
    startPprofServer(ctx, pprofAddr)
}
if err := mgr.Start(ctx); err != nil { ... }
```

Then use that `ctx` here instead of `context.Background()`.

cmd/rbgs/main.go
```diff
@@ -519,6 +532,43 @@ func setupWebhookCertController(mgr ctrl.Manager, result *webhookBootstrapResult
...
+			shutdownCtx, cancelFunc := context.WithTimeout(context.Background(), 60*time.Second)
+			defer cancelFunc()
+
+			if err := pprofServer.Shutdown(shutdownCtx); err != nil {
+				setupLog.Error(err, "Failed to shutdown pprof HTTP server")
+			}
+		}()
+
+		if err := pprofServer.ListenAndServe(); !errors.Is(err, http.ErrServerClosed) {
+			setupLog.Error(err, "Failed to start pprof HTTP server")
```
cheyang: If the port is already in use, this logs an error but the controller keeps running — the user won't notice pprof isn't available until they try to connect. Since pprof is explicitly opt-in via `--enable-pprof=true`, it might be worth failing fast here, or at least documenting that pprof startup is intentionally best-effort.

cmd/rbgs/main.go
```diff
@@ -519,6 +532,43 @@ func setupWebhookCertController(mgr ctrl.Manager, result *webhookBootstrapResult
 	return webhookCertReconciler.SetupWithManager(mgr, options)
 }
 
+func startPprofServer(pprofAddr string) {
+	setupLog.Info("Enabling pprof", "addr", pprofAddr)
```
cheyang: Nit: "Enabling pprof" here and "Starting pprof HTTP server" on L551 convey the same thing — keeping just one would reduce noise.

cheyang (approved):

PR #349, opened 2026-05-22
fix: prevent shouldRecreateInstance from triggering during updates

### LINKED ISSUE #348: [RFC] fix shouldRecreateInstance incorrectly triggers during in-place updates
## Summary

When `RestartPolicy=RecreateRoleInstanceOnPodRestart` (the default) is used together with `InPlaceIfPossible` update strategy, in-place image updates are defeated by the restart policy. The in-place update changes the container image, kubelet restarts the container, and `shouldRecreateInstance` sees `RestartCount > 0` — interpreting the expected restart as a crash — and triggers full instance recreation.

This problem also affects rolling updates under certain timing conditions.

## Root Cause

The current guard in `shouldRecreateInstance`:

```go
if instance.Generation != instance.Status.ObservedGeneration {
    return false
}
```

Only protects during the **first reconcile** after a spec change. By the second reconcile (triggered by the container restart event from the in-place update), `ObservedGeneration` has already caught up to `Generation`, and the guard no longer applies.

### Timeline of the bug

```
1. User updates image → Generation incremented
2. Reconcile #1: Generation != ObservedGeneration → shouldRecreateInstance returns false ✓
3. Controller applies in-place update, sets ObservedGeneration = Generation
4. Kubelet detects image change → restarts container → RestartCount goes from 0 to 1
5. Pod event triggers Reconcile #2: Generation == ObservedGeneration now
6. shouldRecreateInstance sees RestartCount > 0 → returns true ✗
7. All pods deleted and recreated → in-place update defeated
```

## Proposed Fix

Add three layers of protection in `shouldRecreateInstance`:

### Layer 1: Rolling update guard

```go
if instance.Status.CurrentRevision != instance.Status.UpdateRevision {
    return false
}
```

`CurrentRevision != UpdateRevision` means pods haven't all converged to the target revision. Container restarts during this window are from the update process, not unexpected failures.

### Layer 2: In-place update in-progress guard

```go
func isPodInPlaceUpdating(pod *v1.Pod) bool {
    cond := GetInPlaceCondition(pod)
    return cond != nil && cond.Status == v1.ConditionFalse
}
```

When an in-place update starts, the `InPlaceUpdateReady` pod condition is set to `False`. If any non-ignored pod has this condition, the container restart is expected.

### Layer 3: Post-completion RestartCount baseline comparison (status-level map)

> **Update (2026-05-25):** The original Layer 3 design used Pod annotations (`InPlaceUpdateState`) to record `RestartCount` per container. This has been replaced with a status-level map approach. See [Rationale for replacing annotation-based Layer 3](#rationale-for-replacing-annotation-based-layer-3) below.

Record each container's `RestartCount` before applying the in-place update into `RoleInstanceStatus.InPlaceUpdateContainerRestartCounts`. This is a `map[string]map[string]int32` where the outer key is the pod name, inner key is container name, and value is the pre-update RestartCount baseline.

After the update completes, compare:

- `current RestartCount <= baseline + 1` → restart is from the in-place update → skip
- `current RestartCount > baseline + 1` → additional crash after update → trigger recreation
- Container not in baseline map but restarted → not an in-place updated container → real crash → trigger recreation

```go
func isContainerRestartExpected(pod *v1.Pod, baselines map[string]map[string]int32) bool {
    if baselines == nil {
        return false
    }
    containerBaselines, ok := baselines[pod.Name]
    if !ok {
        return false
    }
    for _, cs := range pod.Status.ContainerStatuses {
        if cs.RestartCount == 0 {
            continue
        }
        baseline, wasUpdated := containerBaselines[cs.Name]
        if !wasUpdated {
            // Container not in-place updated but restarted → real crash
            return false
        }
        if cs.RestartCount > baseline+1 {
            // More restarts than expected → real crash
            return false
        }
    }
    return true
}
```

#### Lifecycle management

The baselines map is managed entirely by the controller:

1. **Set**: In the Update flow (`updatePod`), when an in-place update succeeds, `recordInPlaceUpdateBaselines(pod, newStatus)` writes the current `RestartCount` for all containers of that pod into `newStatus.InPlaceUpdateContainerRestartCounts[podName]`.
2. **Inherit**: On each reconcile, `newStatus.InPlaceUpdateContainerRestartCounts` is initialized from `instance.Status.InPlaceUpdateContainerRestartCounts`.
3. **Clear**: In `calculateStatus`, when all pods have converged to the `UpdateRevision` (i.e., `CurrentRevision = UpdateRevision`), the baselines are set to `nil`. Any subsequent container restart is correctly treated as a real crash.

### Modified `shouldRecreateInstance` logic

```go
func shouldRecreateInstance(instance, pods, baselines) bool {
    if restartPolicy != RecreateRoleInstanceOnPodRestart { return false }
    if len(pods) == 0 { return false }
    if !wasInstanceReady || Generation != ObservedGeneration { return false }
    if CurrentRevision != UpdateRevision { return false }          // Layer 1

    for _, p := range pods {
        if hasTriggerPolicyIgnore(p) { continue }
        if isPodInPlaceUpdating(p) { return false }                // Layer 2
        if p.Phase == Failed && !deleting { return true }
        if containerRestarted(p) &&
           !isContainerRestartExpected(p, baselines) {             // Layer 3
            return true
        }
    }
    return false
}
```

## Rationale for replacing annotation-based Layer 3

The initial implementation (PR #349 v1) used the `InPlaceUpdateState` Pod annotation to record `RestartCount` per container. This approach had three problems:

1. **No lifecycle management**: The annotation is never cleaned up. After pods converge to the new revision, the stale annotation can mask real crashes if the `RestartCount` happens to fall within `baseline+1`.

2. **Backward compatibility issues**: The `RestartCount` field was added with `omitempty` to `InPlaceUpdateContainerStatus`. In upgrade scenarios, existing pods have no `RestartCount` in their annotation — the zero value is indistinguishable from "not present", causing false positives where all restarts are treated as expected.

3. **Multi-container ambiguity**: A unified boolean condition (e.g., `InPlaceUpdateInProgress=True`) cannot distinguish which container's restart is expected. Consider a pod with containers A (in-place updated) and B (not updated):
   - Container A restarts: expected (from image change) → should NOT trigger recreation
   - Container B restarts: unexpected (real crash) → SHOULD trigger recreation
   
   Only a per-container map can correctly distinguish these cases.

The status-level map approach solves all three:
- **Lifecycle**: Controller explicitly clears baselines when pods converge
- **Backward compat**: New field — old instances have `nil` map → `isContainerRestartExpected` returns `false` → behaves as before (recreation triggered)
- **Multi-container**: Per-container baselines allow precise identification of expected vs unexpected restarts

## No Loop Problem

A concern is whether the fix could cause a loop: update → restart → recreation → restart → recreation...

This does **not** happen because:

- **During in-place update**: Layer 2 (`InPlaceUpdateReady=False`) blocks recreation
- **After in-place update completes**: Layer 3 (RestartCount baseline in status) blocks recreation for the expected single restart
- **During rolling update**: Layer 1 (`CurrentRevision != UpdateRevision`) blocks recreation
- **After recreation**: New pods have `RestartCount=0` and no baselines in status, so `containerRestarted()` returns false — no loop
- **After pods converge**: Baselines are cleared from status. A real crash (RestartCount > 0 with no baselines) correctly triggers recreation

## Files Changed

| File | Change |
|------|--------|
| `api/workloads/v1alpha2/roleinstance_types.go` | Add `InPlaceUpdateContainerRestartCounts map[string]map[string]int32` to `RoleInstanceStatus` |
| `api/workloads/v1alpha2/zz_generated.deepcopy.go` | Auto-generated deep copy for new nested map field |
| `pkg/reconciler/roleinstance/sync/api.go` | `Update()` interface adds `newStatus` parameter |
| `pkg/reconciler/roleinstance/sync/instance_update.go` | Record baselines via `recordInPlaceUpdateBaselines()` after successful in-place update |
| `pkg/reconciler/roleinstance/sync/instance_scale.go` | Replace annotation-based `isContainerRestartExpectedFromInPlaceUpdate` with status-based `isContainerRestartExpected`; pass baselines to `shouldRecreateInstance` |
| `pkg/reconciler/roleinstance/instance_reconciler.go` | Inherit baselines in `newStatus` initialization; pass `newStatus` to `Update()` |
| `pkg/reconciler/roleinstance/instance_status.go` | Clear baselines when pods converge in `calculateStatus`; add `inconsistentBaselines` for status diff |
| `api/workloads/pub/inplace_update/inplace_update.go` | Revert: remove `RestartCount` field from `InPlaceUpdateContainerStatus` |
| `pkg/inplace/pod/inplaceupdate/inplace_update.go` | Revert: remove `RestartCount` recording from `updatePodInPlace` |

## Testing

All existing unit tests pass. New and updated test coverage:

### `instance_scale_test.go`
- `CurrentRevision != UpdateRevision` → should NOT recreate (Layer 1)
- `InPlaceUpdateReady=False` → should NOT recreate (Layer 2)
- Status baselines: `RestartCount == baseline+1` → should NOT recreate (expected restart)
- Status baselines: `RestartCount > baseline+1` → should recreate (real crash after update)
- Status baselines: container not in baseline map but restarted → should recreate
- **Multi-container**: Container A in-place updated (expected restart) + Container B crash → should recreate
- **Multi-container**: All containers in-place updated, all expected restarts → should NOT recreate
- **Nil baselines** (backward compat): container restart with no baselines → should recreate
- **Pod name mismatch**: baselines exist for different pod → should recreate
- `isContainerRestartExpected` unit tests: 10 sub-cases covering nil baselines, pod not found, zero restart, within baseline, over baseline, multi-container mixed scenarios, etc.
- Restarting cache prevents recreation (passes nil baselines)

### `instance_update_test.go`
- `recordInPlaceUpdateBaselines`: records baselines for all containers
- `recordInPlaceUpdateBaselines`: overwrites existing baselines for same pod
- `recordInPlaceUpdateBaselines`: preserves baselines for other pods
- `updatePod` integration: verifies baselines are recorded after successful in-place update

### `instance_status_test.go`
- `inconsistentBaselines`: 9 cases covering nil, empty, same content, different pod/container names, different counts, extra pods, both empty maps

Full `make test` and `make lint` pass.

Syspretor:
## Summary

Fixes #348

When `RestartPolicy=RecreateRoleInstanceOnPodRestart` (the default) is used with `InPlaceIfPossible` update strategy, in-place image updates are defeated by the restart policy. The in-place update changes the container image, kubelet restarts the container, and `shouldRecreateInstance` interprets the expected restart as a crash — triggering full instance recreation.

The previous guard (`Generation != ObservedGeneration`) only protects during the first reconcile after a spec change. By the second reconcile (triggered by the container restart event), `ObservedGeneration` has caught up and the guard no longer applies.

This commit adds three layers of protection:

- **Layer 1**: `CurrentRevision != UpdateRevision` — skip recreation during rolling updates when pods haven't converged to the target revision
- **Layer 2**: `isPodInPlaceUpdating` — skip recreation when any pod has `InPlaceUpdateReady=False` (in-place update in progress)
- **Layer 3**: `isContainerRestartExpectedFromInPlaceUpdate` — record pre-update `RestartCount` baseline in `InPlaceUpdateContainerStatus`; a single restart (`current <= baseline+1`) is expected from the in-place update, additional restarts indicate a real crash

## Test plan

- [x] 5 new unit test cases added for all guard conditions
- [x] All existing unit tests pass (`make test`)
- [ ] E2e: InPlaceIfPossible update with default RestartPolicy should not trigger recreation


pkg/inplace/pod/inplaceupdate/inplace_update.go
```diff
@@ -293,7 +293,8 @@ func (c *realControl) updatePodInPlace(pod *v1.Pod, spec *UpdateSpec, opts *Upda
 		for _, c := range clone.Status.ContainerStatuses {
 			if _, ok := spec.ContainerImages[c.Name]; ok {
 				inPlaceUpdateState.LastContainerStatuses[c.Name] = inplaceapi.InPlaceUpdateContainerStatus{
-					ImageID: c.ImageID,
+					ImageID:      c.ImageID,
+					RestartCount: c.RestartCount,
```
cheyang: Recording the baseline only for containers in `spec.ContainerImages` breaks sequential in-place updates that touch different containers. After an update of container A, the next in-place update of container B rebuilds `LastContainerStatuses` with only B's entry, so A's existing `RestartCount > 0` is no longer in the baseline. On the next reconcile, `isContainerRestartExpectedFromInPlaceUpdate` hits the `wasUpdated == false` branch for A and returns false, which causes `shouldRecreateInstance` to trigger a full recreation even though nothing crashed. I think the snapshot should record `RestartCount` for every container in `clone.Status.ContainerStatuses`, not just the ones being patched this round.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -387,6 +400,48 @@ func containerRestarted(pod *v1.Pod) bool {
...
+	for _, cs := range pod.Status.ContainerStatuses {
+		if cs.RestartCount == 0 {
+			continue
+		}
+		lastStatus, wasUpdated := state.LastContainerStatuses[cs.Name]
+		if !wasUpdated {
+			// Container was not in-place updated but has restarted → real crash
+			return false
+		}
+		if cs.RestartCount > lastStatus.RestartCount+1 {
```
cheyang: The `+1` allowance here is never retired, so it leaks into normal operation long after the in-place update has finished. Once an in-place update completes, `LastContainerStatuses` keeps the pre-update `RestartCount=N` baseline, and `cs.RestartCount > N+1` only becomes true at the *second* real crash. That means a single genuine OOM/crash days later still satisfies `RestartCount == baseline+1`, so this returns true ("expected") and `shouldRecreateInstance` skips recreation — defeating `RecreateRoleInstanceOnPodRestart` outside the update window. The baseline needs to be advanced or the annotation cleared once the in-place update is observed complete (e.g., when `InPlaceUpdateReady` flips back to True and the container has converged), so the one-restart credit is consumed exactly once.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -387,6 +400,48 @@ func containerRestarted(pod *v1.Pod) bool {
...
+	}
+	var state inplaceapi.InPlaceUpdateState
+	if err := json.Unmarshal([]byte(stateStr), &state); err != nil {
+		return false
+	}
+	for _, cs := range pod.Status.ContainerStatuses {
+		if cs.RestartCount == 0 {
+			continue
+		}
+		lastStatus, wasUpdated := state.LastContainerStatuses[cs.Name]
```
cheyang: Worth comparing `cs.ImageID` against `lastStatus.ImageID` here as well. RestartCount alone can't tell whether the restart we are excusing is actually the in-place update's restart, especially once the snapshot also covers untouched containers (per the other comment). Allowing `+1` only when the image actually changed (`cs.ImageID != lastStatus.ImageID`) makes the heuristic much harder to misread for sidecars or for containers that were not part of this revision's patch.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -354,22 +356,33 @@ func shouldRecreateInstance(instance *workloadsv1alpha2.RoleInstance, pods []*v1
...
+	}
 
 	for _, p := range pods {
 		if hasTriggerPolicyIgnore(p) {
 			continue
 		}
+		// If any pod is currently undergoing an in-place update, skip recreation entirely.
+		// The container restart is expected and should not trigger instance recreation.
+		if isPodInPlaceUpdating(p) {
+			return false
```
cheyang: Returning `false` for the whole instance as soon as any single pod has `InPlaceUpdateReady=False` is a pretty wide hammer for multi-pod RoleInstances. While one pod is mid-update, a real `PodFailed` or repeated crash on a sibling pod will be silently ignored until the updating pod converges. Suggest narrowing the bypass to the updating pod itself (e.g., `continue` instead of returning), so genuine failures on other pods still drive recreation.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -387,6 +400,48 @@ func containerRestarted(pod *v1.Pod) bool {
...
+func isContainerRestartExpectedFromInPlaceUpdate(pod *v1.Pod) bool {
+	if pod.Annotations == nil {
+		return false
+	}
+	stateStr, ok := pod.Annotations[constants.InPlaceUpdateStateKey]
+	if !ok {
+		return false
+	}
+	var state inplaceapi.InPlaceUpdateState
+	if err := json.Unmarshal([]byte(stateStr), &state); err != nil {
```
cheyang: Silent `return false` on `json.Unmarshal` failure means a corrupted `InPlaceUpdateState` annotation collapses back to the old behavior (any restart → recreate) without any signal. A `klog.Warning` here would make this much easier to debug if it ever fires in production.

pkg/reconciler/roleinstance/sync/instance_scale_test.go
```diff
@@ -539,6 +541,262 @@ func TestShouldRecreateInstance(t *testing.T) {
...
+							},
+						},
+					},
+				}
+			}(),
+			expected: true,
+		},
 	}
 
 	for _, tt := range tests {
```
cheyang, line 1127: A few cases worth adding: (1) two sequential in-place updates targeting different containers, asserting no recreation after the second one (covers the sequential-update gap); (2) `InPlaceUpdateReady=True` plus a real crash long after update completion, asserting recreation does happen (covers the persistent `+1` allowance); (3) one pod mid-update and a sibling pod in `PodFailed`, asserting the sibling failure is not masked; (4) a malformed `InPlaceUpdateState` annotation, asserting expected fallback behavior.

cheyang (commented):
Direction looks right — pulling update-window awareness into `shouldRecreateInstance` is the right level. Two things stood out as needing changes before merge: the baseline only covers containers being patched this round, which breaks sequential in-place updates of different containers; and the `+1` restart allowance is never retired, so a real crash long after the update window still gets excused. I also agree with the `gemini-code-assist` comments on those same two spots and have left additional suggestions inline.

cheyang:
Direction is right and the second commit's switch from annotation to a status field is a clear improvement over the earlier annotation-based approach. Three of the layered guards still have gaps that, taken together, fire in normal operation:

- Baselines are wiped on convergence (`instance_status.go:144`), so the very next reconcile after a successful in-place update mistakes the kubelet-driven restart for a crash and recreates the instance.
- The `+1` allowance in `isContainerRestartExpected` (`instance_scale.go:431`) is granted to every container tracked in a pod's baseline, regardless of whether that container's image actually changed — so a real one-off crash on a non-updated container is silently absorbed.
- The `isPodInPlaceUpdating` bypass returns from the whole loop (`instance_scale.go:376`) instead of skipping just that pod, which masks `PodFailed` on a sibling pod for as long as the updating pod hasn't converged.

Details and concrete fix sketches inline. Once these are addressed, the design lands in a good place.

pkg/reconciler/roleinstance/instance_status.go
```diff
@@ -139,6 +139,9 @@ func (r *realStatusUpdater) calculateStatus(instance *workloadsv1alpha2.RoleInst
 	}
 	if newStatusUpdatedReplicas == newStatusReplicas && newStatusReplicas == componentSize {
 		newStatus.CurrentRevision = newStatus.UpdateRevision
+		// All pods have converged to the update revision — clear in-place update
+		// baselines so that any future container restart is treated as a real crash.
+		newStatus.InPlaceUpdateContainerRestartCounts = nil
```
cheyang: Clearing `InPlaceUpdateContainerRestartCounts` here at the same moment `CurrentRevision` is set to `UpdateRevision` opens a window that defeats Layer 3 right when it's needed most.

`pod.Status.ContainerStatuses[*].RestartCount` is a lifetime counter — kubelet does not reset it after an in-place image swap. So once we converge:

- `CurrentRevision == UpdateRevision` → Layer 1 doesn't apply
- `InPlaceUpdateReady=True` → Layer 2 doesn't apply
- `baselines == nil` → `isContainerRestartExpected` returns false → Layer 3 doesn't apply
- `containerRestarted(p)` is still true because RestartCount stays at 1
- `shouldRecreateInstance` returns true → instance gets recreated

This fires on the next reconcile after convergence (probe update, resync, any unrelated status write — easy to hit in practice). It's not a rare race; it's the expected outcome of a successful in-place update once a single follow-up reconcile happens.

The existing `"No baselines (nil map) with container restart - should recreate"` test pins this behavior in. The intent there ("nil baselines on first upgrade") is fine, but combined with the convergence-time clear it becomes a bug: every pod that just finished an in-place update lands in this branch.

Recommendation: don't wipe baselines on convergence. If `isContainerRestartExpected` learns to compare `ImageID` (see the comment on `instance_scale.go`), baselines self-classify and don't need to expire. What's actually needed here is **orphan pruning** — drop entries for pods that no longer exist in the active pod list, instead of clearing the whole map:

```go
activePods := make(map[string]struct{}, len(pods))
for _, p := range pods {
    activePods[p.Name] = struct{}{}
}
for podName := range newStatus.InPlaceUpdateContainerRestartCounts {
    if _, alive := activePods[podName]; !alive {
        delete(newStatus.InPlaceUpdateContainerRestartCounts, podName)
    }
}
```

A test that converges then triggers a second reconcile (no real crash) and asserts no recreation would catch this.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -387,6 +398,44 @@ func containerRestarted(pod *v1.Pod) bool {
...
+	for _, cs := range pod.Status.ContainerStatuses {
+		if cs.RestartCount == 0 {
+			continue
+		}
+		baseline, wasUpdated := containerBaselines[cs.Name]
+		if !wasUpdated {
+			// Container was not in-place updated but has restarted → real crash
+			return false
+		}
+		if cs.RestartCount > baseline+1 {
```
cheyang: `recordInPlaceUpdateBaselines` correctly snapshots every container in the pod (this fixes the sequential-update gap raised in the previous round), but the comparison here only looks at `RestartCount`. That gives every tracked container — including ones whose image was not changed this round — a free `+1` against its baseline.

Concrete failure mode (single pod, two containers `inference` + `sidecar`, only `inference`'s image was patched):

1. baseline = `{inference: 0, sidecar: 0}`
2. kubelet restarts `inference` → `RestartCount=1` (expected)
3. While baselines are still around (e.g., other pods haven't converged), `sidecar` legitimately OOMs once → `RestartCount=1`
4. Check: `1 > 0+1` is false → `isContainerRestartExpected` returns true → recreation skipped

A real crash on a non-updated container is silently absorbed. This widens further across multi-pod rolling updates where the baseline map lingers for the whole window.

gemini-code-assist and the previous review thread both pointed at `ImageID` as the right discriminator. The earlier `InPlaceUpdateContainerStatus { ImageID, RestartCount }` shape had it; this revision dropped it. Recommendation:

1. Change the value type to a struct that carries `ImageID` alongside `RestartCount` (rename the field to `InPlaceUpdateContainerBaselines` while at it — `v1alpha2` is alpha and unreleased, the cost is low).
2. Snapshot `cs.ImageID` together with `cs.RestartCount` in `recordInPlaceUpdateBaselines`.
3. Only grant `+1` here when `cs.ImageID != baseline.ImageID`:

```go
allowed := baseline.RestartCount
if cs.ImageID != baseline.ImageID {
    allowed++ // image actually changed; one kubelet-driven restart is expected
}
if cs.RestartCount > allowed {
    return false
}
```

Note: do **not** revert to "only record updated containers" — that brings back the sequential-update bug fixed earlier. Recording every container is correct; the discriminator belongs in the value, not in which keys are present.

With ImageID-aware logic, B1's clearing also becomes unnecessary, since baselines self-distinguish stale-but-harmless entries from real crashes.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -354,22 +354,33 @@ func shouldRecreateInstance(instance *workloadsv1alpha2.RoleInstance, pods []*v1
...
+	}
 
 	for _, p := range pods {
 		if hasTriggerPolicyIgnore(p) {
 			continue
 		}
+		// If any pod is currently undergoing an in-place update, skip recreation entirely.
+		// The container restart is expected and should not trigger instance recreation.
+		if isPodInPlaceUpdating(p) {
+			return false
```
cheyang: `isPodInPlaceUpdating(p)` is a per-pod predicate, but `return false` here aborts the whole loop. In a multi-pod RoleInstance, if `pod-0` is mid in-place update and `pod-1` is in `PodFailed` (a real failure unrelated to the update), the function returns false on `pod-0` and the genuine failure on `pod-1` is masked until `pod-0`'s update converges — which can stall on a bad image, an unsatisfied readiness gate, etc.

Same point was raised on the earlier revision; the rewrite kept the same shape. Suggest narrowing the bypass to the updating pod itself:

```go
if isPodInPlaceUpdating(p) {
    continue // skip THIS pod's restart check; siblings still drive recreation
}
```

A matching test (one pod with `InPlaceUpdateReady=False`, sibling pod in `PodFailed`, `expected=true`) would pin the new behavior and prevent regression.

pkg/reconciler/roleinstance/sync/instance_scale_test.go
```diff
@@ -1038,3 +1424,156 @@ func TestContainerRestarted(t *testing.T) {
 		})
 	}
 }
+
+func TestIsContainerRestartExpected(t *testing.T) {
```
cheyang, line 1625: The unit-level coverage of `isContainerRestartExpected` is solid, but a few cases that map directly to the issues raised would make regressions visible:

1. **Post-convergence reconcile (B1)**: drive `calculateStatus` to convergence so baselines get cleared, then call `shouldRecreateInstance` again on a pod whose `RestartCount` is still `1`. With current code this asserts `true` (recreate); after the fix it should assert `false`.
2. **Image unchanged + single restart (B2)**: single pod, baseline tracks both containers, only `inference`'s `ImageID` changes; `sidecar.RestartCount=1`. Today this asserts `expected=false` (recreate skipped — the bug); after the ImageID gate it should assert `expected=true` (recreate).
3. **Sibling failure during in-place update (B3)**: pod-0 with `InPlaceUpdateReady=False`, pod-1 in `PodFailed` — should assert `true` once the bypass becomes per-pod.
4. **Sequential in-place updates** (the motivating scenario for the second commit's refactor): revision N updates container A, revision N+1 updates container B; assert no recreation after either restart. The component-level helpers are tested but not the end-to-end path.

Also: the existing `"No baselines (nil map) with container restart - should recreate"` test will need to be revisited once B1 is fixed, since post-convergence baselines won't be nil anymore.

cheyang:
Good progress — the three blocking issues from the previous round are fixed correctly. One remaining concern: `recordInPlaceUpdateBaselines` overwrites the entire pod entry on each call instead of merging, which drops baselines from a prior in-place update of a different container set. Details inline.

pkg/reconciler/roleinstance/sync/instance_update.go
```diff
@@ -103,3 +110,27 @@ func (c *realControl) updatePod(ctx context.Context, instance *workloadsv1alpha2
...
+	for _, name := range updatedContainers {
+		updatedSet[name] = true
+	}
+	containerBaselines := make(map[string]int32, len(updatedContainers))
+	for _, cs := range pod.Status.ContainerStatuses {
+		if updatedSet[cs.Name] {
+			containerBaselines[cs.Name] = cs.RestartCount
+		}
+	}
+	newStatus.InPlaceUpdateContainerRestartCounts[pod.Name] = containerBaselines
```
cheyang: This line replaces the pod's entire baseline map on each in-place update:

```go
newStatus.InPlaceUpdateContainerRestartCounts[pod.Name] = containerBaselines
```

With the new selective-recording strategy (only updated containers go into `containerBaselines`), this overwrites any prior entries for the same pod.

Sequential update scenario:
1. Revision N updates container A → baselines become `{"pod-0": {"A": 0}}`
2. Kubelet restarts A → `RestartCount=1` (expected)
3. Revision N+1 updates container B → baselines become `{"pod-0": {"B": 0}}` — A's entry is gone
4. Next reconcile: `isContainerRestartExpected` checks A, finds it not in baselines → `!wasUpdated → return false` → triggers recreation even though A's restart is still legitimate

Layer 1 (`CurrentRevision != UpdateRevision`) covers most practical timing, but if the first update fully converges before the second spec change arrives, this path is reachable.

Fix: merge into existing baselines instead of overwriting:

```go
existing := newStatus.InPlaceUpdateContainerRestartCounts[pod.Name]
if existing == nil {
    existing = make(map[string]int32, len(updatedContainers))
}
for _, cs := range pod.Status.ContainerStatuses {
    if updatedSet[cs.Name] {
        existing[cs.Name] = cs.RestartCount
    }
}
newStatus.InPlaceUpdateContainerRestartCounts[pod.Name] = existing
```

This preserves A's baseline while adding B's. The GC in `calculateStatus` still handles stale entries for deleted pods.

pkg/reconciler/roleinstance/sync/instance_update.go
```diff
@@ -86,11 +87,17 @@ func (c *realControl) updatePod(ctx context.Context, instance *workloadsv1alpha2
...
+	opts := coreControl.GetUpdateOptions()
+	res := c.inplaceControl.Update(ctx, pod, oldRevision, updateRevision, opts)
 	if res.InPlaceUpdate {
 		if res.UpdateErr == nil {
 			c.recorder.Eventf(instance, v1.EventTypeNormal, "SuccessfulUpdatePodInPlace", "successfully update pod %s in-place", pod.Name)
 			instanceutil.ResourceVersionExpectations.Expect(&metav1.ObjectMeta{UID: pod.UID, ResourceVersion: res.NewResourceVersion})
+			// Record pre-update RestartCount baselines only for containers that were
+			// actually in-place updated (image changed). Non-updated containers should
+			// not be protected — their restarts indicate real crashes.
+			updatedContainers := c.inplaceControl.GetUpdatedContainerNames(ctx, pod, oldRevision, updateRevision, opts)
```
cheyang, line 99: `GetUpdatedContainerNames` silently returns nil on any internal error (revision parsing, component mismatch, nil spec). Since `recordInPlaceUpdateBaselines` exits early on `len(updatedContainers) == 0`, a successful in-place update with a nil return here means no baseline gets recorded — and the next reconcile will misinterpret the expected restart as a crash.

At minimum, a `klog.Warning` on the nil path would surface the problem in production. A safer fallback would be to record all containers (the pre-refactor behavior) when the updated-container list can't be determined, so the failure mode is "too lenient" rather than "false recreation".

pkg/reconciler/roleinstance/sync/instance_update_test.go
```diff
@@ -180,3 +194,198 @@ func TestUpdatePod(t *testing.T) {
...
+		}
+		newStatus := &workloadsv1alpha2.RoleInstanceStatus{}
+		recordInPlaceUpdateBaselines(pod, newStatus, nil)
+
+		if newStatus.InPlaceUpdateContainerRestartCounts != nil {
+			t.Error("expected no baselines when updatedContainers is nil")
+		}
+	})
+
+	t.Run("overwrites existing baselines for same pod", func(t *testing.T) {
```
cheyang: The `"overwrites existing baselines for same pod"` test currently verifies that calling `recordInPlaceUpdateBaselines` a second time replaces the prior entry. Once the overwrite-to-merge fix lands, this test should be updated to verify that a second call for a different container **preserves** the first container's baseline while adding the new one. Something like:

```go
t.Run("merges baselines across sequential updates", func(t *testing.T) {
    pod := &v1.Pod{
        ObjectMeta: metav1.ObjectMeta{Name: "pod-0"},
        Status: v1.PodStatus{
            ContainerStatuses: []v1.ContainerStatus{
                {Name: "A", RestartCount: 1},
                {Name: "B", RestartCount: 0},
            },
        },
    }
    newStatus := &workloadsv1alpha2.RoleInstanceStatus{
        InPlaceUpdateContainerRestartCounts: map[string]map[string]int32{
            "pod-0": {"A": 0}, // from prior update
        },
    }
    recordInPlaceUpdateBaselines(pod, newStatus, []string{"B"})
    // A's baseline preserved, B's added
    assert.Equal(t, int32(0), newStatus.InPlaceUpdateContainerRestartCounts["pod-0"]["A"])
    assert.Equal(t, int32(0), newStatus.InPlaceUpdateContainerRestartCounts["pod-0"]["B"])
})
```

Also worth adding: a `calculateStatus` unit test that exercises the GC prune path (pod present in baselines but not in the active pod list → entry removed).

config/crd/bases/workloads.x-k8s.io_roleinstances.yaml
```diff
@@ -210,6 +210,32 @@ spec:
                 description: CurrentRevision is a hash value that changes when the
                   spec is changed.
                 type: string
+              inPlaceUpdateContainerBaselines:
```
NoobDream2568, line 213: Nit:
The RoleInstance CRD in deploy/kubectl/manifests.yaml is missing the newly added inPlaceUpdateContainerBaselines field — config/crd/bases/workloads.x-k8s.io_roleinstances.yaml has been updated, but the bundled manifest wasn't regenerated.
Consider running make manifests (or the equivalent generation command) to sync it, so that users deploying via kubectl apply -f deploy/kubectl/manifests.yaml get the complete status schema.
Not blocking — can be addressed in a follow-up.

cheyang (approved):
All three blocking issues from the initial review and the follow-up (sequential-update baseline overwrite) have been addressed correctly:

1. **Baseline clearing** → GC-based pruning that retains entries for active pods.
2. **ImageID-aware +1 allowance** → `ContainerUpdateBaseline` struct carrying both `RestartCount` and `ImageID`; only grants the restart credit when the image actually changed.
3. **Sibling pod masking** → `return false` → `continue`.
4. **Sequential update merge** → baselines are merged into existing entries instead of overwriting.
5. **Deep-copy to prevent aliasing** — good catch on the `inconsistentBaselines` false-negative.
6. **Stale baseline detection** — `RestartCount < baseline` correctly handles post-recreation pods.

Unit tests, envtest, and e2e all green. CI passes on all checks.

/lgtm
/approve

Issue #351, opened 2026-05-26
[Feature] Support In-place Scheduling


NoobDream2568:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

In AI inference workloads, model weight files are typically tens to hundreds of gigabytes. When a Pod is scheduled to a node for the first time, these files must be downloaded from remote storage, which can take minutes. However, if the same node is already hosting a Pod running the same model, the weight files are already cached on the node's local disk or in memory.

The primary scenario is rolling upgrades with the recreate pod update policy. During a rolling upgrade with MaxSurge, new Pods are created before old Pods are deleted. If the new Pods can land on the same nodes as the old Pods, they immediately reuse the cached model weights and GPU warm-up state, enabling faster upgrade.

Without in-place scheduling, the Kubernetes scheduler has no awareness of where the old Pods are running. New Pods may be placed on entirely different nodes, forcing a full re-download of model weights (10+ minutes for large models) and defeating the purpose of a fast rolling upgrade.

Thus, we propose introducing the In-place Scheduling mechanism.

PR #352, opened 2026-05-26
kep: support inplace scheduling

NoobDream2568:

None. This is [kep-351](https://github.com/sgl-project/rbg/issues/351)

## Checklist

### DIFF
--- a/keps/351-inplace-scheduling/inplace-scheduling-diagram-component.svg
+++ b/keps/351-inplace-scheduling/inplace-scheduling-diagram-component.svg
@@ -1,0 +1,111 @@
+<svg viewBox="0 0 920 420" xmlns="http://www.w3.org/2000/svg" style="font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, sans-serif; background: #ffffff;">
+
+<!-- Title -->
+<text x="460" y="32" text-anchor="middle" fill="#1a1a2e" font-size="20" font-weight="500">Inplace Scheduling: Component Strategy</text>
+
+<!-- ===== Step 1: Running State ===== -->
+<text x="150" y="65" text-anchor="middle" fill="#7c3aed" font-size="13" font-weight="500">1. Running State</text>
+
+<!-- Node A -->
+<rect x="20" y="80" width="260" height="130" rx="8" fill="#f8f9fa" stroke="#dee2e6" stroke-width="1.5"/>
+<text x="32" y="100" fill="#1a1a2e" font-size="12" font-weight="500">Node A</text>
+<rect x="36" y="110" width="95" height="38" rx="4" fill="#dbeafe" stroke="#3b82f6" stroke-width="1"/>
+<text x="83" y="133" text-anchor="middle" fill="#2563eb" font-size="10">abc-Prefill-0 (v1)</text>
+<rect x="36" y="158" width="228" height="24" rx="4" fill="#dcfce7" stroke="#22c55e" stroke-width="1" stroke-dasharray="3,2"/>
+<text x="150" y="174" text-anchor="middle" fill="#16a34a" font-size="9">Model Weights Cache (140GB)</text>
+
+<!-- Node B -->
+<rect x="20" y="225" width="260" height="130" rx="8" fill="#f8f9fa" stroke="#dee2e6" stroke-width="1.5"/>
+<text x="32" y="245" fill="#1a1a2e" font-size="12" font-weight="500">Node B</text>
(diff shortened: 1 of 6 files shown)


cheyang (changes requested):
## KEP-351 Review: Inplace Scheduling

The motivation is solid and the document is well-structured overall. However, there is a design gap in the primary use case that needs to be addressed before this KEP can move forward.

### Critical: Binding Lookup Does Not Work for MaxSurge Rolling Updates

The KEP states the primary scenario is rolling upgrades with `maxSurge > 0` and `maxUnavailable: 0` (create-before-delete). It also says:

> "Pod names are deterministic (derived from instance name + component + ID), so a recreated Pod has the same name and picks up the same binding."

But in a MaxSurge rolling update, new Pods are created **while old Pods are still running**. This means the new surge Pods **cannot** have the same name as the old Pods (Kubernetes does not allow two Pods with the same name in the same namespace simultaneously).

The node binding store uses `{rbgUID}/{podName}` as the key. When a new surge Pod (with a different name/index than the old Pod) is created, its lookup into the store by its own name will find **no binding** — because the stored binding is under the old Pod's name.

This contradicts the timeline diagram in User Story 1:
```
New Pods created:   [new-Pod-0 → prefers node-A]
                    [new-Pod-1 → prefers node-B]
```

How does `new-Pod-0` know to look up the binding for `Pod-0`? The design section does not describe any mapping from new surge instances to old instances.

Please clarify:
1. How does the RBG rolling update controller map new surge instances to old instances?
2. What is the actual Pod naming during MaxSurge? If new instances get new indices, the pod-name-based lookup won't find old bindings.
3. If the intention is that bindings are looked up by some index-based mapping rather than literal Pod name, the design section needs to describe this explicitly.

This is the core mechanism that makes the feature work for its primary use case — it needs to be airtight.

### Minor: Observability

The KEP doesn't mention any observability hooks. Consider adding:
- A metric or event when a binding is found and affinity is injected (helps operators confirm the feature is working)
- A metric for "binding miss" (lookup found no binding — expected on first creation, but useful for debugging if it happens during rolling updates)

### Nit: kep.yaml missing trailing newline

The `kep.yaml` file is missing a final newline character.

---

Overall the writing quality is good and the design is mostly sound. The critical gap above needs to be resolved — once the binding lookup mechanism for MaxSurge is clarified, this KEP should be in good shape.

cheyang (commented):
## Re-review: KEP-351 Inplace Scheduling (sha 94e0922)

### Previous Critical Feedback: Resolved

The MaxSurge binding lookup gap from last round is now properly addressed. The two-granularity design is the right approach:

- **Component granularity** for Stateless + MaxSurge: surge Pods look up `{rbgUID}/{roleName}-{componentName}` and get the full node set `{A, B}`. Pod names are irrelevant — the lookup is by component type. This cleanly solves the name mismatch problem I flagged.
- **Pod granularity** for Stateful + no-surge: Pod names are stable, old Pods are deleted first (`maxSurge: 0`), new Pods with identical names are created and find their per-Pod binding. No overlap.
- **Smart defaults** (Stateful→Pod, Stateless→Component) prevent misconfiguration.

The design is now coherent for the primary scenarios. The user stories, timelines, and sequence diagrams clearly illustrate both paths.

### Issue: PR Bundles Unrelated Changes

This commit includes multiple changes that have no connection to the inplace scheduling KEP:

1. `pkg/utils/client/client.go` — `NewClientWithUserAgent` (user-agent injection)
2. `internal/controller/workloads/rolebasedgroup_controller.go` — `applyRBGSAReplicasOverride` (scaling adapter drift fix)
3. `calculateNextRollingTarget` fix (rolling progress when `nextRollingStep == 0`)
4. E2E test for coordinated rolling update with single-replica roles
5. Dynamo examples with Model Express documentation

These are independently useful fixes with good test coverage, but they should be in separate PRs. A KEP PR should contain only the KEP documents. Please split the unrelated code into one or more follow-up PRs.

### Minor: Observability (Carried Over)

Still no mention of observability hooks. When implementing, consider emitting a metric or event on affinity injection (binding hit) vs. no-op (first creation / binding miss). Helps operators confirm the feature is working during rolling upgrades. Not a blocker for the KEP itself, but worth adding a brief section noting what metrics the implementation will expose.

### Nit: Fixed

The `kep.yaml` trailing newline issue from last round is fixed.

---

**Summary**: KEP design is solid now. Please split the unrelated code changes into separate PRs so this can be merged cleanly as a KEP-only change.

cheyang (approved):
/lgtm
/approve

PR #353, opened 2026-05-27
test: add controller stress test framework with KWOK simulation

Syspretor:
## Summary

- Add a comprehensive stress testing framework for the RBG controller using in-cluster KWOK to simulate workload pods at scale
- Go stress client (`test/stress/`) with rate-limited create/update/delete phases, E2E latency measurement (P50/P90/P99), and integrated pprof collection
- HTML report generation with latency distribution charts, controller configuration display, classified log analysis, and profiling data
- Helper scripts for KWOK setup, controller deployment with tuning parameters, profiling collection, and teardown

### Architecture

```
Real Nodes                    Fake Nodes (kwok)
┌─────────────────────┐      ┌─────────────────────┐
│ RBG Controller Pod  │      │ kwok-node-0..N       │
│ (real, Helm-managed)│      │ (simulated by kwok)  │
│                     │      │                      │
│ kwok Controller Pod │      │ Workload Pods        │
│ (simulates pods)    │      │ (auto Ready by kwok) │
└─────────────────────┘      └─────────────────────┘
```

### Test Phases

| Phase | What Happens | Metrics |
|-------|-------------|---------|
| Create | Submit N RBGs at target QPS, wait for Ready | E2E latency (submit → Ready) |
| Update | Update all RBGs, wait for completion | E2E latency (update → Ready) |
| Delete | Delete all RBGs, wait for removal | E2E latency (delete → 404) |

### Validated Results

Tested with 100 and 1000 RBG instances (3 roles each):

| Scale | Create P50 | Create P99 | Update P50 | Delete P50 | Success Rate |
|-------|-----------|-----------|-----------|-----------|-------------|
| 100 RBGs @ 5 QPS | 696ms | 713ms | 133ms | 131ms | 100% |
| 1000 RBGs @ 10 QPS | 40.6s | 60.0s | 135ms | 137ms | 100% |

## Test plan

- [x] `go build ./test/stress/` compiles successfully
- [x] `go vet ./test/stress/` passes
- [x] 100 RBG stress test with pprof collection produces complete HTML report
- [x] 1000 RBG stress test validates scale behavior
- [ ] Reviewer runs `go run ./test/stress/ --total-rbgs=10 --roles-per-rbg=2` against a cluster with KWOK


cheyang (changes requested):
## Review: Stress Test Framework with KWOK Simulation

Solid design overall — the KWOK-based simulation approach is the right call for testing controller scaling without burning real compute. The phased approach (create/update/delete with rate limiting) and integrated pprof collection are well thought out. A few things need fixing before this can land.

### Blocking: CI Lint Failures

The `lint` job fails with 10 errors:
- 5x `errcheck` — unchecked `Close()` returns in `pprof.go`, `report.go`, `timing.go`
- 4x `prealloc` — slices that should be pre-allocated in `create.go`, `delete.go`, `update.go`, `timing.go`
- 1x `staticcheck SA9003` — empty branch in `report.go:331`

These are straightforward fixes. The empty branch (`p.ActualQPS < p.ActualQPS*0.5`) is also logically broken — it compares a value against half of itself, which is always false for positive numbers. Looks like the intended check was comparing against the *target* QPS.

### Script Duplication

Every shell script is duplicated verbatim between `.claude/skills/stress-test/scripts/` and `test/stress/scripts/` (same git blob SHA). Pick one canonical location and either symlink or source from there. I'd suggest keeping them in `test/stress/scripts/` as the source of truth and having the skill reference that path.

### `.claude/skills/` Committed to Repo

The `.gitignore` change allows `.claude/skills/` to be committed. This is a design choice that probably warrants a broader maintainer decision — committing AI skill definitions into the repo means all contributors pull them, and the SKILL.md format is specific to Claude Code. If the intent is to make the stress test runnable by any contributor (not just via the AI skill), the test/stress/ directory already handles that.

### Concurrency Model

The create/update/delete phases launch one goroutine per RBG and rate-limit only the *start*. At `--total-rbgs=1000`, this means up to 1000 goroutines concurrently polling `WaitForRBGReady` every 500ms — that's 2000 API requests/sec just from the stress client's polling loop. Consider:
1. A worker pool / semaphore to cap concurrent waiters
2. Using the watch interface (`WatchRBGs` exists but is unused) instead of polling

### Minor

- `WatchRBGs` method is defined but never called — dead code or planned for future?
- `collectControllerLogs` shells out to `kubectl` but doesn't capture stderr; a failed `kubectl logs` silently produces an empty file
- No unit tests for the framework itself (e.g. `percentile()`, `computeStats()`, template generation). Given this is test infrastructure, some self-tests would build confidence.

### Verdict: needs-work

Fix lint, deduplicate scripts, address the dead-code branch. The concurrency concern is lower priority but worth at least documenting the expected API load.

cheyang:
The stress test framework is well-structured overall: the three-phase design (create/update/delete), rate limiting via `x/time/rate`, and integrated pprof collection are all solid choices that fit the project's testing patterns.

That said, CI is currently failing with lint errors that need to be cleaned up. Beyond lint, there are a few things worth addressing:

1. `report.go:331` has a self-comparison that's always false and was flagged with a TODO comment but left unfixed.
2. The update phase timing measurement returns immediately because the RBG is already Ready, so it doesn't capture the real reconciliation latency.
3. All scripts and templates are fully duplicated between `test/stress/` and `.claude/skills/stress-test/`, which will be a maintenance headache.

See inline comments for details.

test/stress/pprof.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+// IsAvailable checks if the pprof endpoint is reachable.
+func (p *PprofCollector) IsAvailable() bool {
+	client := &http.Client{Timeout: 3 * time.Second}
+	resp, err := client.Get(fmt.Sprintf("http://%s/debug/pprof/", p.addr))
+	if err != nil {
+		return false
+	}
+	resp.Body.Close()
```
cheyang: The lint job is failing with 10 errors. Five of them are `errcheck` violations for unchecked `Close()` calls (here, pprof.go:170, pprof.go:180, report.go:122, timing.go:192). The remaining are `prealloc` suggestions for `var errs []error` in create.go:85, delete.go:84, update.go:102, and `var durations []float64` in timing.go:125. These need to be fixed for CI to pass.

test/stress/report.go
```diff
@@ -0,0 +1,636 @@
...
+	for _, p := range phases {
+		if p.Failed > 0 {
+			issues = append(issues, fmt.Sprintf("[%s] %d/%d operations failed (%.1f%% failure rate)",
+				p.Name, p.Failed, p.Total, float64(p.Failed)/float64(p.Total)*100))
+		}
+		if p.P99Ms > 10000 {
+			issues = append(issues, fmt.Sprintf("[%s] P99 latency %.0fms exceeds 10s threshold — controller may be overloaded",
+				p.Name, p.P99Ms))
+		}
+		if p.Total > 0 && p.ActualQPS < p.ActualQPS*0.5 {
```
cheyang: `p.ActualQPS < p.ActualQPS*0.5` is always false for non-negative values. The comment inside the branch acknowledges this ("This check doesn't make sense as written, fix below") but no fix was applied. staticcheck flags this as SA9003. Either remove the dead branch or implement the intended comparison against the scenario's target QPS (which would require passing target QPS into `detectIssues`).

test/stress/update.go
```diff
@@ -0,0 +1,111 @@
...
+				recorder.Record(rec)
+				errCh <- fmt.Errorf("update %s: %w", name, err)
+				return
+			}
+
+			// Wait for Ready condition (after update propagates)
+			waitCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+			defer cancel()
+
+			if err := client.WaitForRBGReady(waitCtx, scenario.Namespace, name); err != nil {
```
cheyang, line 97: `WaitForRBGReady` polls for `Ready=True`, but after an update the RBG is likely *already* Ready before the controller processes the change. The poll returns immediately, so the measured "update latency" only captures the API write round-trip, not actual reconciliation time. Consider waiting for `metadata.generation > status.observedGeneration` after the update call, then waiting for them to match, which would give the real end-to-end time.

.claude/skills/stress-test/scripts/setup-kwok.sh
```diff
@@ -0,0 +1,99 @@
+#!/bin/bash
```
cheyang: All 5 shell scripts and kwok-stage.yaml are byte-for-byte identical between `test/stress/` and `.claude/skills/stress-test/`. Any future bug fix or enhancement must be applied in both places. Consider keeping one canonical copy (in `test/stress/`) and symlinking from `.claude/skills/stress-test/`, or just referencing the `test/stress/` paths in SKILL.md.

test/stress/templates/kwok-stage.yaml
```diff
@@ -0,0 +1,140 @@
...
+    apiGroup: v1
+    kind: Pod
+  selector:
+    matchExpressions:
+      - key: '.metadata.deletionTimestamp'
+        operator: 'DoesNotExist'
+      - key: '.status.podIP'
+        operator: 'DoesNotExist'
+  next:
+    event:
```
cheyang, line 19: The `pod-initialize` Stage has two `next:` keys at the same nesting level (lines 19 and 42). Most YAML parsers silently use the last value, making the first `next` block (with the scheduling event) dead. The file isn't applied by setup-kwok.sh (Helm stage-fast is used), but it's misleading as reference material. Split into separate Stage resources or merge into one `next` block.

.gitignore
```diff
@@ -26,14 +26,14 @@ main
 .idea
 .vscode
 .qoder
-.claude
+.claude/*
```
cheyang, line 30: The change from ignoring `.claude` to `.claude/*` + `!.claude/skills/` means all future `.claude/skills/` content from any contributor is tracked. Is this the intended team convention, or should only specific skills be tracked?

test/stress/client.go
```diff
@@ -0,0 +1,237 @@
...
+// WatchRBGs returns a watch interface for RBGs in the namespace.
+func (c *StressClient) WatchRBGs(ctx context.Context, namespace string) (watch.Interface, error) {
+	return c.dynamic.Resource(rbgGVR).Namespace(namespace).Watch(ctx, metav1.ListOptions{})
+}
+
+// GetControllerInfo fetches details about the deployed RBG controller.
+func (c *StressClient) GetControllerInfo(ctx context.Context) ControllerInfo {
+	info := ControllerInfo{}
+
+	deploy, err := c.clientset.AppsV1().Deployments("rbgs-system").Get(ctx, "rbgs-controller-manager", metav1.GetOptions{})
```
cheyang: The namespace `rbgs-system` and label `control-plane=rbgs-controller` are hardcoded. If the controller is deployed to a different namespace, log and pprof collection will silently fail. Adding `--controller-namespace` and `--controller-label` flags would make this more flexible.

test/stress/templates.go
```diff
@@ -0,0 +1,225 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+const (
+	testImage        = "registry.k8s.io/pause:3.9"
+	testImageUpdated = "registry.k8s.io/pause:3.10"
+)
+
+// GenerateRBG creates an unstructured RBG object for stress testing.
+func GenerateRBG(index int, scenario *Scenario) *unstructured.Unstructured {
```
cheyang, line 32: Functions like `GenerateRBG`, `computeStats`, `percentile`, and `ApplyUpdate` are pure and well-suited for table-driven unit tests. This would catch regressions (like the self-comparison bug) before they reach CI.

cheyang:
Round 2 review: all three previously blocking issues are resolved.

**Lint** -- errcheck (`_ = resp.Body.Close()`), prealloc (`make([]error, 0, cap)`), dead code branch (removed), and goconst (`labelValueTrue`) all fixed; CI passes.

**Script duplication** -- `.claude/skills/stress-test/scripts/` and `templates/kwok-stage.yaml` removed; SKILL.md now references `test/stress/` canonical paths.

**Concurrency** -- the rate-limiter-gated pattern is reasonable for a stress test client. Noted as a non-blocking observation for future improvement.

Additional improvements in this iteration look solid: update latency now correctly waits for `observedGeneration`, kwok-stage.yaml splits the duplicate `next:` into separate stages, controller namespace/label are configurable via flags, and good table-driven unit tests cover pure helper functions.

LGTM -- ready to merge.

test/stress/create.go
```diff
@@ -0,0 +1,94 @@
...
+	var wg sync.WaitGroup
+	errCh := make(chan error, scenario.TotalRBGs)
+
+	for i := 0; i < scenario.TotalRBGs; i++ {
+		if err := limiter.Wait(ctx); err != nil {
+			return fmt.Errorf("rate limiter: %w", err)
+		}
+
+		wg.Add(1)
+		go func(index int) {
```
cheyang, line 44: nit: The concurrency model spawns one goroutine per RBG with no upper bound on in-flight waiters. At `--total-rbgs=1000` and `--create-qps=10`, after ~100s all 1000 goroutines poll concurrently at 500ms intervals (up to 2000 req/s against the client's 200 QPS limit). For a stress test tool this is acceptable since client-side throttling absorbs the pressure, but adding a `--max-concurrent-waiters` flag would give operators a knob if they hit OOM or socket exhaustion on very large runs.

test/stress/client.go
```diff
@@ -0,0 +1,263 @@
...
+
+		if isRBGReady(rbg) {
+			return nil
+		}
+		time.Sleep(500 * time.Millisecond)
+	}
+}
+
+// WaitForRBGObservedGeneration polls until status.observedGeneration >= generation.
+func (c *StressClient) WaitForRBGObservedGeneration(ctx context.Context, namespace, name string, generation int64) error {
```
cheyang: nit: Consider using a watch-based wait instead of polling at 500ms for `WaitForRBGObservedGeneration` and `WaitForRBGReady` -- would reduce API server load during waits and improve latency accuracy. Not blocking since this is a test tool.

test/stress/create.go
Syspretor: Done in eabd1c5f. Added `--max-concurrent-waiters` flag (default 0 = unlimited). When set, a semaphore bounds in-flight goroutines across all three phases (create/update/delete), preventing unbounded polling at scale.

test/stress/client.go
Syspretor: Done in eabd1c5f. Replaced 500ms polling in `WaitForRBGReady`, `WaitForRBGObservedGeneration`, and `WaitForRBGDeleted` with watch-based waiting using `fieldSelector=metadata.name=<name>`. Each watch event is checked immediately — no sleep loop. Falls back to a single poll + retry if the watch encounters errors.

test/stress/client.go
```diff
@@ -0,0 +1,308 @@
...
+// GetControllerInfo fetches details about the deployed RBG controller.
+func (c *StressClient) GetControllerInfo(ctx context.Context, controllerNamespace, controllerLabel string) ControllerInfo {
+	info := ControllerInfo{}
+
+	deploy, err := c.clientset.AppsV1().Deployments(controllerNamespace).Get(ctx, "rbgs-controller-manager", metav1.GetOptions{})
+	if err != nil {
+		return info
+	}
+
+	info.Replicas = int(*deploy.Spec.Replicas)
```
cheyang, line 256: nit: `deploy.Spec.Replicas` is `*int32` and technically nullable. A nil guard avoids a panic if someone applies a Deployment manifest without an explicit replicas field:

```go
if deploy.Spec.Replicas != nil {
    info.Replicas = int(*deploy.Spec.Replicas)
} else {
    info.Replicas = 1
}
```

test/stress/client.go
```diff
@@ -0,0 +1,308 @@
...
+				}
+			case watch.Error:
+				return false, nil // retry watch
+			}
+		}
+	}
+}
+
+// WatchRBGs returns a watch interface for RBGs in the namespace.
+func (c *StressClient) WatchRBGs(ctx context.Context, namespace string) (watch.Interface, error) {
```
cheyang, line 243: nit: `WatchRBGs` (namespace-wide watch) is no longer called after the per-resource `watchAndWait` refactor. Remove if there is no planned use.

cheyang (approved):
Latest commit (watch-based waits + semaphore) looks good — addresses the remaining nits cleanly. Standard K8s watch pattern, correct fallback, consistent integration across all three phases. Two minor nits below, neither blocking.

cheyang:
Round 3 (incremental review of eabd1c5 vs 97968bd): The new commit replaces all three polling-based waits with a unified `watchAndWait` pattern using per-resource K8s watches with correct RV-based startup and error fallback. The channel-based semaphore (`--max-concurrent-waiters`) is cleanly integrated across create/update/delete phases. Both address the round-2 nits. Two prior non-blocking nits (remove unused `WatchRBGs`, nil-guard `deploy.Spec.Replicas`) remain open but are already communicated. All CI checks pass. No new issues.

PR #357, opened 2026-06-02
refactor: move restartPolicy from RoleSpec to pattern types

Syspretor:
## Summary
- Move `restartPolicy` from `RoleSpec` (role level) into `LeaderWorkerPattern` and `CustomComponentsPattern`, since it only semantically applies to multi-pod patterns
- `StandalonePattern` is hardcoded to `None` (single pod — recreating instance is equivalent to normal pod replacement)
- Default for `LeaderWorkerPattern` and `CustomComponentsPattern` remains `RecreateRoleInstanceOnPodRestart`

## Changes
- **API types**: Remove `RestartPolicy` from `RoleSpec`, add to `LeaderWorkerPattern` and `CustomComponentsPattern`, add `GetRestartPolicy()` helper
- **Conversion**: Update v1alpha1 ↔ v1alpha2 conversion to map `restartPolicy` into/from pattern types
- **Reconcilers**: Use `role.GetRestartPolicy()` instead of `role.RestartPolicy`
- **client-go**: Move `WithRestartPolicy` from `RoleSpec` to pattern apply configs
- **CRDs**: Regenerated via `make generate && make manifests`
- **Tests**: Update e2e tests and wrappers to match new API shape
- **Examples & docs**: Move `restartPolicy` to pattern level in all YAMLs, update `api.md` and `failure-handling.md`

## Migration / Breaking change note

> **v1alpha2 is the storage version.** After this CRD upgrade, the `restartPolicy` field at `RoleSpec` level will be pruned from objects in etcd.
>
> - For **LeaderWorkerPattern** and **CustomComponentsPattern** roles: if `restartPolicy` was unset or `RecreateRoleInstanceOnPodRestart`, no behavior change — the new pattern-level default matches.
> - For roles that **explicitly set `restartPolicy: None` on a LeaderWorkerPattern/CustomComponentsPattern**: after upgrade the field is pruned and the default flips to `RecreateRoleInstanceOnPodRestart`. Users should re-apply their manifests with `restartPolicy: None` under the pattern to preserve the original behavior.
> - For **StandalonePattern** roles: no impact — standalone was always effectively `None`.
>
> This is an alpha API, so field-level migration is expected. Users should review their manifests after upgrading.

## Test plan
- [x] `make build` passes
- [x] `make lint` passes
- [x] `make test` passes (all unit tests including conversion round-trip tests)
- [x] `make test-e2e` — v1alpha1 revision and v1alpha2 tests pass

### DIFF
--- a/api/workloads/v1alpha2/helper.go
+++ b/api/workloads/v1alpha2/helper.go
@@ -269,6 +269,26 @@ func (r *RoleSpec) GetWorkerTemplatePatch() *runtime.RawExtension {
 	return r.LeaderWorkerPattern.WorkerTemplatePatch
 }
 
+// GetRestartPolicy returns the effective restart policy for this role based on its pattern.
+// StandalonePattern always returns None (single pod, no instance-level restart).
+// LeaderWorkerPattern and CustomComponentsPattern default to RecreateRoleInstanceOnPodRestart.
+func (r *RoleSpec) GetRestartPolicy() RestartPolicyType {
+	if r.LeaderWorkerPattern != nil {
+		if r.LeaderWorkerPattern.RestartPolicy != "" {
+			return r.LeaderWorkerPattern.RestartPolicy
+		}
+		return RecreateRoleInstanceOnPodRestart
+	}
+	if r.CustomComponentsPattern != nil {
+		if r.CustomComponentsPattern.RestartPolicy != "" {
+			return r.CustomComponentsPattern.RestartPolicy
+		}
+		return RecreateRoleInstanceOnPodRestart
+	}
+	// StandalonePattern or no pattern: single pod, no instance-level restart
+	return RestartPolicyNone
+}
+
 // GetLeaderWorkerSize returns the size of the leader-worker group.
 // Returns nil if not using LeaderWorkerPattern.
 func (r *RoleSpec) GetLeaderWorkerSize() *int32 {
--- a/api/workloads/v1alpha1/rolebasedgroup_conversion.go
+++ b/api/workloads/v1alpha1/rolebasedgroup_conversion.go
@@ -160,7 +160,7 @@ func convertRoleV1alpha1ToV2(src *RoleSpec, dst *v2.RoleSpec) error {
 			Size:                src.LeaderWorkerSet.Size,
 			LeaderTemplatePatch: src.LeaderWorkerSet.PatchLeaderTemplate,
 			WorkerTemplatePatch: src.LeaderWorkerSet.PatchWorkerTemplate,
-			RestartPolicy:       v2.RestartPolicyType(src.RestartPolicy),
+			RestartPolicy:       convertRestartPolicyV1alpha1ToV2(src.RestartPolicy),
 		}
 		// Template source
 		if src.TemplateSource.Template != nil || src.TemplateSource.TemplateRef != nil {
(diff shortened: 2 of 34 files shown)


api/workloads/v1alpha2/helper.go
```diff
@@ -269,6 +269,26 @@ func (r *RoleSpec) GetWorkerTemplatePatch() *runtime.RawExtension {
...
 
+// GetRestartPolicy returns the effective restart policy for this role based on its pattern.
+// StandalonePattern always returns None (single pod, no instance-level restart).
+// LeaderWorkerPattern and CustomComponentsPattern default to RecreateRoleInstanceOnPodRestart.
+func (r *RoleSpec) GetRestartPolicy() RestartPolicyType {
+	if r.LeaderWorkerPattern != nil {
+		if r.LeaderWorkerPattern.RestartPolicy != "" {
+			return r.LeaderWorkerPattern.RestartPolicy
+		}
+		return RecreateRoleInstanceOnPodRestart
```
cheyang, line 280: Worth flagging for the release notes: v1alpha2 is the storage version, so any RBG already in etcd that set `restartPolicy` at the RoleSpec level will have that field pruned on the CRD upgrade. For most cases the effective default lines up, but if someone explicitly set `None` on a LeaderWorkerPattern role, after upgrade it falls through to this branch and flips to `RecreateRoleInstanceOnPodRestart` with no warning. Probably fine for an alpha API, but it'd help to call this out so users can re-check their manifests.

api/workloads/v1alpha2/helper.go
```diff
@@ -269,6 +269,26 @@ func (r *RoleSpec) GetWorkerTemplatePatch() *runtime.RawExtension {
 	return r.LeaderWorkerPattern.WorkerTemplatePatch
 }
 
+// GetRestartPolicy returns the effective restart policy for this role based on its pattern.
+// StandalonePattern always returns None (single pod, no instance-level restart).
+// LeaderWorkerPattern and CustomComponentsPattern default to RecreateRoleInstanceOnPodRestart.
+func (r *RoleSpec) GetRestartPolicy() RestartPolicyType {
```
cheyang, line 275: This will panic if `r` is nil. The other RoleSpec helpers here don't nil-check either, so this matches the existing style and the callers iterating over `rbg.Spec.Roles` are safe — not asking you to change it, just noting it since gemini-code-assist raised the same point.

api/workloads/v1alpha2/helper.go
Syspretor: Good call — added a **Migration / Breaking change note** section to the PR description covering this exact scenario. Since this is an alpha API, the pruning behavior is expected, but users who explicitly set `None` on LWP/CCP roles should re-apply their manifests with the field under the pattern.

api/workloads/v1alpha2/helper.go
Syspretor: Agreed — keeping it consistent with the existing style. All callers iterate over `rbg.Spec.Roles` so `r` is always a valid pointer.

api/workloads/v1alpha2/helper.go
```diff
@@ -269,6 +269,26 @@ func (r *RoleSpec) GetWorkerTemplatePatch() *runtime.RawExtension {
...
+	}
+	if r.CustomComponentsPattern != nil {
+		if r.CustomComponentsPattern.RestartPolicy != "" {
+			return r.CustomComponentsPattern.RestartPolicy
+		}
+		return RecreateRoleInstanceOnPodRestart
+	}
+	// StandalonePattern or no pattern: single pod, no instance-level restart
+	return RestartPolicyNone
+}
```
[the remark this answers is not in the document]
Syspretor: The existing RoleSpec helpers (`GetWorkloadSpec`, `GetTemplate`, etc.) don't nil-check `r` either — all callers iterate over `rbg.Spec.Roles` which guarantees non-nil. Keeping the style consistent.

cheyang (approved):

PR #360, opened 2026-06-03
fix: update restart policy tests to match pattern-level RestartPolicy API

Syspretor:
## Summary
- PR #357 moved `RestartPolicy` from `RoleSpec` to pattern types but `StandalonePattern` does not support it (defaults to `None`). The envtest `restart_policy` tests still called `WithRestartPolicy` on `StandaloneRoleWrapper`, causing compilation errors that broke lint/unit-test/envtest CI on PR #358
- Changed test cases requiring `RestartPolicy` to use `LeaderWorkerPattern` with `Size=1` (single pod per instance) instead of `StandalonePattern`
- Added a dedicated `StandalonePattern` test verifying the default `None` behavior: pod failure does not trigger instance-level recreation and `Restarting` condition is not set

## Test plan
- [x] `go vet ./...` passes
- [x] `make test` — all unit tests pass
- [x] `make test-envtest` — all 11 restart_policy specs pass (10 existing + 1 new standalone test)

### DIFF
--- a/api/workloads/v1alpha2/helper.go
+++ b/api/workloads/v1alpha2/helper.go
@@ -270,7 +270,7 @@ func (r *RoleSpec) GetWorkerTemplatePatch() *runtime.RawExtension {
 }
 
 // GetRestartPolicy returns the effective restart policy for this role based on its pattern.
-// StandalonePattern defaults to None.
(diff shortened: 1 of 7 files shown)


cheyang (approved):
/lgtm
/approve

PR #361, opened 2026-06-03
ci: build images when helm Chart appVersion changes

Syspretor:
## Summary
- Add a GitHub Actions workflow that triggers image builds when `deploy/helm/rbgs/Chart.yaml` `appVersion` is updated on main
- Extracts the commit hash from `appVersion` (format: `x.y.z-<hash>`), finds the main branch commit before the merge, and builds multi-arch images at that commit
- Uses `make docker-buildx-push` to build and push `rbgs-controller`, `rbgs-upgrade-crd`, and `rbgs-patio-runtime` images tagged with the full `appVersion`

## Test plan
- [ ] Verify workflow triggers on Chart.yaml appVersion change
- [ ] Verify commit hash extraction logic works correctly
- [ ] Verify merge-base detection finds the correct main commit
- [ ] Verify images are built and pushed with correct tags

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -15,7 +15,9 @@ env:
   SCHEDULER_PLUGINS_VERSION: v0.32.7
 
 # Declare default permissions as read only.
-permissions: read-all
+permissions:
+  contents: read
+  actions: read
 
 jobs:
   check-changes:
@@ -250,6 +252,50 @@ jobs:
       run:
         working-directory: ${{ env.GOPATH }}/src/sigs.k8s.io/rbgs
     steps:
+      - name: Wait for image build workflow (if running)
+        if: github.event_name == 'pull_request'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          echo "Checking if 'Build Image on Helm Release Update' workflow is running for this PR..."
+          HEAD_SHA="${{ github.event.pull_request.head.sha }}"
+
+          # Wait up to 30 minutes for the image build workflow to complete
+          MAX_ATTEMPTS=60
+          SLEEP_INTERVAL=30
+
+          for i in $(seq 1 $MAX_ATTEMPTS); do
+            # Find workflow runs for this commit
+            RUN_INFO=$(gh api repos/${{ github.repository }}/actions/workflows/helm-release-image.yml/runs \
+              --jq ".workflow_runs[] | select(.head_sha == \"$HEAD_SHA\") | {id: .id, status: .status, conclusion: .conclusion}" \
+              2>/dev/null | head -n 1)
+
+            if [ -z "$RUN_INFO" ]; then
+              echo "No image build workflow run found for this PR, proceeding."
+              break
+            fi
+
+            STATUS=$(echo "$RUN_INFO" | jq -r '.status')
+            CONCLUSION=$(echo "$RUN_INFO" | jq -r '.conclusion')
+            RUN_ID=$(echo "$RUN_INFO" | jq -r '.id')
+
+            if [ "$STATUS" == "completed" ]; then
+              if [ "$CONCLUSION" == "success" ]; then
+                echo "Image build workflow completed successfully (run: $RUN_ID)."
+              else
+                echo "::warning::Image build workflow completed with conclusion: $CONCLUSION (run: $RUN_ID)"
+              fi
+              break
+            fi
+
+            echo "Image build workflow is still running (status: $STATUS, run: $RUN_ID). Waiting ${SLEEP_INTERVAL}s... (attempt $i/$MAX_ATTEMPTS)"
+            sleep $SLEEP_INTERVAL
+          done
+
+          if [ "$i" -eq "$MAX_ATTEMPTS" ] && [ "$STATUS" != "completed" ]; then
+            echo "::warning::Timed out waiting for image build workflow. Proceeding anyway."
+          fi
+
       - name: Set up Go
         uses: actions/setup-go@v5.5.0
         with:
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -254,30 +254,34 @@ jobs:
     steps:
       - name: Wait for image build workflow (if running)
         if: github.event_name == 'pull_request'
+        working-directory: ${{ github.workspace }}
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
           echo "Checking if 'Build Image on Helm Release Update' workflow is running for this PR..."
           HEAD_SHA="${{ github.event.pull_request.head.sha }}"
 
+          # Wait a bit for the other workflow to be queued
+          echo "Waiting 60s for image build workflow to be queued..."
+          sleep 60
+
           # Wait up to 30 minutes for the image build workflow to complete
           MAX_ATTEMPTS=60
           SLEEP_INTERVAL=30
+          STATUS=""
 
           for i in $(seq 1 $MAX_ATTEMPTS); do
-            # Find workflow runs for this commit
-            RUN_INFO=$(gh api repos/${{ github.repository }}/actions/workflows/helm-release-image.yml/runs \
-              --jq ".workflow_runs[] | select(.head_sha == \"$HEAD_SHA\") | {id: .id, status: .status, conclusion: .conclusion}" \
-              2>/dev/null | head -n 1)
+            # Find the latest workflow run for this commit
+            RUN_ID=$(gh api "repos/${{ github.repository }}/actions/workflows/helm-release-image.yml/runs?head_sha=$HEAD_SHA" \
+              --jq '.workflow_runs[0].id // empty' 2>/dev/null)
 
-            if [ -z "$RUN_INFO" ]; then
+            if [ -z "$RUN_ID" ]; then
               echo "No image build workflow run found for this PR, proceeding."
               break
             fi
 
-            STATUS=$(echo "$RUN_INFO" | jq -r '.status')
-            CONCLUSION=$(echo "$RUN_INFO" | jq -r '.conclusion')
-            RUN_ID=$(echo "$RUN_INFO" | jq -r '.id')
+            STATUS=$(gh api "repos/${{ github.repository }}/actions/runs/$RUN_ID" --jq '.status' 2>/dev/null)
+            CONCLUSION=$(gh api "repos/${{ github.repository }}/actions/runs/$RUN_ID" --jq '.conclusion // empty' 2>/dev/null)
 
             if [ "$STATUS" == "completed" ]; then
               if [ "$CONCLUSION" == "success" ]; then
--- a/.github/workflows/helm-release-image.yml
+++ b/.github/workflows/helm-release-image.yml
@@ -1,7 +1,7 @@
 name: Build Image on Helm Release Update
 
 on:
-  pull_request:
+  pull_request_target:
     branches:
       - main
     paths:
@@ -19,7 +19,7 @@ jobs:
   check-appversion-change:
     runs-on: ubuntu-latest
     if: |
-      (github.event_name == 'pull_request' &&
+      (github.event_name == 'pull_request_target' &&
        contains(fromJSON('["Syspretor", "cheyang"]'), github.event.pull_request.user.login)) ||
       (github.event_name == 'workflow_dispatch' &&
        contains(fromJSON('["Syspretor", "cheyang"]'), github.actor))
@@ -28,9 +28,10 @@ jobs:
       app_version: ${{ steps.check.outputs.app_version }}
       commit_hash: ${{ steps.check.outputs.commit_hash }}
     steps:
-      - name: Checkout code
+      - name: Checkout PR head
         uses: actions/checkout@v4
         with:
+          ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 0
 
       - name: Check if appVersion changed
@@ -41,6 +42,7 @@ jobs:
           echo "Current appVersion: $CURRENT_APP_VERSION"
 
           # Get appVersion from base branch (main)
+          git fetch origin main
           PREV_APP_VERSION=$(git show origin/main:deploy/helm/rbgs/Chart.yaml 2>/dev/null | grep '^appVersion:' | awk '{print $2}' || echo "")
           echo "Base branch appVersion: $PREV_APP_VERSION"
 
--- a/.github/workflows/helm-release-image.yml
+++ b/.github/workflows/helm-release-image.yml
@@ -1,7 +1,7 @@
 name: Build Image on Helm Release Update
 
 on:
-  push:
+  pull_request:
     branches:
       - main
     paths:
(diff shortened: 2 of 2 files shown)


.github/workflows/e2e-test.yml
```diff
@@ -250,6 +252,54 @@ jobs:
...
+          sleep 60
+
+          # Wait up to 30 minutes for the image build workflow to complete
+          MAX_ATTEMPTS=60
+          SLEEP_INTERVAL=30
+          STATUS=""
+
+          for i in $(seq 1 $MAX_ATTEMPTS); do
+            # Find the latest workflow run for this commit
+            RUN_ID=$(gh api "repos/${{ github.repository }}/actions/workflows/helm-release-image.yml/runs?head_sha=$HEAD_SHA" \
```
cheyang, line 275: This step is what's currently failing `e2e-test-manifest`, and it will fail on every PR. GitHub runs `run:` blocks with `bash -eo pipefail`, so when this `gh api` call returns non-zero the whole job aborts. Right now `helm-release-image.yml` doesn't exist on the default branch yet, so the API returns 404 and the step dies right after the 60s sleep (the run log shows `Process completed with exit code 1`). Note that `2>/dev/null` only hides stderr and `// empty` is a jq fallback, so neither affects the shell exit code. Please guard the call, e.g. `RUN_ID=$(gh api ... 2>/dev/null || echo "")`, and do the same for the `STATUS`/`CONCLUSION` queries below so a transient API error can't take down the manifest job.

.github/workflows/e2e-test.yml
```diff
@@ -250,6 +252,54 @@ jobs:
...
         working-directory: ${{ env.GOPATH }}/src/sigs.k8s.io/rbgs
     steps:
+      - name: Wait for image build workflow (if running)
+        if: github.event_name == 'pull_request'
+        working-directory: ${{ github.workspace }}
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          echo "Checking if 'Build Image on Helm Release Update' workflow is running for this PR..."
+          HEAD_SHA="${{ github.event.pull_request.head.sha }}"
```
cheyang, line 262: Heads up that this lookup probably won't match once the workflow does exist. `helm-release-image.yml` is triggered by `pull_request_target`, so its run's `head_sha` is the base-branch SHA, not `pull_request.head.sha`. Querying `?head_sha=$HEAD_SHA` with the PR head will usually return no run, so the wait quietly becomes a no-op. If the goal is to actually block on the image build, you'll likely need to correlate on something else (branch, a created-after timestamp, or a dispatched run id).

.github/workflows/helm-release-image.yml
```diff
@@ -0,0 +1,151 @@
...
+
+          # Find the full commit hash
+          FULL_HASH=$(git rev-parse "$COMMIT_HASH")
+          echo "Full commit hash: $FULL_HASH"
+
+          # Find the merge commit that brought this commit into main.
+          # The commit referenced by appVersion is the HEAD of a feature branch
+          # that was merged into main. We want the main branch commit just before
+          # that merge (i.e., the parent of the merge commit on main side).
+          MERGE_COMMIT=$(git log --merges --ancestry-path "${FULL_HASH}..main" --format='%H' --reverse | head -n 1)
```
cheyang, line 88: This target resolution assumes a merge-commit history, but the repo squash-merges: recent `main` is all squashed commits with no merge commits. So `git log --merges --ancestry-path "${FULL_HASH}..main"` finds nothing and you fall through to `${FULL_HASH}~1`, which builds the *parent* of the intended commit. Since the hash embedded in `appVersion` is itself a direct commit on `main`, the commit you actually want to build is `FULL_HASH`, not its parent. Can you confirm the intended commit? This likely simplifies to just checking out `FULL_HASH`.

.github/workflows/helm-release-image.yml
```diff
@@ -0,0 +1,151 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  check-appversion-change:
+    runs-on: ubuntu-latest
+    if: |
+      (github.event_name == 'pull_request_target' &&
+       contains(fromJSON('["Syspretor", "cheyang"]'), github.event.pull_request.user.login)) ||
```
cheyang, line 23: Since this is a `pull_request_target` workflow that holds the Docker Hub push secrets, this allowlist is effectively the only thing gating a privileged, secret-bearing job, and it's hardcoded in two places. Two suggestions: (a) move the list to a repo/org variable so it can change without editing the workflow (`vars` is available in `if:`, whereas `secrets` is not, so a secret wouldn't work here anyway); and (b) for real hardening, put the `build-and-push` job behind a protected environment with required reviewers, so even a compromised allowlisted account can't push images without human approval.

.github/workflows/helm-release-image.yml
```diff
@@ -0,0 +1,151 @@
...
+          PREV_APP_VERSION=$(git show origin/main:deploy/helm/rbgs/Chart.yaml 2>/dev/null | grep '^appVersion:' | awk '{print $2}' || echo "")
+          echo "Base branch appVersion: $PREV_APP_VERSION"
+
+          if [ "$CURRENT_APP_VERSION" != "$PREV_APP_VERSION" ]; then
+            echo "appVersion changed from '$PREV_APP_VERSION' to '$CURRENT_APP_VERSION'"
+            echo "changed=true" >> $GITHUB_OUTPUT
+            echo "app_version=$CURRENT_APP_VERSION" >> $GITHUB_OUTPUT
+
+            # Extract commit hash from appVersion (format: x.y.z-<hash>)
+            COMMIT_HASH=$(echo "$CURRENT_APP_VERSION" | sed 's/.*-//')
```
cheyang, line 55: `appVersion` comes from the PR's `Chart.yaml`, so it's attacker-controlled input, and it flows unvalidated into a few sensitive places: into `$GITHUB_OUTPUT` (lines 52/57, where a multi-line value can inject extra outputs), and later into `TAG`, which `make docker-buildx-push` expands into a shell `docker ... --tag repo:${TAG}` in the secret-bearing build job (shell-metachar injection). Recommend validating the format strictly before using it, e.g. fail unless it matches something like `^[0-9]+\.[0-9]+\.[0-9]+-[0-9a-f]{7,40}$`, and only then write it to outputs / pass it to `make`.

.github/workflows/e2e-test.yml
cheyang, line 275: This `gh api` call will exit non-zero if the workflow file doesn't exist yet (HTTP 404), and `2>/dev/null` only suppresses stderr — it doesn't mask the exit code. With `set -e` (default in GHA bash), this aborts the step.

This is what's causing the `e2e-test-manifest` CI failure right now.

Fix:
```bash
RUN_ID=$(gh api "repos/${{ github.repository }}/actions/workflows/helm-release-image.yml/runs?head_sha=$HEAD_SHA" \
  --jq '.workflow_runs[0].id // empty' 2>/dev/null || true)
```

.github/workflows/e2e-test.yml
```diff
@@ -250,6 +252,54 @@ jobs:
       run:
         working-directory: ${{ env.GOPATH }}/src/sigs.k8s.io/rbgs
     steps:
+      - name: Wait for image build workflow (if running)
+        if: github.event_name == 'pull_request'
```
cheyang, line 256: This condition triggers for **all** pull requests, meaning every PR incurs a mandatory 60-second sleep before the e2e-test-manifest job proceeds — even PRs that don't touch `Chart.yaml` at all.

Consider gating this on whether Chart.yaml was actually modified in the PR (e.g., check `git diff --name-only` or pass an output from the `check-changes` job).

.github/workflows/helm-release-image.yml
```diff
@@ -0,0 +1,151 @@
...
+          # that merge (i.e., the parent of the merge commit on main side).
+          MERGE_COMMIT=$(git log --merges --ancestry-path "${FULL_HASH}..main" --format='%H' --reverse | head -n 1)
+
+          if [ -n "$MERGE_COMMIT" ]; then
+            # Get the first parent of the merge commit (the main branch side)
+            TARGET_COMMIT=$(git rev-parse "${MERGE_COMMIT}^1")
+            echo "Merge commit: $MERGE_COMMIT"
+            echo "Target commit (main before merge): $TARGET_COMMIT"
+          else
+            # If no merge commit found, the commit might already be on main directly.
```
cheyang, line 96: `MERGE_COMMIT^1` resolves to the first parent of the merge commit, which is the main branch state *before* the feature was merged. This means the build excludes all the feature branch's code changes.

If the intent is to build the code that the appVersion references (i.e., the state after the feature is on main), you likely want either:
- The merge commit itself: `TARGET_COMMIT=$MERGE_COMMIT`
- Or the commit hash directly: `TARGET_COMMIT=$FULL_HASH`

Can you clarify what state should actually be built here?

.github/workflows/helm-release-image.yml
```diff
@@ -0,0 +1,151 @@
...
+      commit_hash: ${{ steps.check.outputs.commit_hash }}
+    steps:
+      - name: Checkout PR head
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+
+      - name: Check if appVersion changed
+        id: check
```
cheyang, line 38: Heads-up: checking out `pull_request.head.sha` in a `pull_request_target` workflow gives the PR's (potentially untrusted) code access to secrets. The user allowlist on line 27 mitigates this, but if the allowlist ever gets stale or is bypassed, this becomes a secret-exfiltration vector.

Since this job only reads `Chart.yaml`, consider whether you can use `github.event.pull_request.merge_commit_sha` or limit the checkout to the specific file needed.

cheyang (changes requested):
Reviewed the CI workflow for building images when Chart appVersion changes. The overall approach makes sense, but there are several issues that need to be addressed before this can be merged.

**CI status:** `e2e-test-manifest` is failing (non-flaky) due to a bug in the wait step introduced here.

**Summary of issues:**
1. The `gh api` call in the wait step lacks error handling and fails the job when the workflow file isn't found (HTTP 404 causes non-zero exit under `set -e`).
2. The unconditional 60-second sleep runs on ALL pull requests, not just those modifying Chart.yaml. This penalizes every PR's CI runtime.
3. The target commit logic in `build-and-push` builds from `MERGE_COMMIT^1` (i.e., main *before* the merge), which excludes the feature's code changes. This appears incorrect unless there's a specific reason documented.
4. `pull_request_target` + checkout of PR head gives secrets access to PR code. Mitigated by user allowlist, but worth a note.

cheyang:
The new `helm-release-image` workflow approach is sound, but the current implementation breaks `e2e-test-manifest` CI on every PR. The `gh api` call in the wait step exits non-zero under `set -e` when no matching workflow run exists, and the unconditional 60s sleep delays all deploy-touching PRs unnecessarily. The target-commit logic in `build-and-push` also appears to build from the wrong commit (`MERGE_COMMIT^1` instead of the appVersion commit itself). These need to be addressed before merging.

.github/workflows/e2e-test.yml
```diff
@@ -250,6 +252,54 @@ jobs:
...
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          echo "Checking if 'Build Image on Helm Release Update' workflow is running for this PR..."
+          HEAD_SHA="${{ github.event.pull_request.head.sha }}"
+
+          # Wait a bit for the other workflow to be queued
+          echo "Waiting 60s for image build workflow to be queued..."
+          sleep 60
+
+          # Wait up to 30 minutes for the image build workflow to complete
```
cheyang, line 268: GitHub Actions runs bash with `set -eo pipefail` by default. When `helm-release-image.yml` doesn't exist yet (this PR adds it) or no run matches, `gh api` returns a non-zero exit code and kills the step before the `-z "$RUN_ID"` guard is evaluated.

This is the root cause of the current `e2e-test-manifest` CI failure.

Fix: append `|| true` to the `gh api` invocation, or wrap the entire polling block in `set +e` / `set -e`.

```suggestion
            RUN_ID=$(gh api "repos/${{ github.repository }}/actions/workflows/helm-release-image.yml/runs?head_sha=$HEAD_SHA" \
              --jq '.workflow_runs[0].id // empty' 2>/dev/null || true)
```

.github/workflows/e2e-test.yml
cheyang, line 262: The wait step is gated on `github.event_name == 'pull_request'` but not on whether `deploy/helm/rbgs/Chart.yaml` was actually modified. Every PR that triggers `e2e-test-manifest` (i.e. any deploy/ change) now pays a mandatory 60-second delay even when the image build workflow was never queued.

Consider conditioning the sleep (and the entire polling loop) on a check that Chart.yaml is among the changed files, or at least move the sleep inside the loop after confirming a run actually exists.

.github/workflows/helm-release-image.yml
```diff
@@ -0,0 +1,151 @@
...
+          COMMIT_HASH="${{ needs.check-appversion-change.outputs.commit_hash }}"
+          echo "Looking for commit: $COMMIT_HASH"
+
+          # Find the full commit hash
+          FULL_HASH=$(git rev-parse "$COMMIT_HASH")
+          echo "Full commit hash: $FULL_HASH"
+
+          # Find the merge commit that brought this commit into main.
+          # The commit referenced by appVersion is the HEAD of a feature branch
+          # that was merged into main. We want the main branch commit just before
```
cheyang, line 86: The target commit logic finds the merge commit that brought the appVersion commit into main, then checks out its first parent (`MERGE_COMMIT^1`). That parent is the main branch state *before* the merge -- it does not contain the feature branch changes. If the intent is to build the code at the appVersion commit, `git checkout $FULL_HASH` would be more direct. If there's a deliberate reason for using the pre-merge state, a comment explaining it would help future readers.

.github/workflows/helm-release-image.yml
```diff
@@ -0,0 +1,151 @@
+name: Build Image on Helm Release Update
+
+on:
+  pull_request_target:
+    branches:
+      - main
+    paths:
```
cheyang, line 7: Using `pull_request_target` gives the workflow access to repo secrets, and the `check-appversion-change` job checks out `github.event.pull_request.head.sha`. The hardcoded user allowlist (`Syspretor`, `cheyang`) mitigates arbitrary fork exploitation, but a compromised allowed account or a future allowlist expansion could expose Docker Hub credentials to untrusted code. Worth documenting the trust boundary here with a code comment.

.github/workflows/e2e-test.yml
```diff
@@ -15,7 +15,9 @@ env:
   SCHEDULER_PLUGINS_VERSION: v0.32.7
 
 # Declare default permissions as read only.
-permissions: read-all
+permissions:
```
cheyang, line 18: The original `permissions: read-all` also granted `statuses: read`, `checks: read`, `packages: read`, etc. The replacement only grants `contents: read` and `actions: read`. Verify that no existing step in this workflow depends on the broader set of read permissions (e.g., status checks, package reads).

PR #364, opened 2026-06-08
feat: add ua for every controller

vie-serendipity:
### Ⅰ. Motivation

All controllers used `mgr.GetClient()` directly, so their API requests could not be distinguished by controller in apiserver audit logs.

### Ⅱ. Modifications

Added `NewClientWithUserAgent(mgr, name)` in `pkg/utils/client/client.go`, which returns a cache-backed client tagged with a per-controller UserAgent. The UA follows the client-go `DefaultKubernetesUserAgent` convention: `<name>/<version> (<os>/<arch>) rbg-controller/<commit>` (version/commit from the `version` package).

NONE

No new tests. This is a pure observability change.

Run the controller against a cluster and check the apiserver audit log.

## Checklist

### DIFF
--- a/pkg/utils/client/client.go
+++ b/pkg/utils/client/client.go
@@ -18,10 +18,15 @@ package client
 
 import (
 	"fmt"
+	"runtime"
+	"strings"
 
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	"sigs.k8s.io/rbgs/api/workloads/constants"
+	"sigs.k8s.io/rbgs/version"
 )
 
 func NewClientFromManager(mgr manager.Manager, name string) client.Client {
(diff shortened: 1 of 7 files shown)


pkg/utils/client/client.go
```diff
@@ -36,3 +41,44 @@ func NewClientFromManager(mgr manager.Manager, name string) client.Client {
 	})
 	return delegatingClient
 }
+
+func NewClientWithUserAgent(mgr manager.Manager, name string) client.Client {
+	cfg := rest.CopyConfig(mgr.GetConfig())
+	cfg.UserAgent = buildUserAgent(name)
+
+	delegatingClient, _ := client.New(cfg, client.Options{
```
cheyang, line 49: The error from `client.New` is discarded here. With Scheme and Mapper both provided, failure is unlikely — but if it does happen the nil client will panic downstream with no useful context.

A small guard would make startup failures much easier to diagnose:
```go
delegatingClient, err := client.New(cfg, client.Options{...})
if err != nil {
    panic(fmt.Sprintf("failed to create client for %s: %v", name, err))
}
```

(Same pattern exists in `NewClientFromManager` above — not introduced by this PR, but worth fixing while you are here.)

pkg/utils/client/client.go
```diff
@@ -36,3 +41,44 @@ func NewClientFromManager(mgr manager.Manager, name string) client.Client {
...
+	return fmt.Sprintf("%s/%s (%s/%s) %s/%s",
+		name,
+		adjustVersion(version.Version),
+		runtime.GOOS,
+		runtime.GOARCH,
+		constants.ControllerName,
+		adjustCommit(version.GitCommit))
+}
+
+func adjustVersion(v string) string {
```
cheyang, line 69: Nit: this strips everything after the first `-`, so `v0.7.0-alpha.3` becomes `v0.7.0`. Is that intentional? During operational triage it can be useful to know whether you are running an alpha or a release build.

vie-serendipity: I directly copy this function from k8s.io/client-go/rest/config.go. I'm not sure the adjustVersion stripping is necessary here.

```golang
// adjustVersion strips "alpha", "beta", etc. from version in form
// major.minor.patch-[alpha|beta|etc].
func adjustVersion(v string) string {
	if len(v) == 0 {
		return "unknown"
	}
	seg := strings.SplitN(v, "-", 2)
	return seg[0]
}
```

pkg/utils/client/client.go
vie-serendipity: OK

cheyang (approved):
LGTM overall — the structured user-agent per controller is a solid observability win for audit logs and debugging.

A few minor observations (non-blocking):

1. **Error suppression in `NewClientWithUserAgent`** — `client.New` returns an error that's discarded with `_`. I see `NewClientFromManager` does the same, so this is a pre-existing pattern. In practice this call is unlikely to fail after manager startup, but a `panic` or early-exit would make the failure mode explicit rather than producing a nil-client crash downstream.

2. **Two similar factory functions** — After this PR, both `NewClientFromManager` and `NewClientWithUserAgent` coexist with slightly different behaviors (the old one passes `Unstructured: true`, the new one passes `Scheme`/`Mapper` explicitly). Worth considering whether `NewClientFromManager` should be updated to use `buildUserAgent` too, or deprecated in a follow-up, to keep a single code path.

3. **statefulmode controller not updated** — `statefulmode/stateful_instance_set_controller.go` still uses `NewClientFromManager`. If the goal is to have a consistent UA across all controllers, that one could be migrated in a follow-up.

4. **Unit tests** — `buildUserAgent`, `adjustVersion`, and `adjustCommit` are simple but testable. Adding a small `_test.go` would prevent regressions if the format evolves later.

None of these block merge. Nice contribution.

cheyang (approved):
LGTM. Clean implementation adding structured user-agent strings across all controllers. All CI green.

PR #368, opened 2026-06-15
feat(rbg): support inplace scheduling

NoobDream2568:
### Ⅰ. Motivation

This PR implements **Inplace Scheduling** — a feature that injects nodeAffinity into recreated Pods so they are steered toward nodes where the same workload has previously run, enabling reuse of node-local cached resources.

#### Core Implementation

- **`node_binding.go`** — New file implementing the in-place scheduling engine:
  - `NodeBindingStore` — two-level `map[uid]*rbgBindings{map[subKey]→nodes}`, enabling O(1) eviction by RBG UID when an RBG is deleted (via `EvictByUID` called from an RBG delete event handler):
    - Pod-level inner key: `{podName}` → single-node set
    - Component-level inner key: `{roleName}-{componentName}` → multi-node set
    - `Load()` returns a **cloned snapshot** (`sets.Clone()`) to avoid data races when callers iterate the set concurrently with `Add()` (safe under `MaxConcurrentReconciles > 1`)
    - `EvictByUID()` and the RBG delete handler emit `klog.InfoS()` logs for observability
    - `Len()` exposed for testing and debugging
  - `RecordNodeBindings()` — observes Running+Ready Pods and records their node bindings (emits `klog.V(4)` when a new key is first recorded)
  - `InjectInPlaceScheduling()` — injects nodeAffinity on Pod creation:
    - Mode check (Preferred / Required)
    - Exclusive topology bypass
    - **Avoid anti-affinity**: `role-inplace-scheduling-avoid` annotation injects `requiredDuringSchedulingIgnoredDuringExecution` with `DoesNotExist` operator to hard-exclude nodes carrying a user-specified label
    - Granularity-aware affinity injection: `preferredDuringSchedulingIgnoredDuringExecution` (weight 100) or `requiredDuringSchedulingIgnoredDuringExecution`
  - `resolveGranularity()` — smart defaults: Stateful → Pod, Stateless → Component (detected via `RoleInstanceIndexLabelKey` label)

- **`instance_scale.go`** — Integration points:
  - `RecordNodeBindings()` called at the beginning of `Scale()` (piggybacks on already-fetched Pods)
  - `InjectInPlaceScheduling()` called in `createPods()` before Pod creation

#### Annotation Propagation

- **Stateful mode** (`stateful_instance_set_utils.go`): propagates `role-inplace-scheduling`, `role-inplace-scheduling-granularity`, and `role-inplace-scheduling-avoid` from RoleInstanceSet to RoleInstance
- **Stateless mode** (`implement.go`): adds the same three annotations to `annotationsToCopy`
- **RBG → RoleInstanceSet** (`roleinstanceset_reconciler.go`): propagates the real RBG UID (`RBGOwnerUIDLabelKey`) as a label when in-place scheduling is enabled, ensuring globally unique binding keys across RBG lifecycles

#### GC — Binding Cleanup on RBG Deletion

- **`api.go`** (`sync` package): exports `EvictNodeBindings(uid)` as the cross-package entry point for cleaning up bindings
- **`rolebasedgroup_controller.go`**: registers a `handler.Funcs` delete handler via `.Watches()` on RBG. When an RBG is deleted, the handler calls `EvictNodeBindings(string(obj.GetUID()))` for O(1) cleanup. No reconcile is enqueued (`.For()` already handles that) — the handler exists purely for the side-effect of freeing in-memory bindings using the UID (which is only available from the delete event, not from a subsequent `Get`)

#### Constants

- **`annotation.go`**: new constants for mode (`Preferred`/`Required`), granularity (`Pod`/`Component`), and avoid annotation
- **`label.go`**: `RBGOwnerUIDLabelKey` for RBG lifecycle isolation

#### Tests

- **`node_binding_test.go`** — 63 unit tests covering:
  - `NodeBindingStore`: Add/Load, idempotency, multi-node accumulation, empty key, **EvictByUID (basic + non-existent + add-after-evict)**, **Len**
  - `splitKey`: valid key, no slash, empty parts
  - `buildKey`: Pod/Component keys, missing fields, unknown granularity
  - `resolveGranularity`: explicit override, Stateful→Pod, Stateless→Component
  - `isPodRunningAndReady`: 5 Pod phase scenarios
  - `RecordNodeBindings`: Stateful/Pod, Stateless/Component, skip non-Ready, skip missing UID
  - `InjectInPlaceScheduling`: Preferred/Required affinity, Component multi-node, no annotation, invalid mode, exclusive topology bypass, no binding, missing label, preserve existing affinity, **Required+avoid merged AND semantics**
  - Avoid anti-affinity: injection, no-binding fallback, no-avoid skip, exclusive skip, no-mode skip
  - Record→Inject integration: Stateful same-name hit, Stateless surge component hit

- **`inplace_scheduling.go`** — 3 E2E tests:
  1. **Dual-role affinity injection after rolling upgrade**:
     - `role-pod` (Stateful, Pod-level granularity, smart default)
     - `role-component` (Stateless, Component-level granularity, smart default)
     - Both use Preferred mode, `maxSurge=0`, `maxUnavailable=1`
     - Rolling upgrade triggered via env var change (forces Pod recreation)
     - Asserts new Pods carry `preferredDuringSchedulingIgnoredDuringExecution` with `weight=100`, `key=kubernetes.io/hostname`, values containing the Pod's actual node
  2. **Avoid anti-affinity + preferred affinity after Pod deletion**:
     - Single role with 1 replica, Preferred mode, `role-inplace-scheduling-avoid` set to `workloads.x-k8s.io/heavy-tenant`
     - Pod deleted manually → controller recreates it
     - Asserts new Pod has both:
       - `preferredDuringSchedulingIgnoredDuringExecution` (weight 100, node binding)
       - `requiredDuringSchedulingIgnoredDuringExecution` with `DoesNotExist` operator (avoid anti-affinity)
  3. **Required mode + avoid: merged AND semantics**:
     - Single role with 1 replica, Required mode + avoid annotation
     - Historical node labelled with avoid key → Pod deleted
     - Asserts merged `NodeSelectorTerm` (hostname In + avoid DoesNotExist in same term = AND)
     - Asserts Pod is either Pending (single-node) or scheduled to a different node (multi-node)
     - Node label cleaned up via `ginkgo.DeferCleanup`

- **`role_wrapper.go`** — added `WithAnnotations()` builder method for test convenience

No. See KEP #351

| Type      | File                                                    | Count | Coverage                                                                                                                                                                                                      |
|-----------|---------------------------------------------------------|-------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Unit Test | `pkg/reconciler/roleinstance/sync/node_binding_test.go` | 63    | NodeBindingStore (incl. EvictByUID, Len), splitKey, buildKey, resolveGranularity, isPodRunningAndReady, RecordNodeBindings, InjectInPlaceScheduling (incl. avoid + Required merge), Record→Inject integration |
| E2E Test  | `test/e2e/testcase/v1alpha2/inplace_scheduling.go`      | 3     | 1) Dual-role affinity after rolling upgrade; 2) Avoid anti-affinity after Pod deletion; 3) Required+avoid merged AND semantics with node label cleanup                                                        |

1. **Unit tests**: `go test ./pkg/reconciler/roleinstance/sync/... -v`
2. **E2E test**: Run the controller locally, then `go test ./test/e2e/ -v -ginkgo.v --ginkgo.focus="inplace-scheduling" -timeout 10m`
3. **Manual verification**:
   - Create an RBG with annotation `rbg.workloads.x-k8s.io/role-inplace-scheduling: "Preferred"` on a role
   - Wait for Pods to become Ready (bindings are recorded)
   - Trigger a rolling upgrade (e.g., change env var)
   - Observe that new Pods carry `preferredDuringSchedulingIgnoredDuringExecution` nodeAffinity with `weight=100` pointing to historical nodes
   - For avoid: set `rbg.workloads.x-k8s.io/role-inplace-scheduling-avoid: "some-label-key"` and verify `requiredDuringSchedulingIgnoredDuringExecution` with `DoesNotExist` is injected

- The `NodeBindingStore` uses a **two-level map** keyed by RBG UID (`map[uid]→rbgBindings{map[subKey]→nodes}`). When an RBG is deleted, a `handler.Funcs` delete handler registered in the RBG reconciler's `SetupWithManager` calls `EvictNodeBindings(uid)` for O(1) cleanup — no TTL, webhooks, finalizers, or background GC goroutines needed. A single `sync.RWMutex` protects both levels; all operations hold the lock for very short durations.
- `Load()` returns a **cloned snapshot** (`sets.Clone()`) rather than a direct reference to the inner set. This prevents data races when callers iterate the returned set (e.g., `sets.List(nodes)`) concurrently with `Add()` modifying the same key — critical under `MaxConcurrentReconciles > 1` with Component-level granularity where multiple RoleInstances share the same store key.
- The avoid anti-affinity uses `requiredDuringSchedulingIgnoredDuringExecution` + `DoesNotExist` (hard constraint), **not** negative weight on preferred terms — Kubernetes scheduler does not support negative weights.
- In **Required** mode, the avoid `DoesNotExist` expression is **merged into the same `NodeSelectorTerm`** as the hostname `In` constraint. Separate terms are ORed by K8s, but we need AND semantics (historical node AND no avoid label).
- Granularity auto-detection relies on the `rbg.workloads.x-k8s.io/role-instance-index` label (present on Stateful RoleInstances, absent on Stateless ones). No annotation propagation needed for detection.
- The `RBGOwnerUIDLabelKey` ensures node bindings are scoped to the current RBG lifecycle. Deleting and recreating the RBG produces a new UID, so stale bindings from a previous RBG are never reused.

## Checklist

### DIFF
--- a/api/workloads/constants/annotation.go
+++ b/api/workloads/constants/annotation.go
@@ -163,3 +163,54 @@ const (
 	// whose failures should not cascade to the main workload.
 	RestartTriggerPolicyIgnore = "Ignore"
 )
+
+// Inplace scheduling annotations and values.
+const (
+	// RoleInplaceSchedulingAnnotationKey enables in-place scheduling for a role.
+	// When set, recreated Pods (due to rolling upgrade or failure recovery) are
+	// injected with nodeAffinity to prefer or require scheduling to nodes where
+	// the same component type has previously run, enabling reuse of node-local
+	// cached resources.
+	// Valid values:
+	//   - "Preferred": inject preferredDuringSchedulingIgnoredDuringExecution with weight 100.
+	//   - "Required": inject requiredDuringSchedulingIgnoredDuringExecution.
+	// When not set, no in-place scheduling affinity is injected.
+	// Example: rbg.workloads.x-k8s.io/role-inplace-scheduling: "Preferred"
+	RoleInplaceSchedulingAnnotationKey = RBGPrefix + "role-inplace-scheduling"
+
+	// InplaceSchedulingPreferred injects preferredDuringSchedulingIgnoredDuringExecution
+	// nodeAffinity with weight 100, steering recreated Pods toward historical nodes
+	// while allowing fallback to any available node.
+	InplaceSchedulingPreferred = "Preferred"
+
+	// InplaceSchedulingRequired injects requiredDuringSchedulingIgnoredDuringExecution
+	// nodeAffinity, requiring recreated Pods to land on a historical node.
+	// If no historical node is available, the Pod remains Pending.
+	InplaceSchedulingRequired = "Required"
+
+	// RoleInplaceSchedulingGranularityAnnotationKey controls the binding granularity
+	// for in-place scheduling. When not set, the default is auto-detected:
+	//   - Stateful mode → Pod
+	//   - Stateless mode → Component
+	// Example: rbg.workloads.x-k8s.io/role-inplace-scheduling-granularity: "Component"
+	RoleInplaceSchedulingGranularityAnnotationKey = RBGPrefix + "role-inplace-scheduling-granularity"
+
+	// InplaceSchedulingGranularityPod uses per-Pod binding: each Pod returns to
+	// its own historical node. Key: {rbgUID}/{podName} → single node.
+	InplaceSchedulingGranularityPod = "Pod"
+
+	// InplaceSchedulingGranularityComponent uses component-level binding: Pod
+	// prefers any node that has hosted the same component type.
+	// Key: {rbgUID}/{roleName}-{componentName} → node set.
+	InplaceSchedulingGranularityComponent = "Component"
+
+	// RoleInplaceSchedulingAvoidAnnotationKey specifies a node label key. When set,
+	// a RequiredDuringSchedulingIgnoredDuringExecution term with DoesNotExist
+	// operator is injected into the Pod's nodeAffinity, hard-excluding nodes
+	// that carry this label.
+	// This is useful to steer Pods away from nodes with certain characteristics
+	// (e.g., nodes running other heavy workloads).
+	// The annotation value is the label key to match against.
+	// Example: rbg.workloads.x-k8s.io/role-inplace-scheduling-avoid: "workloads.x-k8s.io/heavy-tenant"
+	RoleInplaceSchedulingAvoidAnnotationKey = RBGPrefix + "role-inplace-scheduling-avoid"
+)
--- a/api/workloads/constants/annotation.go
+++ b/api/workloads/constants/annotation.go
@@ -175,6 +175,12 @@ const (
 	//   - "Preferred": inject preferredDuringSchedulingIgnoredDuringExecution with weight 100.
 	//   - "Required": inject requiredDuringSchedulingIgnoredDuringExecution.
 	// When not set, no in-place scheduling affinity is injected.
+	//
+	// Note: when adding to an existing RBG, old RoleInstances lack the label so
+	// they are unaffected. New RoleInstances start recording bindings,
+	// but no affinity is injected on the first recreation (empty binding).
+	// Affinity takes effect on subsequent recreations.
+	//
 	// Example: rbg.workloads.x-k8s.io/role-inplace-scheduling: "Preferred"
 	RoleInplaceSchedulingAnnotationKey = RBGPrefix + "role-inplace-scheduling"
 
@@ -185,7 +191,9 @@ const (
 
 	// InplaceSchedulingRequired injects requiredDuringSchedulingIgnoredDuringExecution
 	// nodeAffinity, requiring recreated Pods to land on a historical node.
-	// If no historical node is available, the Pod remains Pending.
+	// If a binding exists but no historical node is healthy, the Pod remains Pending.
+	// On cold start (empty binding store, e.g. after controller restart), no
+	// affinity is injected until RecordNodeBindings reseeds the store.
 	InplaceSchedulingRequired = "Required"
 
 	// RoleInplaceSchedulingGranularityAnnotationKey controls the binding granularity
@@ -204,13 +212,11 @@ const (
 	// Key: {rbgUID}/{roleName}-{componentName} → node set.
 	InplaceSchedulingGranularityComponent = "Component"
 
-	// RoleInplaceSchedulingAvoidAnnotationKey specifies a node label key. When set,
-	// a RequiredDuringSchedulingIgnoredDuringExecution term with DoesNotExist
-	// operator is injected into the Pod's nodeAffinity, hard-excluding nodes
-	// that carry this label.
-	// This is useful to steer Pods away from nodes with certain characteristics
-	// (e.g., nodes running other heavy workloads).
-	// The annotation value is the label key to match against.
-	// Example: rbg.workloads.x-k8s.io/role-inplace-scheduling-avoid: "workloads.x-k8s.io/heavy-tenant"
+	// RoleInplaceSchedulingAvoidAnnotationKey specifies one or more node label
+	// keys. When set, a RequiredDuringSchedulingIgnoredDuringExecution term
+	// with DoesNotExist operator is injected for each key, hard-excluding nodes
+	// that carry any of these labels.
+	// The annotation value is a single label key or a comma-separated list.
+	// Example: rbg.workloads.x-k8s.io/role-inplace-scheduling-avoid: "key1,key2"
 	RoleInplaceSchedulingAvoidAnnotationKey = RBGPrefix + "role-inplace-scheduling-avoid"
 )
--- a/internal/controller/workloads/rolebasedgroup_controller.go
+++ b/internal/controller/workloads/rolebasedgroup_controller.go
@@ -42,6 +42,7 @@ import (
 	coreapplyv1 "k8s.io/client-go/applyconfigurations/core/v1"
 	metaapplyv1 "k8s.io/client-go/applyconfigurations/meta/v1"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -49,6 +50,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -60,6 +62,7 @@ import (
 	"sigs.k8s.io/rbgs/pkg/dependency"
 	"sigs.k8s.io/rbgs/pkg/discovery"
 	"sigs.k8s.io/rbgs/pkg/reconciler"
+	instancesync "sigs.k8s.io/rbgs/pkg/reconciler/roleinstance/sync"
 	"sigs.k8s.io/rbgs/pkg/scale"
 	"sigs.k8s.io/rbgs/pkg/scheduler"
 	"sigs.k8s.io/rbgs/pkg/utils"
@@ -1013,6 +1016,19 @@ func (r *RoleBasedGroupReconciler) SetupWithManager(mgr ctrl.Manager, options co
 		Owns(&workloadsv1alpha2.RoleInstanceSet{}, builder.WithPredicates(WorkloadPredicate())).
 		Owns(&corev1.Service{}).
 		Owns(&workloadsv1alpha2.RoleBasedGroupScalingAdapter{}, builder.MatchEveryOwner, builder.WithPredicates(RBGScalingAdapterPredicate())).
+		Watches(&workloadsv1alpha2.RoleBasedGroup{}, handler.Funcs{
+			// Clean up in-place scheduling bindings when an RBG is deleted.
+			// We don't enqueue a reconcile here — .For() already does that.
+			// This handler exists purely for the side-effect of freeing
+			// in-memory bindings using the UID (which is only available
+			// from the delete event, not from a subsequent Get).
+			DeleteFunc: func(ctx context.Context, e event.DeleteEvent, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+				if e.Object != nil {
+					uid := string(e.Object.GetUID())
+					instancesync.EvictNodeBindings(uid)
+				}
+			},
+		}).
 		Named("workloads-rolebasedgroup")
 
 	err := utils.CheckCrdExists(r.apiReader, utils.LwsCrdName)
--- a/internal/controller/workloads/rolebasedgroup_controller.go
+++ b/internal/controller/workloads/rolebasedgroup_controller.go
@@ -89,9 +89,13 @@ type RoleBasedGroupReconciler struct {
 	workloadReconciler map[string]reconciler.WorkloadReconciler
 	reconcilerMu       sync.RWMutex
 	podGroupManager    scheduler.PodGroupManager
+	// NodeBindings is the in-place scheduling binding store, shared with
+	// the RoleInstance reconciler. Injected at wire-up time so both consumers
+	// operate on the same instance.
+	NodeBindings *instancesync.NodeBindingStore
 }
 
-func NewRoleBasedGroupReconciler(mgr ctrl.Manager, schedulerName scheduler.SchedulerPluginType) (*RoleBasedGroupReconciler, error) {
+func NewRoleBasedGroupReconciler(mgr ctrl.Manager, schedulerName scheduler.SchedulerPluginType, bindings *instancesync.NodeBindingStore) (*RoleBasedGroupReconciler, error) {
 	c := utilclient.NewClientWithUserAgent(mgr, "rolebasedgroup")
 	podGroupManager, err := scheduler.NewPodGroupManager(schedulerName, c)
 	if err != nil {
@@ -104,6 +108,7 @@ func NewRoleBasedGroupReconciler(mgr ctrl.Manager, schedulerName scheduler.Sched
 		recorder:           mgr.GetEventRecorderFor("RoleBasedGroup"),
 		workloadReconciler: make(map[string]reconciler.WorkloadReconciler),
 		podGroupManager:    podGroupManager,
+		NodeBindings:       bindings,
 	}, nil
 }
 
@@ -1027,9 +1032,9 @@ func (r *RoleBasedGroupReconciler) SetupWithManager(mgr ctrl.Manager, options co
 			// in-memory bindings using the UID (which is only available
 			// from the delete event, not from a subsequent Get).
 			DeleteFunc: func(ctx context.Context, e event.DeleteEvent, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
-				if e.Object != nil {
+				if e.Object != nil && r.NodeBindings != nil {
 					uid := string(e.Object.GetUID())
-					instancesync.EvictNodeBindings(uid)
+					r.NodeBindings.EvictByUID(uid)
 				}
 			},
 		}).
--- a/internal/controller/workloads/rolebasedgroup_controller.go
+++ b/internal/controller/workloads/rolebasedgroup_controller.go
@@ -1017,6 +1017,10 @@ func (r *RoleBasedGroupReconciler) SetupWithManager(mgr ctrl.Manager, options co
 		Owns(&corev1.Service{}).
 		Owns(&workloadsv1alpha2.RoleBasedGroupScalingAdapter{}, builder.MatchEveryOwner, builder.WithPredicates(RBGScalingAdapterPredicate())).
 		Watches(&workloadsv1alpha2.RoleBasedGroup{}, handler.Funcs{
+			// CreateFunc, UpdateFunc, GenericFunc are intentionally nil (no-op).
+			// .For() already handles reconcile enqueue for those event types.
+			// This Watches exists solely for the DeleteFunc side-effect below.
+			//
 			// Clean up in-place scheduling bindings when an RBG is deleted.
 			// We don't enqueue a reconcile here — .For() already does that.
 			// This handler exists purely for the side-effect of freeing
--- a/pkg/reconciler/roleinstance/sync/api.go
+++ b/pkg/reconciler/roleinstance/sync/api.go
@@ -44,33 +44,22 @@ type Interface interface {
 // informer cache catching up with the latest status write.
 var restartingCache sync.Map
 
-// nodeBindings is the package-level node binding store shared across all
-// RoleInstance reconcile loops. It uses a two-level map keyed by RBG UID,
-// enabling O(1) eviction of all bindings when an RBG is deleted via
-// EvictByUID (called from the RBG delete event handler).
-var nodeBindings = NewNodeBindingStore()
-
 type realControl struct {
 	client.Client
 	apiReader        client.Reader
 	inplaceControl   inplaceupdate.Interface
 	recorder         record.EventRecorder
 	lifecycleControl lifecycle.Interface
+	bindings         *NodeBindingStore
 }
 
-// EvictNodeBindings removes all in-place scheduling bindings for the
-// given RBG UID. Intended to be called from the RBG delete event handler
-// so that bindings are cleaned up immediately when an RBG is deleted.
-func EvictNodeBindings(uid string) {
-	nodeBindings.EvictByUID(uid)
-}
-
-func New(c client.Client, apiReader client.Reader, recorder record.EventRecorder) Interface {
+func New(c client.Client, apiReader client.Reader, recorder record.EventRecorder, bindings *NodeBindingStore) Interface {
 	return &realControl{
 		Client:           c,
 		apiReader:        apiReader,
 		inplaceControl:   inplaceupdate.New(c),
 		lifecycleControl: lifecycle.New(c),
 		recorder:         recorder,
+		bindings:         bindings,
 	}
 }
--- a/pkg/reconciler/roleinstance/sync/api.go
+++ b/pkg/reconciler/roleinstance/sync/api.go
@@ -45,9 +45,9 @@ type Interface interface {
 var restartingCache sync.Map
 
 // nodeBindings is the package-level node binding store shared across all
-// RoleInstance reconcile loops. It uses a unified map[string]sets.Set[string]
-// structure for both Pod-level and Component-level in-place scheduling
-// granularities; the difference is determined by the key format.
+// RoleInstance reconcile loops. It uses a two-level map keyed by RBG UID,
+// enabling O(1) eviction of all bindings when an RBG is deleted via
+// EvictByUID (called from the RBG delete event handler).
 var nodeBindings = NewNodeBindingStore()
 
 type realControl struct {
@@ -56,6 +56,13 @@ type realControl struct {
 	inplaceControl   inplaceupdate.Interface
 	recorder         record.EventRecorder
 	lifecycleControl lifecycle.Interface
+}
+
+// EvictNodeBindings removes all in-place scheduling bindings for the
+// given RBG UID. Intended to be called from the RBG delete event handler
+// so that bindings are cleaned up immediately when an RBG is deleted.
+func EvictNodeBindings(uid string) {
+	nodeBindings.EvictByUID(uid)
 }
 
 func New(c client.Client, apiReader client.Reader, recorder record.EventRecorder) Interface {
--- a/pkg/reconciler/roleinstance/sync/api.go
+++ b/pkg/reconciler/roleinstance/sync/api.go
@@ -44,6 +44,12 @@ type Interface interface {
 // informer cache catching up with the latest status write.
 var restartingCache sync.Map
 
+// nodeBindings is the package-level node binding store shared across all
+// RoleInstance reconcile loops. It uses a unified map[string]sets.Set[string]
+// structure for both Pod-level and Component-level in-place scheduling
+// granularities; the difference is determined by the key format.
+var nodeBindings = NewNodeBindingStore()
+
 type realControl struct {
 	client.Client
 	apiReader        client.Reader
--- a/pkg/reconciler/roleinstance/sync/node_binding.go
+++ b/pkg/reconciler/roleinstance/sync/node_binding.go
@@ -115,6 +115,9 @@ func (s *NodeBindingStore) Load(key string) sets.Set[string] {
 // EvictByUID removes all bindings for the given RBG UID in O(1).
 // Intended to be called from an RBG delete event handler so that
 // bindings are cleaned up immediately when an RBG is deleted.
+//
+// NOTE: A small race window exists if RecordNodeBindings runs concurrently
+// with eviction. Persisting bindings in a CRD would eliminate this.
 func (s *NodeBindingStore) EvictByUID(uid string) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
@@ -173,6 +176,9 @@ func buildKey(granularity string, instance *workloadsv1alpha2.RoleInstance, pod 
 		return fmt.Sprintf("%s/%s-%s", rbgUID, roleName, componentName)
 
 	default:
+		klog.InfoS("in-place scheduling: unrecognized granularity, skipping injection",
+			"instance", klog.KObj(instance), "granularity", granularity,
+			"validValues", fmt.Sprintf("%q or %q", constants.InplaceSchedulingGranularityPod, constants.InplaceSchedulingGranularityComponent))
 		return ""
 	}
 }
@@ -220,6 +226,9 @@ func RecordNodeBindings(store *NodeBindingStore, instance *workloadsv1alpha2.Rol
 
 		// Log when a key is first recorded — helps operators debug scheduling
 		// decisions without noise at default log levels.
+		// NOTE: TOCTOU race between Load and Add — concurrent reconciles on the
+		// same key may both log "new binding". Benign: Add is idempotent, and the
+		// log is informational only.
 		if store.Load(key) == nil {
 			klog.V(4).InfoS("in-place scheduling: recording new node binding",
 				"key", key, "node", pod.Spec.NodeName,
@@ -281,16 +290,24 @@ func InjectInPlaceScheduling(pod *v1.Pod, instance *workloadsv1alpha2.RoleInstan
 		return
 	}
 
-	// 3. Build avoid expression if configured (used in step 6).
-	// In Required mode, this expression is merged into the same NodeSelectorTerm
-	// as the in-place hostname constraint (AND semantics within a single term).
-	// In Preferred mode (or when no binding exists), it is injected as a
-	// standalone required term — the scheduler ANDs required and preferred types.
-	var avoidExpr *v1.NodeSelectorRequirement
-	if avoidLabelKey := instance.Annotations[constants.RoleInplaceSchedulingAvoidAnnotationKey]; avoidLabelKey != "" {
-		avoidExpr = &v1.NodeSelectorRequirement{
-			Key:      avoidLabelKey,
-			Operator: v1.NodeSelectorOpDoesNotExist,
+	// 3. Build avoid expressions if configured (used in steps 6 and 7).
+	// The avoid annotation value may be a single label key or a comma-separated
+	// list (e.g. "key1,key2"). Each key generates a DoesNotExist expression.
+	// In Required mode, these are merged into the same NodeSelectorTerm as the
+	// in-place hostname constraint (AND semantics within a single term).
+	// In Preferred mode (or when no binding exists), they are folded into every
+	// existing required term to preserve AND semantics across all constraints.
+	var avoidExprs []v1.NodeSelectorRequirement
+	if raw := instance.Annotations[constants.RoleInplaceSchedulingAvoidAnnotationKey]; raw != "" {
+		for _, key := range strings.Split(raw, ",") {
+			key = strings.TrimSpace(key)
+			if key == "" {
+				continue
+			}
+			avoidExprs = append(avoidExprs, v1.NodeSelectorRequirement{
+				Key:      key,
+				Operator: v1.NodeSelectorOpDoesNotExist,
+			})
 		}
 	}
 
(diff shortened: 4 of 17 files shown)


cheyang:
## In-Place Scheduling Review

Solid feature implementation with clean separation of concerns (store, record, inject), comprehensive unit tests (~890 lines), and good e2e coverage. The design gracefully handles cold-start (no bindings after controller restart) and uses RBG UID to prevent stale binding inheritance.

**Blocking:** The avoid anti-affinity term and in-place Required term are appended as separate `NodeSelectorTerms` which Kubernetes ORs. When both are active with Required mode, this silently weakens the hard-exclude guarantee. Needs a fix or documented restriction.

**Non-blocking:** In-memory store lacks eviction for deleted RBGs (acceptable given restart self-correction, but worth a follow-up GC mechanism for large-scale deployments).

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -0,0 +1,288 @@
...
+			v1.NodeSelectorTerm{
+				MatchExpressions: []v1.NodeSelectorRequirement{
+					{
+						Key:      avoidLabelKey,
+						Operator: v1.NodeSelectorOpDoesNotExist,
+					},
+				},
+			},
+		)
+	}
```
cheyang, line 320: When both `RoleInplaceSchedulingAvoidAnnotationKey` and `RoleInplaceSchedulingAnnotationKey=Required` are active, the avoid term and the in-place term are appended as separate `NodeSelectorTerms`. In Kubernetes, multiple NodeSelectorTerms are ORed, so a pod can schedule on nodes satisfying EITHER (no-avoid-label) OR (historical-node). The intent is AND: only land on a historical node that also lacks the avoid label.

Fix: merge the avoid MatchExpression into the same NodeSelectorTerm as the in-place Required constraint, or document that avoid is only supported with Preferred mode and add a guard.

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -0,0 +1,288 @@
...
+// gets a different UID and does not inherit stale bindings.
+//
+// The store is naturally bounded and self-correcting after controller restart.
+type NodeBindingStore struct {
+	mu       sync.RWMutex
+	bindings map[string]sets.Set[string]
+}
+
+// NewNodeBindingStore creates a new empty NodeBindingStore.
+func NewNodeBindingStore() *NodeBindingStore {
```
cheyang, line 63: The package-level `nodeBindings` store accumulates entries keyed by RBG UID. When an RBG is deleted, its entries are never evicted. For long-running controllers managing many RBGs with frequent delete/recreate cycles, this leaks memory until the next controller restart. Acceptable for initial merge given the self-correction on restart, but worth a follow-up GC mechanism for large-scale deployments.

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -0,0 +1,288 @@
...
+//
+// In Preferred mode, a preferred scheduling term (weight 100) is injected.
+// In Required mode, a hard constraint is injected.
+//
+// The injection is skipped when:
+//   - The in-place scheduling annotation is not set on the instance.
+//   - The annotation mode is not "Preferred" or "Required" (with a warning).
+//   - The pod has exclusive topology affinity (to avoid conflicts).
+//   - No node binding exists for this Pod (initial creation or controller just restarted).
+//   - The granularity is Component but the Pod has no component-name label.
```
cheyang, line 275: The e2e tests only exercise avoid with Preferred mode. If avoid + Required is not intended to be used together, adding a validation guard (or at least a log warning) would prevent users from hitting the OR-semantics issue silently. Could you clarify the intended scope of the avoid annotation?

pkg/reconciler/roleinstance/sync/node_binding_test.go
```diff
@@ -0,0 +1,891 @@
...
+	}
+
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{Name: "prefill-0"},
+	}
+
+	InjectInPlaceScheduling(pod, instance, store)
+	// No mode annotation → entire injection skipped (including avoid)
+	assert.Nil(t, pod.Spec.Affinity)
+}
```
cheyang, line 1026: Unit tests cover avoid with Preferred mode but not the Required mode combination, which would expose the OR-semantics issue. Adding a test that asserts correct AND behavior would lock down the expected semantics and prevent regressions once the bug is fixed.

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -0,0 +1,288 @@
...
+		if pod.Spec.NodeName == "" {
+			continue
+		}
+		if !isPodRunningAndReady(pod) {
+			continue
+		}
+
+		key := buildKey(granularity, instance, pod)
+		if key == "" {
+			continue
```
cheyang, line 225: RecordNodeBindings silently populates the store. A `klog.V(4).InfoS` line when a new key is first recorded would help operators debug scheduling decisions without adding noise at default log levels.

pkg/reconciler/roleinstance/sync/node_binding.go
NoobDream2568: Good catch! I will correct this soon. Merge the avoid DoesNotExist expression into the same NodeSelectorTerm as the hostname In constraint when mode is Required.

pkg/reconciler/roleinstance/sync/node_binding_test.go
NoobDream2568: UT added.

pkg/reconciler/roleinstance/sync/node_binding.go
NoobDream2568: E2e test added.

pkg/reconciler/roleinstance/sync/api.go
```diff
@@ -44,6 +44,12 @@ type Interface interface {
 // informer cache catching up with the latest status write.
 var restartingCache sync.Map
 
+// nodeBindings is the package-level node binding store shared across all
+// RoleInstance reconcile loops. It uses a unified map[string]sets.Set[string]
+// structure for both Pod-level and Component-level in-place scheduling
+// granularities; the difference is determined by the key format.
```
cheyang: Consider replacing the custom `NodeBindingStore` with `k8s.io/apimachinery/pkg/util/cache.Expiring` — it is already vendored in this repo and gives you per-key TTL with automatic GC (runs inline during `Set`, no extra goroutine needed).

Rough sketch:

```go
import utilcache "k8s.io/apimachinery/pkg/util/cache"

var nodeBindings = utilcache.NewExpiring()

// Record: each Set() refreshes the TTL; deleted-RBG keys expire naturally.
func RecordNodeBindings(...) {
    // ...
    nodeBindings.Set(key, nodes.Clone(), 30*time.Minute)
}

// Inject: Get() returns (nil, false) for expired keys.
func InjectInPlaceScheduling(...) {
    val, ok := nodeBindings.Get(key)
    if !ok { return }
    nodes := val.(sets.Set[string])
    // ...
}
```

This also fixes the data-race in the current `Load` — right now it returns a direct reference to the underlying `sets.Set`, which callers iterate outside the lock (`sets.List(nodes)` at line 260 of `node_binding.go`). A concurrent `Add` on the same key modifies the same map → `fatal error: concurrent map read and map write`. Storing a `.Clone()` on `Set` and returning an independent copy on `Get` eliminates the race.

Two birds with one stone: TTL-based GC for deleted RBGs + concurrency safety.

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -0,0 +1,317 @@
...
+
+// Load returns the set of node names for key. Returns nil if no binding exists.
+func (s *NodeBindingStore) Load(key string) sets.Set[string] {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return s.bindings[key]
+}
+
+// buildKey constructs the binding store key based on the given granularity.
+//
```
cheyang, line 150: **Data race**: `Load` returns a direct reference to the underlying `sets.Set[string]`. Callers iterate this set outside the lock (e.g., `sets.List(nodes)` at line 260). With `MaxConcurrentReconciles > 1`, a concurrent `Add` on the same key modifies the same map, triggering `fatal error: concurrent map read and map write`. Most likely with Component-level granularity where multiple RoleInstances share the same key.

**Fix**: clone under the read lock:

```go
func (s *NodeBindingStore) Load(key string) sets.Set[string] {
    s.mu.RLock()
    defer s.mu.RUnlock()
    v := s.bindings[key]
    if v == nil {
        return nil
    }
    return v.Union(nil) // clone
}
```

pkg/reconciler/roleinstance/sync/api.go
cheyang: **GC for deleted RBGs**: This store accumulates entries keyed by RBG UID and never evicts them. Entries persist until controller restart.

Instead of TTL (my earlier suggestion — please disregard), a cleaner approach: add a `DeleteByPrefix` method and wire up a controller-runtime delete event handler on RBG.

**(1) Add prefix deletion to the store:**

```go
func (s *NodeBindingStore) DeleteByPrefix(prefix string) {
    s.mu.Lock()
    defer s.mu.Unlock()
    for k := range s.bindings {
        if strings.HasPrefix(k, prefix+"/") {
            delete(s.bindings, k)
        }
    }
}
```

**(2) Wire up a delete event handler on the RoleInstance controller:**

```go
ctrl.NewControllerManagedBy(mgr).
    For(&workloadsv1alpha2.RoleInstance{}).
    Watches(&workloadsv1alpha2.RoleBasedGroup{},
        handler.Funcs{
            DeleteFunc: func(ctx context.Context, e event.DeleteEvent, q workqueue.RateLimitingInterface) {
                sync.CleanupNodeBindings(string(e.Object.GetUID()))
            },
        },
    ).
    Complete(r)
```

The delete event fires before the informer cache evicts the object, so the UID is available. If the controller is down when an RBG is deleted, the missed event is harmless — controller restart clears the entire in-memory store, which is the same self-correction behavior documented in this PR.

No TTL, no external cache dependency, no finalizer.

NoobDream2568: I've refactored the NodeBinding struct into a nested hierarchy: the first level is keyed by rbg-uid, and the second level is a map that manages individual component and pod entries. This enables O(1) cleanup in the RBG reconciler via the informer's delete event handler.

pkg/reconciler/roleinstance/sync/node_binding.go
NoobDream2568: Fixed.

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -0,0 +1,388 @@
...
+	granularity := resolveGranularity(instance)
+	key := buildKey(granularity, instance, pod)
+	if key == "" {
+		return
+	}
+
+	// 5. Load binding
+	nodes := store.Load(key)
+	if len(nodes) == 0 {
+		// No binding — if avoid is configured, inject it as a standalone
```
cheyang: Required + avoid was correctly merged into one NodeSelectorTerm, but this no-binding fallback (and the Preferred branch below) still appends the avoid expression as a separate NodeSelectorTerm. NodeSelectorTerms are ORed by Kubernetes, so if the user already declared their own RequiredDuringSchedulingIgnoredDuringExecution terms on the Pod template (e.g. a zone constraint), a node that satisfies the user's term and also carries the avoid label will still be admitted, which weakens the avoid. Either fold the avoid expression into each pre-existing required term, or document that this feature assumes no pre-existing required nodeAffinity. A unit test that pre-populates RequiredDuringSchedulingIgnoredDuringExecution would lock the chosen behavior down.

api/workloads/constants/annotation.go
```diff
@@ -163,3 +163,54 @@ const (
...
+	RoleInplaceSchedulingAnnotationKey = RBGPrefix + "role-inplace-scheduling"
+
+	// InplaceSchedulingPreferred injects preferredDuringSchedulingIgnoredDuringExecution
+	// nodeAffinity with weight 100, steering recreated Pods toward historical nodes
+	// while allowing fallback to any available node.
+	InplaceSchedulingPreferred = "Preferred"
+
+	// InplaceSchedulingRequired injects requiredDuringSchedulingIgnoredDuringExecution
+	// nodeAffinity, requiring recreated Pods to land on a historical node.
+	// If no historical node is available, the Pod remains Pending.
```
cheyang: This says "If no historical node is available, the Pod remains Pending", but the cold-start path in InjectInPlaceScheduling treats an empty binding as a no-op and skips Required entirely, so after a controller restart Required degrades to "no affinity at all" until RecordNodeBindings reseeds. That's the right availability call, but the comment will mislead operators. Could you narrow it to the case where a binding exists but no historical node is healthy, and add one line about the cold-start behavior?

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -0,0 +1,388 @@
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	rb := s.bindings[uid]
+	if rb == nil {
+		rb = &rbgBindings{keys: make(map[string]sets.Set[string])}
+		s.bindings[uid] = rb
+	}
+	if rb.keys[subKey] == nil {
+		rb.keys[subKey] = sets.New[string]()
```
cheyang, line 87: Component-level sets only grow over the lifetime of an RBG. There's no per-node TTL or scrub against the live node list, so drained, deleted, or relabelled nodes stay in the affinity values forever. Preferred mode shrugs this off because the scheduler ignores absent nodes, but Required + Component can keep a freshly-created Pod Pending against ghost nodes after enough cluster churn. A known-limitation note here, or a follow-up that prunes against the current node list during reconcile, would be worth having.

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -0,0 +1,388 @@
...
+	if v == nil {
+		return nil
+	}
+	return v.Clone()
+}
+
+// EvictByUID removes all bindings for the given RBG UID in O(1).
+// Intended to be called from an RBG delete event handler so that
+// bindings are cleaned up immediately when an RBG is deleted.
+func (s *NodeBindingStore) EvictByUID(uid string) {
```
cheyang, line 122: Small race window between EvictByUID and an in-flight Scale() reconcile for one of the deleted RBG's RoleInstances: if the delete event fires while RecordNodeBindings is still running, entries can be re-added under the now-defunct UID right after eviction and will never be cleaned up (no future delete event will carry that UID). Bounded by concurrent reconciles per delete and self-heals on controller restart, so not blocking, but a comment acknowledging the window — or a quick existence check on the RBG before recording — would help future readers.

pkg/reconciler/roleinstanceset_reconciler.go
```diff
@@ -144,6 +144,14 @@ func (r *RoleInstanceSetReconciler) constructRoleInstanceSetApplyConfiguration(
 	roleInstanceSetLabel := maps.Clone(matchLabels)
 	roleInstanceSetLabel[fmt.Sprintf(constants.RoleRevisionLabelKeyFmt, role.Name)] = revisionKey
 
+	// When in-place scheduling is enabled for this role, propagate the real
+	// RBG UID (Kubernetes object UID) as a label. This enables the node
+	// binding store to use a globally unique key that changes on RBG
+	// delete/recreate, preventing stale binding inheritance.
+	if role.Annotations[constants.RoleInplaceSchedulingAnnotationKey] != "" {
```
cheyang, line 151: RBGOwnerUIDLabelKey is propagated only when the annotation is set right now. If a user enables the feature later, the label only appears on RoleInstanceSets / RoleInstances that get rebuilt afterwards; existing instances stay un-labelled, so buildKey returns "" for them and they neither record nor receive in-place affinity until they are recreated. That is consistent with the feature being opt-in at recreation time, but it would be worth one sentence in the annotation doc so users don't expect it to take effect on already-running pods.

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -0,0 +1,388 @@
...
+
+	case constants.InplaceSchedulingGranularityComponent:
+		roleName := instance.Labels[constants.RoleNameLabelKey]
+		componentName := pod.Labels[constants.ComponentNameLabelKey]
+		if roleName == "" || componentName == "" {
+			return ""
+		}
+		return fmt.Sprintf("%s/%s-%s", rbgUID, roleName, componentName)
+
+	default:
```
cheyang, line 179: Unknown granularity values land here, buildKey returns "", and the entire injection is silently skipped. The mode validation above logs a klog.InfoS for unrecognized values; a parallel log line here would surface user typos instead of leaving them silently inert.

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -0,0 +1,388 @@
...
+	if mode != constants.InplaceSchedulingPreferred && mode != constants.InplaceSchedulingRequired {
+		klog.InfoS("in-place scheduling: unrecognized mode, skipping injection",
+			"instance", klog.KObj(instance), "mode", mode,
+			"validValues", fmt.Sprintf("%q or %q", constants.InplaceSchedulingPreferred, constants.InplaceSchedulingRequired))
+		return
+	}
+
+	// 2. Check exclusive topology conflict
+	if pod.Annotations[constants.GroupExclusiveTopologyKey] != "" {
+		return
```
vicoooo26, line 293: Maybe we could add some debug log here, just like the way we handle `InplaceScheduling` mode.

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -0,0 +1,388 @@
...
+	}
+
+	// 5. Load binding
+	nodes := store.Load(key)
+	if len(nodes) == 0 {
+		// No binding — if avoid is configured, inject it as a standalone
+		// hard constraint so the pod still avoids labelled nodes.
+		if avoidExpr != nil {
+			ensureNodeAffinity(pod)
```
vicoooo26: `ensureNodeAffinity(pod)` is duplicated, maybe we could move it before phase 5 without changing the logic.

cheyang:
Round 2 review. The previously-blocking OR-semantics bug in Required mode + Avoid is fixed: avoid and hostname expressions are now merged into a single NodeSelectorTerm (AND within a term). The NodeBindingStore refactor to a two-level map is clean and enables O(1) eviction on RBG delete. Unit and E2E coverage for the fix is thorough. No blocking issues remain — LGTM.

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -0,0 +1,388 @@
...
+		if pod.Spec.NodeName == "" {
+			continue
+		}
+		if !isPodRunningAndReady(pod) {
+			continue
+		}
+
+		key := buildKey(granularity, instance, pod)
+		if key == "" {
+			continue
```
cheyang, line 225: Load (RLock, clone, release) then Add (Lock) leaves a gap where the V(4) 'recording new node binding' log may fire more than once for the same key under high concurrency. Harmless since it's only a debug log, but worth noting.

cheyang, line 225: The Load-then-Add pattern for logging could be replaced with an AddIfAbsent that returns a boolean, eliminating the TOCTOU race on the log. Very minor — only affects debug log accuracy under concurrent recording.

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -0,0 +1,388 @@
...
+		return
+	}
+
+	// 3. Build avoid expression if configured (used in step 6).
+	// In Required mode, this expression is merged into the same NodeSelectorTerm
+	// as the in-place hostname constraint (AND semantics within a single term).
+	// In Preferred mode (or when no binding exists), it is injected as a
+	// standalone required term — the scheduler ANDs required and preferred types.
+	var avoidExpr *v1.NodeSelectorRequirement
+	if avoidLabelKey := instance.Annotations[constants.RoleInplaceSchedulingAvoidAnnotationKey]; avoidLabelKey != "" {
```
Syspretor: Should the value for constants.RoleInplaceSchedulingAvoidAnnotationKey be a comma-separated string list to support multiple avoid keys?

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -0,0 +1,388 @@
...
+		avoidExpr = &v1.NodeSelectorRequirement{
```
vicoooo26: Should we add some validation for `avoidLabelKey` to prevent invalid key names and reconcile errors?
eg: 
`"message":"Reconciler error","controller":"roleinstance","controllerGroup":"workloads.x-k8s.io","controllerKind":"RoleInstance","RoleInstance":{"name":"vllm-agg-inference-sim-v1a2-inference-0","namespace":"default"},"namespace":"default","name":"vllm-agg-inference-sim-v1a2-inference-0","reconcileID":"fb760f44-f72d-4e9d-806b-f00755bb8f9d","error":"Pod \"vllm-agg-inference-sim-v1a2-inference-0\" is invalid: spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[0].matchExpressions[0].key: Invalid value: \"not a valid key!\": name part must consist of alphanumeric characters, '-', '_' or '.'`

NoobDream2568: Maybe put it into validating webhook. Can do this later.

pkg/reconciler/roleinstance/sync/node_binding.go
NoobDream2568: Docs already clarify that Inplace Scheduling and GroupExclusiveTopology are mutually exclusive. IMO, silent handling is acceptable here; no additional logging needed.

pkg/reconciler/roleinstance/sync/node_binding.go
NoobDream2568: Agreed. Folding the avoid expression into each pre-existing required term is the correct approach here.

pkg/reconciler/roleinstance/sync/node_binding.go
NoobDream2568: Persisting these records in a CRD would inherently solve this issue. In the current approach, checking whether the RBG still exists seems unnecessary.

pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_utils.go
```diff
@@ -215,6 +215,20 @@ func newVersionedInstance(
...
+	if v, ok := currentSet.Annotations[constants.RoleInplaceSchedulingAnnotationKey]; ok {
+		instance.Annotations[constants.RoleInplaceSchedulingAnnotationKey] = v
+	}
+	// Propagate in-place scheduling granularity annotation (Pod vs Component).
+	if v, ok := currentSet.Annotations[constants.RoleInplaceSchedulingGranularityAnnotationKey]; ok {
+		instance.Annotations[constants.RoleInplaceSchedulingGranularityAnnotationKey] = v
+	}
+	// Propagate in-place scheduling avoid annotation (hard-exclude nodes carrying a specific label).
+	if v, ok := currentSet.Annotations[constants.RoleInplaceSchedulingAvoidAnnotationKey]; ok {
+		instance.Annotations[constants.RoleInplaceSchedulingAvoidAnnotationKey] = v
```
Syspretor, line 229: **[High]** `RBGOwnerUIDLabelKey` 作为 **label** 设置在 RoleInstanceSet 上（见 `roleinstanceset_reconciler.go:152`），但这里只传播了 **annotations** 到 RoleInstance，没有看到对应的 label 传播。

`buildKey()` 从 `instance.Labels[constants.RBGOwnerUIDLabelKey]` 读取该值。如果这个 label 没有从 RoleInstanceSet 传播到 RoleInstance，那么 `buildKey()` 会返回空字符串，整个 in-place scheduling 功能将**静默失效**（不记录 bindings，不注入 affinity）。

**建议：** 确认 RoleInstanceSet → RoleInstance 的创建路径是否会自动传播 labels。如果不会，需要在这里显式添加 label 传播，例如：
```go
if v, ok := currentSet.Labels[constants.RBGOwnerUIDLabelKey]; ok {
    if instance.Labels == nil {
        instance.Labels = make(map[string]string)
    }
    instance.Labels[constants.RBGOwnerUIDLabelKey] = v
}
```

同样需要检查 stateless mode 的传播路径（`implement.go` 的 `GenInstanceFromTemplate`）是否也需要补充。

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -0,0 +1,388 @@
...
+		}
+
+		key := buildKey(granularity, instance, pod)
+		if key == "" {
+			continue
+		}
+
+		// Log when a key is first recorded — helps operators debug scheduling
+		// decisions without noise at default log levels.
+		if store.Load(key) == nil {
```
Syspretor, line 233: **[Medium]** `store.Load()` 和 `store.Add()` 之间存在 TOCTOU race：

```
store.Load(key)  →  释放 RLock  →  另一个 goroutine 的 store.Add()  →  store.Add(key, ...)
```

在 `MaxConcurrentReconciles > 1` 时，同一个 key 可能被多个 RoleInstance 的 reconcile 同时处理。这导致 `klog.V(4)` 的 "recording new node binding" 日志可能对同一个 key 输出多次，失去 "首次记录" 的语义。

功能上是无害的（`Add` 是幂等的），但建议二选一：
1. 添加 `AddAndReport(key, node) (isNew bool)` 方法，在持锁期间判断并返回是否为新 key
2. 接受这个 benign race，但在这里加一行注释说明，避免后续维护者困惑

internal/controller/workloads/rolebasedgroup_controller.go
```diff
@@ -1013,6 +1016,19 @@ func (r *RoleBasedGroupReconciler) SetupWithManager(mgr ctrl.Manager, options co
 		Owns(&workloadsv1alpha2.RoleInstanceSet{}, builder.WithPredicates(WorkloadPredicate())).
 		Owns(&corev1.Service{}).
 		Owns(&workloadsv1alpha2.RoleBasedGroupScalingAdapter{}, builder.MatchEveryOwner, builder.WithPredicates(RBGScalingAdapterPredicate())).
+		Watches(&workloadsv1alpha2.RoleBasedGroup{}, handler.Funcs{
```
Syspretor, line 1024: **[Medium]** `handler.Funcs` 只设置了 `DeleteFunc`，`CreateFunc`、`UpdateFunc`、`GenericFunc` 默认为 no-op。

虽然这是有意为之（`.For()` 已处理 reconcile 入队），但未来维护者可能会：
1. 在这里添加其他 Watch 逻辑，却困惑于为什么 Create/Update 事件没有触发
2. 误以为这个 handler 是多余的而删除

建议补充注释明确说明其他 handler 是 intentional no-op：
```go
Watches(&workloadsv1alpha2.RoleBasedGroup{}, handler.Funcs{
    // CreateFunc, UpdateFunc, GenericFunc are intentionally nil (no-op).
    // .For() already handles reconcile enqueue for those event types.
    // This Watches exists solely for the DeleteFunc side-effect below.
    DeleteFunc: func(...) { ... },
})
```

test/e2e/testcase/v1alpha2/inplace_scheduling.go
```diff
@@ -0,0 +1,476 @@
...
+	gomega.Expect(exprs[1].Operator).Should(gomega.Equal(corev1.NodeSelectorOpDoesNotExist))
+}
+
+// assertPodNotScheduledOnOriginalNode waits for a stable period then verifies
+// that the pod is either Pending or Running on a node different from
+// originalNode.
+func assertPodNotScheduledOnOriginalNode(f *framework.Framework, pod corev1.Pod, originalNode string) {
+	ginkgo.GinkgoHelper()
+
+	gomega.Consistently(func() bool {
```
Syspretor, line 386: **[Medium]** 这个 E2E 测试在不同拓扑的集群上行为不一致：

- **单节点 Kind 集群**：新 Pod 只能 stay Pending（唯一节点被 avoid label 标记）
- **多节点集群**：新 Pod 可能成功调度到其他节点，测试通过但原因与预期不同

`Consistently("12s", "2s")` 的断言在两种情况下都能 pass，但语义不同。这可能导致：
1. CI 环境中因集群拓扑变化产生 flaky test
2. 测试通过但实际上 Required+avoid 的 AND 语义没有被正确验证

建议：
- 在测试开头记录集群节点数量，根据节点数选择不同的断言策略
- 或者使用 `ginkgo.Label` 标记此测试为 `multi-node`，在单节点 CI 中跳过
- 至少在注释中说明预期的集群拓扑要求

Syspretor (commented):
## Code Review Summary

Overall the implementation is solid — well-designed concurrency model, correct K8s NodeSelectorTerm semantics, and comprehensive test coverage. Below are a few issues to address.

- **1 High severity** issue around `RBGOwnerUIDLabelKey` label propagation
- **3 Medium severity** issues around race condition, handler clarity, and E2E flakiness

See inline comments for details.

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -0,0 +1,388 @@
...
+//	                                         hosted the same component type
+//
+// Using the RBG's real Kubernetes object UID (RBGOwnerUIDLabelKey) ensures
+// that when an RBG is deleted and recreated with the same name, the new RBG
+// gets a different UID and does not inherit stale bindings.
+//
+// Concurrency: a single sync.RWMutex protects both map levels. All operations
+// (Add, Load, EvictByUID) hold the lock for very short durations (map lookups
+// and set inserts), so contention is negligible in practice.
+type NodeBindingStore struct {
```
vicoooo26, line 50: We could abstract a store interface and use the current in-memory implementation by default so that we can support future implementations, such as file, configmap, cr, etc., with minimal changes.

api/workloads/constants/annotation.go
NoobDream2568: Done.

pkg/reconciler/roleinstance/sync/node_binding.go
NoobDream2568: Okay I've added a log. Going forward, it might be worth considering validating the granularity value in the admission webhook as a more fundamental fix.

pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_utils.go
NoobDream2568: Labels are already propagated in the existing code, as shown below:

Stateful mode — stateful_instance_set_utils.go L180-183：
```go
// Copy all labels from InstanceSet to Instance (consistent with statelessmode)
for k, v := range currentSet.Labels {
    instance.Labels[k] = v
}
```

Stateless mode — implement.go L235-241：
```go
func genInstanceLabelSet(set *workloadsv1alpha2.RoleInstanceSet) labels.Set {
    desiredLabels := make(labels.Set)
    for k, v := range set.Labels {
        desiredLabels[k] = v
    }
    return desiredLabels
}
```

pkg/reconciler/roleinstance/sync/node_binding.go
NoobDream2568: Good suggestion.

internal/controller/workloads/rolebasedgroup_controller.go
NoobDream2568: Added.

test/e2e/testcase/v1alpha2/inplace_scheduling.go
NoobDream2568: We assert that the pod never lands on the same node, regardless of the number of actual nodes in the testing cluster.

pkg/reconciler/roleinstanceset_reconciler.go
NoobDream2568: Added following comment
```go
	// Note: when adding to an existing RBG, old RoleInstances lack the label so
	// they are unaffected. New RoleInstances start recording bindings,
	// but no affinity is injected on the first recreation (empty binding).
	// Affinity takes effect on subsequent recreations.
```

pkg/reconciler/roleinstance/sync/node_binding.go
NoobDream2568: Implemented in this PR. The value now accepts a comma-separated list, which aligns with the OpenKruise.

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -0,0 +1,415 @@
...
+	// The avoid annotation value may be a single label key or a comma-separated
+	// list (e.g. "key1,key2"). Each key generates a DoesNotExist expression.
+	// In Required mode, these are merged into the same NodeSelectorTerm as the
+	// in-place hostname constraint (AND semantics within a single term).
+	// In Preferred mode (or when no binding exists), they are folded into every
+	// existing required term to preserve AND semantics across all constraints.
+	var avoidExprs []v1.NodeSelectorRequirement
+	if raw := instance.Annotations[constants.RoleInplaceSchedulingAvoidAnnotationKey]; raw != "" {
+		for _, key := range strings.Split(raw, ",") {
+			key = strings.TrimSpace(key)
```
cheyang, line 306: After TrimSpace, this still takes the raw annotation value as a label key without validating it. If a user supplies an invalid key (e.g. spaces inside, illegal characters, or a malformed qualified name), the Pod create call will be rejected by the API server on every retry, putting the controller into an infinite reconcile error loop (vicoooo26 provided real error logs for this in the prior round). A one-liner would close it:

```go
if errs := validation.IsQualifiedName(key); len(errs) > 0 {
    klog.InfoS("in-place scheduling: invalid avoid label key, skipping", "key", key, "errors", errs)
    continue
}
```

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -0,0 +1,415 @@
...
+	}
+	if mode != constants.InplaceSchedulingPreferred && mode != constants.InplaceSchedulingRequired {
+		klog.InfoS("in-place scheduling: unrecognized mode, skipping injection",
+			"instance", klog.KObj(instance), "mode", mode,
+			"validValues", fmt.Sprintf("%q or %q", constants.InplaceSchedulingPreferred, constants.InplaceSchedulingRequired))
+		return
+	}
+
+	// 2. Check exclusive topology conflict
+	if pod.Annotations[constants.GroupExclusiveTopologyKey] != "" {
```
cheyang, line 290: The exclusive-topology bypass returns silently here. Every other early-return in this function now has a log line (unrecognized mode at :282, unrecognized granularity at :176). A symmetric klog.V(4) would surface cases where in-place scheduling was intentionally skipped due to topology exclusivity, rather than leaving operators wondering why no affinity appeared.

pkg/reconciler/roleinstance/sync/node_binding.go
cheyang, line 306: After TrimSpace, the key is still used without validation. If a user supplies an invalid label key (e.g. `-invalid`, contains spaces, or exceeds length limits), the Pod create call is rejected by the API server on every retry, putting the controller into an infinite reconcile error loop.

Suggested fix — add `validation.IsQualifiedName` (already vendored via `k8s.io/apimachinery`) right after the empty-string check:

```go
import "k8s.io/apimachinery/pkg/util/validation"

// inside the for loop, after TrimSpace + empty check:
if errs := validation.IsQualifiedName(key); len(errs) > 0 {
    klog.InfoS("in-place scheduling: invalid avoid label key, skipping",
        "instance", klog.KObj(instance), "key", key, "errors", errs)
    continue
}
```

This mirrors the API server's own validation for `NodeSelectorRequirement.Key`, so we reject locally what would be rejected remotely — 3 lines to close the reconcile-loop vector.

cheyang (approved):
/lgtm
/approve

PR #375, opened 2026-06-29
ci: add /release-test workflow and run e2e-manifest on every PR

Syspretor:
## Summary
- Add `/release-test` slash command workflow that runs full e2e tests (Helm + manifest) on PRs when triggered by a comment
- Run e2e-manifest test on every PR (not just when `deploy/` changes), building controller image from PR source

## Test plan
- [ ] Verify `/release-test` comment triggers the release-test workflow
- [ ] Verify e2e-manifest runs on PRs without deploy/ changes
- [ ] Check that the workflow produces correct artifacts and logs

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.github/workflows/release-test.yml
+++ b/.github/workflows/release-test.yml
@@ -1,10 +1,8 @@
-name: E2E Test
+name: Release Test
 
 on:
-  push:
-    branches: [ "main" ]
-  pull_request:
-    branches: [ "main" ]
+  issue_comment:
+    types: [created]
 
 env:
   GO_VERSION: 1.24.1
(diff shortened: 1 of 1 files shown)


.github/workflows/release-test.yml
```diff
@@ -0,0 +1,328 @@
...
+  SCHEDULER_PLUGINS_VERSION: v0.32.7
+
+# Declare default permissions as read only.
+permissions: read-all
+
+jobs:
+  check-command:
+    if: |
+      github.event.issue.pull_request &&
+      contains(github.event.comment.body, '/release-test')
```
cheyang, line 22: The `check-command` job only filters on `contains(github.event.comment.body, '/release-test')`. There is no restriction on `github.event.comment.author_association`, so any GitHub user (including drive-by external commenters with no permissions on this repo) can fire a full Kind + Volcano + scheduler-plugins + manifest deploy + `make test-e2e` job on any open PR. That is both a real cost vector (two heavy E2E jobs spun up per comment, with no concurrency cap) and a small abuse vector (forces runners to fetch and execute the PR head). Suggest gating the job on something like `github.event.comment.author_association == 'MEMBER' || == 'OWNER' || == 'COLLABORATOR'`, mirroring the pattern most slash-command workflows in upstream Kubernetes projects use.

.github/workflows/release-test.yml
```diff
@@ -0,0 +1,328 @@
...
+env:
+  GO_VERSION: 1.24.1
+  KIND_VERSION: v0.31.0
+  KUBERNETES_VERSION: v1.31.12
+  VOLCANO_VERSION: v1.12.0
+  LWS_VERSION: v0.7.0
+  SCHEDULER_PLUGINS_VERSION: v0.32.7
+
+# Declare default permissions as read only.
+permissions: read-all
```
cheyang, line 16: Adding `concurrency: { group: release-test-${{ github.event.issue.number }}, cancel-in-progress: true }` (or similar) would prevent stacked runs when someone types `/release-test` twice, or amends and re-triggers. Right now each comment queues a fresh pair of ~30+ minute jobs.

.github/workflows/e2e-test.yml
```diff
@@ -238,10 +226,10 @@ jobs:
         run: |
           kind delete cluster --name rbgs-e2e || true
 
-  # E2E test using manifest installation (only runs when deploy/ changes)
+  # E2E test using manifest installation (runs on every PR, builds images from PR branch)
```
cheyang, line 229: Removing the `deploy_changed` gate and falling back to `should_run` means every PR with any code change pays for both `e2e-test` and `e2e-test-manifest` — two full Kind clusters, two `make test-e2e` invocations. That is the stated goal of the PR, but please confirm: (a) the existing self-hosted/runner budget can absorb this, and (b) the two jobs are actually exercising different code paths (helm vs. manifest install) often enough to justify the duplication. If the second job mostly re-validates the same controller logic, a lighter `manifest apply + smoke` could replace the full e2e suite.

.github/workflows/release-test.yml
```diff
@@ -0,0 +1,328 @@
...
+            const pr = await github.rest.pulls.get({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.issue.number
+            });
+            core.setOutput('head_sha', pr.data.head.sha);
+            core.setOutput('ref', pr.data.head.ref);
+
+  # E2E test using Helm installation with default chart images (no build)
+  e2e-test-helm:
```
cheyang, line 55: The Kind/LWS/Volcano/scheduler-plugins setup, manifest validation, image pre-load, log collection and cleanup blocks are byte-for-byte copies of the corresponding sections in `e2e-test.yml`. Extracting them into a reusable workflow (`workflow_call`) or a composite action would prevent the two files from drifting (e.g. someone bumps `KIND_VERSION` here but forgets the other). Not a blocker for this PR, but worth a follow-up.

PR #376, opened 2026-06-30
kep: refine warmup API definitions

NoobDream2568:

NONE

## Checklist

### DIFF
--- a/keps/129-rbg-warmup/README.md
+++ b/keps/129-rbg-warmup/README.md
@@ -1,79 +1,69 @@
 # KEP-129: Image and Volume Warmup for RoleBasedGroup
 
-## Table of Contents
-
 <!-- toc -->
 
-- [Release Signoff Checklist](#release-signoff-checklist)
 - [Summary](#summary)
 - [Motivation](#motivation)
     - [Goals](#goals)
     - [Non-Goals](#non-goals)
 - [Proposal](#proposal)
+    - [Architecture](#architecture)
     - [User Stories](#user-stories)
         - [Story 1](#story-1)
         - [Story 2](#story-2)
-    - [Implementation Details](#implementation-details)
-        - [API Changes](#api-changes)
-        - [Example](#yaml-example)
+    - [API Changes](#api-changes)
+        - [Spec](#spec)
+        - [Status](#status)
+    - [YAML Example](#yaml-example)
     - [Risks and Mitigations](#risks-and-mitigations)
 - [Design Details](#design-details)
+    - [Controller Logic](#controller-logic)
+        - [WarmUp Job State Machine](#warmup-job-state-machine)
+        - [Pod Construction](#pod-construction)
+        - [Multi-Role Merge Algorithm](#multi-role-merge-algorithm)
+        - [Pod Naming and State Recovery](#pod-naming-and-state-recovery)
+    - [Resource Consumption](#resource-consumption)
+    - [Cleanup and Garbage Collection](#cleanup-and-garbage-collection)
+    - [Failure Handling](#failure-handling)
+        - [Early Termination Behavior](#early-termination-behavior)
+    - [RBAC Requirements](#rbac-requirements)
+    - [Coordination with RBG Upgrades](#coordination-with-rbg-upgrades)
     - [Test Plan](#test-plan)
-    - [Graduation Criteria](#graduation-criteria)
-- [Implementation History](#implementation-history)
-- [Drawbacks](#drawbacks)
+        - [Unit Tests](#unit-tests)
+        - [Integration Tests (envtest)](#integration-tests-envtest)
+        - [E2E Tests (Kind cluster)](#e2e-tests-kind-cluster)
+- [Alternative Designs](#alternative-designs)
+    - [Architecture Alternatives](#architecture-alternatives)
 
 <!-- /toc -->
 
-## Release Signoff Checklist
-
-- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial
-  KEP PR)
-- [ ] (R) KEP approvers have approved the KEP status as `implementable`
-- [ ] (R) Design details are appropriately documented
-- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test
-  refactors)
-    - [ ] e2e Tests for all Beta API Operations (endpoints)
-    - [ ] (R) Ensure GA e2e tests meet requirements
-      for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
-    - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
-- [ ] (R) Graduation criteria is in place
-    - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit
-      by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
-      within one minor version of promotion to GA
-- [ ] (R) Production readiness review completed
-- [ ] (R) Production readiness review approved
-- [ ] "Implementation History" section is up-to-date for milestone
-- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
-- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings,
-  relevant PRs/issues, release notes
-
 ## Summary
 
-This KEP proposes adding image and volume warmup capabilities to RoleBasedGroup (RBG). Cold start is a common problem for serving large language models(LLMs).
+This KEP proposes adding image and volume warmup capabilities to RoleBasedGroup (RBG). Cold start is a common problem for serving large language models (LLMs).
 Cold start latency can be reduced by pre-warming the model files and the container image before a serving instance starts. In RBG's design, upgrades can be done in a fine-grained manner (e.g. [Role Coordination](../30-role-coordination/README.md)). Image and volume warmup can be an additional step to speed up the upgrade process, avoiding resource utilization degradation.
 
 ## Motivation
 
 In a production environment, availability and GPU resource utilization are key metrics for large language model inference services. In-place update reduces time consumption by avoiding recreating Pods, but the cold start latency of the inference service (e.g. SGLang Worker) can still be high and thus affecting service availability. There are two key factors that affect the cold start latency:
-- The size of container's image size (e.g. SGLang v0.5.5 image size is about 27.5GB after decompression)
+- The size of the container image (e.g. SGLang v0.5.5 image size is about 27.5GB after decompression)
 - The size of large language model files (e.g. Total size of Qwen3-32B is about 61GB)
 
-Image and volume warmup can help reduce the cold start latency. By introducing an extra "warmup stage" before the upgrade process, the upgradation process can be more efficient. It takes less time for RoleBasedGroup to rolling update each inference engine instance, so both the service availability and GPU utilization will be improved. 
+Image and volume warmup can help reduce the cold start latency. By introducing an extra "warmup stage" before the upgrade process, the process becomes more efficient. It takes less time for RoleBasedGroup to rolling update each inference engine instance, so both the service availability and GPU utilization will be improved.
 
 Without WarmUp:
 ![alt text](without-warmup.png)
 
-With WarmUp
+With WarmUp:
 ![alt text](with-warmup-stage.png)
 
 ### Goals
 
-- Support WarmUp job that can be submitted by a user. For each node in the cluster, the job warms up images and volumes according to the job's configuration.
+- Support WarmUp job that can be submitted by a user. For each target node, the job warms up images and volumes according to the job's configuration.
 
 ### Non-Goals
 
-- Although it's possible to upgrade RBG and warm up images & volumes in a coordinated way, co-design between WarmUp and RBG's upgradation process is not a goal in this enhancement. 
+- Although it's possible to upgrade RBG and warm up images & volumes in a coordinated way, co-design between WarmUp and RBG's upgrade process is not a goal in this enhancement.
+- `dataCopy` action (copying model files between volumes) is deferred to the future. The current scope focuses on image preload and customized warmup actions.
 
 ## Proposal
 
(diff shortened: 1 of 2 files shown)


keps/129-rbg-warmup/kep.yaml
```diff
@@ -1,16 +1,20 @@
...
+  - "@cheyang"
 
 stage: alpha
 
-latest-milestone: "v0.5.0"
+latest-milestone: "v0.8.0"
 
 milestone:
-  alpha: "v0.5.0"
+  alpha: "v0.8.0"
```
cheyang, line 20: The CRD samples in README use `apiVersion: workloads.x-k8s.io/v1alpha2`, but this is the first version of a brand-new resource and `stage: alpha` here. Starting at `v1alpha2` for an initial introduction is unusual and will likely confuse downstream tooling and conformance checks. Could you either (a) start at `v1alpha1` to match the project's other workload kinds, or (b) add a paragraph explaining why `v1alpha2` is required (e.g., reusing an existing storage version)? The KEP should make this choice explicit, since it has consequences for upgrade tests and version-tolerant clients.

keps/129-rbg-warmup/README.md
```diff
@@ -1,206 +1,366 @@
...
 
 ![alt text](arch.png)
 
-The WarmUp Controller collects node information of all Pods under the target RBG. It deploys one WarmUp Pod to each node. Each WarmUp Pod contains:
-- X init containers for image preload action, configured with image=<new_image> and a startup command set to "exit 0"
-- Y containers for customized action, which allow users to define a customized warmup process in RoleBasedGroupWarmup.
-  - For customized action, RBG Warmup controller periodically checks the container status. If all of the container shift to phase "Ready", the WarmUp process is considered to be completed.
-  - Use Cases for customized warmup: (1) Model download & copy from a external storage; (2) CUDA kernel build before launching a new model inference instance.
+The WarmUp Controller collects node information based on the target configuration (explicit node list, node selector, or from an existing RoleBasedGroup). It deploys one WarmUp Pod to each target node. Each WarmUp Pod contains:
+- N containers for image preload action, configured with `image=<new_image>` and a startup command set to `exit 0`. These containers pull the image and exit immediately.
```
cheyang, line 75: The Architecture section says image preload runs as `N containers` (regular containers), while the prior revision and the Goals narrative talk about init containers. With `restartPolicy: Never` and image-preload as regular containers, the Pod stays in `Running` (and the node is counted as `active`) until every customized-action container also exits — there is no purely-preload fast path. Please pick a model and state the rationale. If init containers are intentionally avoided (e.g., to keep image preload parallel rather than serial), call it out explicitly, because it directly affects how users interpret `status.active`, completion timing, and the `globalTimeoutSeconds` semantics.

keps/129-rbg-warmup/README.md
```diff
@@ -1,206 +1,366 @@
...
+Each warmup Pod contains:
+
+1. **Image preload containers:** One container per unique image, with `command: ["sh", "-c", "exit 0"]` and `imagePullPolicy: IfNotPresent`. These containers pull the image onto the node's container runtime cache and exit immediately.
+
+2. **Customized action containers:** User-defined containers for custom warmup logic (model download, CUDA compilation, etc.), along with their associated volumes.
+
+3. **Pod-level settings:**
+   - `restartPolicy: Never` — failed Pods are preserved for failure counting and log inspection.
+   - `tolerations` from `spec.tolerations` — allows scheduling on tainted nodes (e.g., GPU nodes with `nvidia.com/gpu` taint).
+   - `imagePullSecrets` — deduplicated from all `imagePreload.pullSecrets`.
```
cheyang, line 307: With `restartPolicy: Never`, retries happen by creating a *new* Pod for the same node. The KEP should state the Pod naming scheme: deterministic (e.g., `<warmup>-<node-hash>-<attempt>`) so reconciliation is idempotent across controller restarts, or generated via `generateName` with the controller scanning labels. The choice affects observability and the 'deterministic order' guarantee promised in Step 2. Please also document how attempt indexing survives controller restart — does the controller walk existing Pods to recover the next attempt number?

keps/129-rbg-warmup/README.md
```diff
@@ -1,206 +1,366 @@
...
+| Warmup Pods may get stuck (e.g., image pull backoff)                         | `globalTimeoutSeconds` provides an overall timeout. Controller restart is safe — Pod objects persist and are re-counted.                                                                       |
+| Controller crash during warmup                                               | No in-memory state is required. On restart, the controller re-lists owned Pods and resumes from the current state. Failed Pods are preserved as persistent counters for `backoffLimitPerNode`. |
+
+## Design Details
 
 ### Controller Logic
 
 #### WarmUp Job State Machine
 
 ![alt text](warmup-job-state-machine.png)
```
cheyang, line 280: 'Containers from multiple roles (in `targetRoleBasedGroup` mode) targeting the same node are merged into a single Pod with deduplication.' Please specify the merge/dedup keys precisely: (1) image-preload images — dedup by image reference string, but what about the same image with different `pullSecrets`? (2) customized-action containers — what is the conflict policy when two roles define a container with the same `name`? (3) volumes — name conflicts? Without a defined merge algorithm this becomes implementation-defined behavior, which is exactly what KEPs should pin down before `implementable`.

keps/129-rbg-warmup/README.md
```diff
@@ -1,206 +1,366 @@
...
+| `parallelism`             | `*int32` | Max number of nodes warmed up concurrently. If unset, all nodes run in parallel.                                                                                                                                     |
+| `backoffLimitPerNode`     | `*int32` | Max retries per node before marking it permanently failed. If unset (nil), nodes are retried indefinitely. If set to 0, no retries — first failure is final.                                                         |
+| `maxFailedNodes`          | `*int32` | Max permanently-failed nodes tolerated before failing the entire job. If unset (nil), any number of failures is tolerated. If set to 0, any failure fails the job. Only effective when `backoffLimitPerNode` is set. |
+| `globalTimeoutSeconds`    | `*int64` | Overall timeout measured from `status.startTime`. When exceeded, active Pods are deleted and the job is marked Failed. If unset (nil), no timeout is applied.                                                        |
```
cheyang: `maxFailedNodes` is documented as 'Only effective when `backoffLimitPerNode` is set'. That implicit coupling is easy to misuse. Two requests: (1) capture this constraint as a CEL rule (analogous to the `targetNodes` / `targetRoleBasedGroup` exclusivity) so the API server rejects invalid combinations; (2) clarify what happens when `backoffLimitPerNode` is unset (infinite retries) — does `maxFailedNodes` simply have no effect, or is the field silently ignored? The current wording leaves both interpretations open.

NoobDream2568: **`maxFailedNodes` requires `backoffLimitPerNode`:** When `backoffLimitPerNode` is unset, nodes are retried indefinitely — a node never reaches the "permanently failed" state. In that scenario, `maxFailedNodes` would have no observable effect since the set of permanently-failed nodes is always empty, making it impossible for the threshold to be exceeded. Rather than silently ignoring the field, the API server rejects this combination at admission time via a CEL validation rule: `!has(self.maxFailedNodes) || has(self.backoffLimitPerNode)`. This ensures users get immediate feedback when their policy configuration is logically inconsistent.

keps/129-rbg-warmup/README.md
NoobDream2568: We use **generateName** with label-based scanning, not deterministic pod names. Here's why and how it works:
 
**Pod identification:** Each warmup Pod carries three labels: warmup-name, warmup-uid, and node-name. On every reconciliation, the controller lists all owned Pods by these labels and classifies them into active/succeeded/failed sets. No in-memory state is maintained.

**Why not deterministic names:** Node names can be up to 253 characters (same as Pod name limit). Hashing introduces collision risk, and attempt indexing would either need persistence or be derived from existing Pods — which is exactly what the label-scanning approach already does, without the naming complexity.

**Clarification on "deterministic order":** This refers to node processing order (sort.Strings on node names), not Pod naming. It guarantees that under a parallelism limit, the same set of nodes gets Pods first, regardless of controller restarts. The Pod names themselves are non-deterministic (random suffix via generateName). For observability, users can query by label: `kubectl get pods -l workloads.x-k8s.io/node-name=<node>`.

I will update the KEP to document this explicitly.

keps/129-rbg-warmup/README.md
NoobDream2568: Good catch — the merge behavior was under-specified. We have now pinned down the exact merge rules in the KEP ("Multi-Role Merge Algorithm" section under Pod Construction) and implemented conflict detection in the controller.

### Merge Rules

| Component                    | Dedup Key                           | Behavior                                                                                                                                                                                                                                                                               |
|------------------------------|-------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Image preload containers     | Image reference string              | One container per unique image. Same image referenced by multiple roles produces one container.                                                                                                                                                                                        |
| Pull secrets                 | Secret name                         | Collected at Pod level (`imagePullSecrets`), deduplicated by name. Same image with different pull secrets across roles is handled correctly — all secrets are included and kubelet tries each when pulling.                                                                            |
| Customized action containers | Content hash (FNV32, name excluded) | Deduplicated by hashing the spec with the `name` field cleared, so containers with identical specs but different user-provided names are treated as one. User-provided names are overwritten with `custom-<index>`. Two containers with different specs always produce two containers. |
| Volumes                      | Volume name (first-wins)            | Deduplicated by name. If two roles define the same volume name with different `VolumeSource`, the first definition is used and a **`VolumeConflict` warning event** is emitted on the RoleBasedGroupWarmup CR. Users should use distinct volume names to avoid ambiguity.              |

### Addressing Each Sub-Question

**(1) Same image with different pullSecrets?**

Not a conflict. Images and pull secrets are independent dimensions — images are deduplicated by reference string into separate containers, while pull secrets are collected into `pod.spec.imagePullSecrets` and deduplicated by secret name. A single image referenced by two roles with different pull secrets produces one image-preload container and both secrets at the Pod level. This follows standard Kubernetes behavior where kubelet tries all available pull secrets when pulling an image.

**(2) Two roles define a container with the same name?**

Container names are not used as dedup keys and are not preserved. All customized action containers are deduplicated by a content hash (FNV32 over the JSON-serialized container spec with the `name` field cleared before hashing). User-provided container names are overwritten with `custom-<index>` to guarantee uniqueness within the Pod. This means:
- Two containers with the same name but different specs → two containers (different hashes)
- Two containers with different names but identical specs → one container (same hash, name excluded)

**(3) Volumes — name conflicts?**

First-wins with a warning. When two roles define a volume with the same name but different `VolumeSource`, the first definition encountered is used. The controller emits a `VolumeConflict` warning event on the CR so the user is notified:

```
Events:
  Type     Reason          Message
  ----     ------          -------
  Warning  VolumeConflict  Volume "model-storage" has conflicting definitions across roles on node gpu-node-1, using first definition
```

Users can resolve this by using distinct volume names for each role. In practice, same-named volumes across roles typically reference the same backing storage (e.g., a shared hostPath for model files), so conflicts are rare.

cheyang:
@gemini-code-assist /review

keps/129-rbg-warmup/kep.yaml
```diff
@@ -1,16 +1,20 @@
...
+creation-date: 2025-12-18
 reviewers:
-  - "cheyang.cy@alibaba-inc.com"
+  - "@syspretor"
+  - "@cheyang"
+approvers:
+  - "@syspretor"
+  - "@cheyang"
 
 stage: alpha
```
cheyang, line 16: Prior round 1 raised this and it is still open. The KEP kicks off as a brand-new resource with `stage: alpha`, yet every YAML example in the README uses `apiVersion: workloads.x-k8s.io/v1alpha2` and the narrative around line 93 explicitly says the CRD is introduced at v1alpha2. Either (a) start at v1alpha1 to match the rest of the workloads group, or (b) keep v1alpha2 but add a short paragraph (in `API Changes` or the summary) explaining why the very first served version is v1alpha2 — otherwise reviewers, downstream tooling, and conformance checks will keep flagging this.

keps/129-rbg-warmup/README.md
```diff
@@ -1,206 +1,442 @@
...
+| `imagePreload`     | `*ImagePreloadAction` | Pull container images onto the node ahead of time.   |
+| `customizedAction` | `*CustomizedAction`   | Run user-defined containers for custom warmup logic. |
+
+At least one of `imagePreload` or `customizedAction` must be specified (enforced via CEL).
+
+**ImagePreloadAction:**
+
+| Field         | Type                     | Description                                                |
+|---------------|--------------------------|------------------------------------------------------------|
+| `images`      | `[]string` (min: 1)      | Container images to preload.                               |
```
cheyang, line 147: The KEP spec table calls the image list field `images`, while the merged Go type at `api/workloads/v1alpha2/rolebasedgroupwarmup_types.go` and the CRD schema use `items`. The YAML examples on lines 207, 245 and 249 have the same mismatch. Since this KEP is meant to be normative for the v1alpha2 API, please pick one name and reconcile both sides in this PR: either rename the Go field to `images` (and update the KEP samples accordingly), or update the KEP tables and YAML examples to use `items`. Independently confirmed against the gemini-code-assist inline suggestions.

keps/129-rbg-warmup/README.md
```diff
@@ -1,206 +1,442 @@
...
+|------------------------------------|-------------------------------------------------|
+| `rolebasedgroupwarmups`            | get, list, watch, create, update, patch, delete |
+| `rolebasedgroupwarmups/status`     | get, update, patch                              |
+| `rolebasedgroupwarmups/finalizers` | update                                          |
+| `rolebasedgroups`                  | get, list, watch                                |
+| `pods`                             | get, list, watch, create, delete                |
+| `nodes`                            | get, list, watch                                |
+| `events`                           | create, patch                                   |
+
+Warmup Pods run under the default service account in the target namespace unless the user specifies otherwise via `customizedAction.containers[].securityContext`.
```
cheyang: This sentence conflates two Kubernetes concepts: ServiceAccount selection is controlled by `Pod.spec.serviceAccountName` (plus token projection), not by `container.securityContext`. `securityContext` only controls UID/GID/capabilities/etc. Either drop the `securityContext` reference and state that ServiceAccount is currently fixed to the namespace default (with a follow-up to expose `serviceAccountName` if needed), or introduce a top-level `spec.serviceAccountName` in the WarmUp API. Independently verified — matches the copilot inline observation on this line.

keps/129-rbg-warmup/README.md
```diff
@@ -1,206 +1,442 @@
...
+|------------------------------|-------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Image preload containers     | Image reference string              | One container per unique image. Same image referenced by multiple roles produces one container.                                                                                                                                                                                                             |
+| Pull secrets                 | Secret name                         | All secrets are collected at Pod level (`imagePullSecrets`), deduplicated by name. Same image with different pull secrets across roles is handled correctly — all secrets are included and kubelet tries each when pulling.                                                                                 |
```
cheyang: Small factual nit on the merge table: `imagePullSecrets` are not "tried one by one" by kubelet. Kubelet builds a merged docker credential set from all referenced secrets and picks credentials by registry/path match. The current wording could mislead operators into thinking secret ordering matters. Suggest tightening to something like "all secrets referenced by any role are attached at the Pod level; kubelet chooses matching credentials by registry when pulling."

keps/129-rbg-warmup/README.md
```diff
@@ -1,206 +1,442 @@
...
+1. **Image preload containers:** One container per unique image, with `command: ["sh", "-c", "exit 0"]` and `imagePullPolicy: IfNotPresent`. These containers pull the image onto the node's container runtime cache and exit immediately.
+
+2. **Customized action containers:** User-defined containers for custom warmup logic (model download, CUDA compilation, etc.), along with their associated volumes.
+
+3. **Pod-level settings:**
+   - `restartPolicy: Never` — failed Pods are preserved for failure counting and log inspection.
+   - `tolerations` from `spec.tolerations` — allows scheduling on tainted nodes (e.g., GPU nodes with `nvidia.com/gpu` taint).
+   - `imagePullSecrets` — deduplicated from all `imagePreload.pullSecrets`.
+
+All warmup containers (both image preload and customized actions) run as regular containers, not init containers. This allows parallel image pulling — with multiple large AI inference images (20-30GB each), serial pulling via init containers would significantly increase total warmup time. The tradeoff is that there is no separate "image preload complete" signal; the Pod is considered complete when all containers exit successfully.
```
cheyang: The new paragraph explicitly commits to regular containers (not init containers) so image pulls run in parallel — good clarification, that addresses one of the round-1 blockers. Two follow-up questions worth pinning down in the KEP: (1) With N containers per Pod, kubelet still serialises `PullImage` calls per Pod on many CRIs (containerd's default `max_concurrent_downloads` is 3, and layer decompression is per-image). Is the parallel-pull assumption you're relying on documented anywhere, or is it a per-cluster tuning knob users need to be aware of? (2) Since there is no longer an "image preload complete" signal, does that affect how `customizedAction` containers observe image availability if they depend on one of the preloaded images? A one-line note would help future implementers.

keps/129-rbg-warmup/README.md
```diff
@@ -1,206 +1,442 @@
...
+| `workloads.x-k8s.io/warmup-name` | Name of the owning RoleBasedGroupWarmup CR                            |
+| `workloads.x-k8s.io/warmup-uid`  | UID of the owning CR (distinguishes recreated CRs with the same name) |
+| `workloads.x-k8s.io/node-name`   | Target node this Pod is warming up                                    |
+
+On every reconciliation, the controller lists Pods by these labels and classifies them into active, succeeded, and failed sets. **No in-memory state is maintained** — the controller fully reconstructs its view from existing Pod objects on each reconcile, making it safe across controller restarts.
+
+**Retry counting** relies on the persistence of failed Pod objects (`restartPolicy: Never` ensures failed Pods are not restarted or garbage-collected by the kubelet). The count of failed Pods for a given `node-name` label IS the attempt count. No separate attempt index or counter is stored.
+
+**Deterministic order** (referenced in Step 2) refers to the processing order of pending nodes (`sort.Strings` on node names), not Pod naming. This ensures that under a parallelism limit, the same set of nodes gets Pods first, regardless of controller restarts.
```
cheyang, line 338: The state-recovery section says retry count is derived from the failed-Pod count for a given `node-name` label, and relies on failed Pods persisting because `restartPolicy: Never`. What is the plan when a cluster has a `PodGC` controller / `TerminatedPodGCThreshold` that reaps Failed Pods once the cluster crosses the threshold (default 12500, but often lowered on large clusters)? If those Pods disappear, the derived attempt count will silently reset and `backoffLimitPerNode` becomes best-effort. Worth calling this out explicitly, or persisting a per-node attempt counter in status.

keps/129-rbg-warmup/README.md
```diff
@@ -1,206 +1,442 @@
...
+
+Each warmup Pod contains:
+
+1. **Image preload containers:** One container per unique image, with `command: ["sh", "-c", "exit 0"]` and `imagePullPolicy: IfNotPresent`. These containers pull the image onto the node's container runtime cache and exit immediately.
+
+2. **Customized action containers:** User-defined containers for custom warmup logic (model download, CUDA compilation, etc.), along with their associated volumes.
+
+3. **Pod-level settings:**
+   - `restartPolicy: Never` — failed Pods are preserved for failure counting and log inspection.
+   - `tolerations` from `spec.tolerations` — allows scheduling on tainted nodes (e.g., GPU nodes with `nvidia.com/gpu` taint).
```
cheyang, line 306: The Merge Algorithm section is a big improvement over round 1. One small addition would help: state what happens when a customized-action container references a `volumeMount` whose volume was dropped by the first-wins rule (i.e., the volume name exists but the spec chosen was from another role). Today this would fail at Pod admission with a confusing error. Either document the intended UX (event + status message) or reject the merge earlier.

keps/129-rbg-warmup/kep.yaml
NoobDream2568: We just keep v1alpha2 to maintain consistency with other resources in this API group. In `API Changes`, a short paragraph will be added.

keps/129-rbg-warmup/README.md
NoobDream2568: Fixed.

keps/129-rbg-warmup/README.md
NoobDream2568: The rename from `items` to `images` is **intentional**, as `images` is more descriptive and self-documenting. Both the Go type and the KEP (spec table + YAML examples) will use `images`. The Go type in `api/workloads/v1alpha2/rolebasedgroupwarmup_types.go` will be updated in a follow-up PR.

keps/129-rbg-warmup/README.md
NoobDream2568: Volume row now documents both the event and the condition, and explicitly calls out that containers from the losing role will see the winning role's volume spec.

keps/129-rbg-warmup/README.md
NoobDream2568: Fixed.

keps/129-rbg-warmup/README.md
NoobDream2568: The following will be added in the doc.
> backoffLimitPerNode counts failed Pods per node. If a cluster's PodGC controller reaps terminated Pods (when TerminatedPodGCThreshold is exceeded), the failure count may reset, causing additional retries. This is a best-effort retry limit, not a strict guarantee.

Persisting a per-node attempt counter in .status is worth considering and may be implemented in a future release (potentially Beta).
However, in large clusters, the CRD may not accommodate per-node state for a massive number of nodes. Further discussion is needed to evaluate scalable alternatives.

cheyang:
@gemini-code-assist /review

keps/129-rbg-warmup/README.md
```diff
@@ -1,206 +1,445 @@
...
 
 ### API Changes
 
+A new CRD `RoleBasedGroupWarmup` is introduced under the `workloads.x-k8s.io` API group (v1alpha2). Although this is a brand-new resource, it starts at v1alpha2 rather than v1alpha1 to align with the current storage version of the core `RoleBasedGroup` CRD. All workloads in this API group have already graduated to v1alpha2, and introducing a new resource at v1alpha1 would require unnecessary conversion webhooks from day one.
+
+#### Spec
+
+| Field                  | Type                    | Description                                                                                             |
+|------------------------|-------------------------|---------------------------------------------------------------------------------------------------------|
+| `paused`               | `*bool`                 | Suspends the warmup job. When true, no new warmup Pods are created, but existing Pods continue running. |
```
vicoooo26, line 99: Just for discussion: Do we need to align the `paused` field's behavior with the standard `job.spec.suspend` semantics?

NoobDream2568: The overall behavior is aligned with job.spec.suspend: One difference is timeout accounting — globalTimeoutSeconds is measured as wall-clock time from status.startTime and does not pause when the job is paused.

cheyang (approved):
/lgtm
/approve

PR #377, opened 2026-07-02
chore(rbgs): update Helm chart and default image versions to 0.8.0-al…

cheyang:

fixes #XXXX

## Checklist

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -5,4 +5,4 @@ resources:
 images:
 - name: controller
   newName: rolebasedgroup/rbgs-controller
-  newTag: v0.7.0-888ade7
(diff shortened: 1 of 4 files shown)


cheyang:
Thanks for the review! Addressed the feedback in 6f2b90e7:

- `Chart.yaml`: quoted `appVersion` — now consistent with the "It is recommended to use it with quotes" note in the file.
- `values.yaml`: quoted both image tags (controller and crd-upgrade).

Note: the actual values (`v0.8.0-cea2a47`) contain letters/hyphens and can't be coerced to numbers or scientific notation by YAML, so this is a defensive/consistency improvement rather than a fix for current breakage — but it guards against future overrides and keeps the style consistent.

Syspretor (approved):

PR #382, opened 2026-07-09
feat: compute maxUnavailable correctly when validating rollout strategy

vie-serendipity:
### Ⅰ. Motivation

Currently, `validateRolloutStrategy` calculates `maxUnavailable` without taking `maxSurge` into account, and this should be fixed.

### Ⅱ. Modifications

Use `maxSurge==0` as roundUp parameter of function `GetScaledValueFromIntOrPercent`.

fixes #XXXX

When the mode is stateful, replicas is less than 4, and maxUnavailable is set to 30%, the final maxUnavailable value will be 0, which is unexpected.

## Checklist

- [-] Format your code `make fmt`.
- [-] Add unit tests or integration tests.

### DIFF
--- a/pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control_test.go
+++ b/pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control_test.go
@@ -949,3 +949,16 @@ func TestIsStablyUnhealthyEdgeCases(t *testing.T) {
 		t.Errorf("instance not yet in the map must return false")
 	}
 }
+
+func TestComputeMaxUnavailable_RoundsUpWhenMaxSurgeIsZero(t *testing.T) {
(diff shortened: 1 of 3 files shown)


cheyang:
The fix correctly passes `maxSurge == 0` as the ceiling parameter to `GetScaledValueFromIntOrPercent`, aligning `validateRolloutStrategy` with the existing `computeMaxUnavailable` implementation and preventing the rolling-update deadlock when a percentage-based maxUnavailable would otherwise resolve to 0. The variable rename from `maxAvailable` to `maxUnavailable` improves clarity. Tests cover the fix from three angles (unit, reconciler integration, validation helper). LGTM.

cheyang (approved):
All CI checks passed and maintainer has approved. Auto-approving to satisfy branch protection.

PR #385, opened 2026-07-10
chore: outdated field name in validation

diw-zw:
The validation error messages in `roletemplate_validation.go` referenced the field as `templatePatch`, which does not match the actual API field path `templateRef.patch`. This inconsistency could confuse users when debugging validation failures. This PR corrects the field name in all error messages and updates the corresponding unit test assertion.

**`api/workloads/v1alpha2/roletemplate_validation.go`**
- Updated the error message for missing `patch` when `templateRef` is set: `templatePatch: required when templateRef is set` → `templateRef.patch is required when templateRef is set`.
- Updated the error message for `patch` set without `templateRef`: `templatePatch: only valid when templateRef is set` → `templateRef.patch is only valid when templateRef is set`.
- Updated the corresponding code comments to use `templateRef.patch` consistently.

**`pkg/reconciler/roleinstanceset_reconciler_test.go`**
- Updated the test assertion in `TestRoleInstanceSetReconciler_ValidateRoleTemplateReferences` to match the corrected error message string.

### DIFF
--- a/api/workloads/v1alpha2/roletemplate_validation.go
+++ b/api/workloads/v1alpha2/roletemplate_validation.go
@@ -138,17 +138,17 @@ func validateRoleTemplateFields(
 		// RawExtension cannot be inspected by CEL, so validate in controller.
 		if !hasTemplatePatch {
 			return fmt.Errorf(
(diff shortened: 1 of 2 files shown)


Syspretor (approved):
/lgtm

cheyang:
@gemini-code-assist /review

PR #386, opened 2026-07-10
doc: add doc best-practice/deploy-inference-service

diw-zw:

PR #388, opened 2026-07-10
doc: add doc best-practice/configuring-rolling-updates

diw-zw:


cheyang:
@gemini-code-assist /review

cheyang:
Solid documentation PR adding rolling update best-practice guides in both English and Chinese. Each language gets a concept doc (explaining parameters, rationale, and architecture) and a hands-on operations guide (step-by-step kubectl workflows with expected outputs). Content is technically accurate and well-structured.

Two minor nits:
1. **Typo** in `doc/best-practice/zh/03-configuring-rolling-updates-guide.md`: In the CoordinatedPolicy verification output, `50mm` should be `50m`.
2. **Dangling links**: The "Related Documents" sections in both concept docs reference `01-deploy-inference-service.md`, `02-using-role-templates.md`, and `08-configuring-autoscaling.md` which do not exist yet.

Neither issue is blocking. LGTM overall.

cheyang:
The documentation examples all use `lmsysorg/sglang:v0.5.9` (10GB+) but the actual command is just `sleep 3600` for control-plane validation. This makes it impractical for users who want to quickly follow along — pulling a multi-GB inference image just to run sleep is heavy.

Could you swap the demo image to something lightweight like `busybox:1.36` or `nginx:1.25-alpine`? The rolling update behavior being demonstrated doesn't depend on the inference runtime, so a small image would make the guide much more accessible for testing.

cheyang:
One more user-experience note (separate from the image-size one above):

**CoordinatedPolicy provenance is unclear.** Operation 4 introduces a `CoordinatedPolicy` CRD but doesn't say whether it comes with the RBG Controller install or requires a separate step. First-time users who follow the earlier docs and then hit this one may get "no matches for kind CoordinatedPolicy" and not know how to recover. A one-line "This CRD is installed together with the RBG Controller via the same Helm chart / kustomize" (or the opposite) would fully unblock this.

**Also:** the `kubectl get pods -w` output in Operation 4 shows some Pods entering `Error` state briefly. That's actually normal — `sleep` receives SIGTERM and exits non-zero — but the doc doesn't say so, and a user watching this stream will assume something is broken. A short annotation like `# Error is normal: sleep exits non-zero on SIGTERM` would prevent panic.

cheyang:
@gemini-code-assist /review

cheyang:
Round 3 re-review — thanks for the quick turnaround.

The previous blocker is fully resolved: all four docs (EN/ZH concept + operations guide) now use `alpine:3.23.5` for the sleep-only demos instead of the multi-GB sglang image, and the operations guide adds a clear note explaining that `sleep 3600` is a placeholder for validating the rolling-update control plane without a GPU, with guidance to swap in a real inference command for functional testing. That gets readers through the walk-through in seconds instead of pulling a 10 GB+ image, and keeps the intent obvious.

Content-wise the guides remain well-structured: basic rolling update, partition canary, pause/resume, and the multi-role CoordinatedPolicy path each pair concept doc with a runnable operations doc, and the CoordinatedPolicy same-namespace/same-name binding constraint is called out prominently.

Doc-only change, all CI green, no further concerns from my side. Looks good to merge once a maintainer applies `lgtm` / `approved`.

cheyang (commented):
I exercised the rollout mechanics on a live cluster and the core behavior the guide relies on is correct: patching a non-image field (`command`) triggers a rollout that recreates pods, the update order is high→low ordinal (3→2→1→0), and `partition: N` correctly holds ordinals `< N` while updating `≥ N`. So the Operation walkthroughs reproduce as written.

A few smaller accuracy items (all present in both en/ and zh/):

- The Coordinated-Upgrade parameter table lists `maxSkew` default `"100%"` and types `maxSkew`/`maxUnavailable` as `string`. In the CRD both are `intOrString` (an absolute integer is also valid), and `maxSkew` has no schema default — so `"100%"` is a controller-behavior claim, not schema-enforced. Worth softening the wording or confirming the effective default.
- Operation 3 sets `maxUnavailable: 4` but the prose says pods update "one by one" — with 4 replicas and maxUnavailable 4 they can all update at once.
- In the verification blocks the expected-output lines sit inside the ```bash fences, so a reader copying the whole block runs the `>`-prefixed lines as shell redirections. Splitting the command from its sample output avoids stray errors/files.
- Same repo-root "Installation Guide" link as #386 (no actual install steps behind it).

cheyang (approved):
Approving. I exercised the rollout mechanics on a live cluster and the behavior the guide relies on holds: patching a non-image field triggers a rebuild rollout, update order is high→low ordinal, and partition correctly stages the update. The remaining notes (maxSkew type/default wording, the maxUnavailable-4 vs "one by one" phrasing, and expected-output lines living inside the bash fences) are non-blocking doc polish.

PR #389, opened 2026-07-10
doc: add doc best-practice/configuring-autoscaling

diw-zw:


cheyang:
@gemini-code-assist /review

doc/best-practice/en/08-configuring-autoscaling-guide.md
```diff
@@ -0,0 +1,882 @@
...
+>   replicas: 2
+>   selector: rbg.workloads.x-k8s.io/group-name=scaling-demo,rbg.workloads.x-k8s.io/group-uid=8481eaa0eb18d5d84918728022b8816dfb465d0c,rbg.workloads.x-k8s.io/role-name=backend
+```
+
+**Expected output:**
+
+- RBGSA `scaling-demo-backend` auto-created (naming convention: `<rbg-name>-<role-name>`)
+- RBGSA's `status.replicas` is 2
+- RBGSA's `status.selector` contains Pod label selector
+- RBGSA's `status.phase` is `Ready`
```
cheyang: The YAML output on line 102 shows `phase: Bound`, but the "Expected output" bullet on line 113 states the phase should be `Ready`. These are contradictory. The same issue exists in `zh/08-configuring-autoscaling-guide.md` line 113. If `Bound` is the correct runtime phase, update the expected output to say `Bound`.

doc/best-practice/en/08-configuring-autoscaling.md
```diff
@@ -0,0 +1,525 @@
...
+        port: 9091             # Metrics port
+```
+
+#### Parameter Description (AutoScaler)
+
+| Parameter | Type | Description |
+| --- | --- | --- |
+| `spec.scalingInterval` | int | Scaling decision interval (seconds), default 180 |
+| `spec.pattern.PDDisaggregated.prefill.roleName` | string | Prefill role name, must match the role name in RBG |
+| `spec.pattern.PDDisAggregated.prefill.minReplicas` | int | Prefill minimum replicas |
```
cheyang: The parameter table uses `spec.pattern.PDDisAggregated.prefill.minReplicas` (capital 'A') on lines 412-413, but all YAML examples and line 411 use `PDDisaggregated` (lowercase 'a'). Users copying the capitalized form from the table will get configuration errors. The same inconsistency exists in `zh/08-configuring-autoscaling.md` lines 409-410. Please standardize to `PDDisaggregated` throughout.

doc/best-practice/en/08-configuring-autoscaling.md
```diff
@@ -0,0 +1,525 @@
...
+kubectl get rbgsa
+
+# Check RBG role replica counts
+kubectl get rbg -o wide
+```
+
+## Related Documents
+
++ [Deploying Inference Services with RBG](01-deploy-inference-service.md)
++ [Using RoleTemplates to Reduce Configuration Duplication](02-using-role-templates.md)
```
cheyang: The Related Documents section references `01-deploy-inference-service.md` and `03-configuring-rolling-updates.md`, neither of which exist in the repository yet. While these may be planned for future PRs, broken links degrade the reader experience. Consider adding a note or removing the links until those docs land. Same in `zh/08-configuring-autoscaling.md`.

cheyang:
User-experience feedback from trying to follow the guide:

**Operation 4 (RBG Planner) will fail out of the box:**

- All YAML uses `namespace: inference` but there's no `kubectl create namespace inference` step. `kubectl apply` will fail with "namespaces not found".
- Prerequisites lists KEDA as required for Operation 3 but gives no install command. Users have to leave the doc to find it.
- The `helm install rbg-planner oci://ghcr.io/sgl-project/charts/rbg-planner` line has no version tag. Newcomers won't know which version to pin.
- Verification steps in Operation 4 are `kubectl get` commands without expected output. Users can't tell whether they succeeded or not.

**Operation 2 (HPA) flow is a bit tangled:**

The Pod uses `sleep 3600` as the command (good — no GPU needed), but then to trigger HPA the user has to `kubectl exec` into the Pod and run a `multiprocessing` burn script. The current text mixes these two things and makes it look like `sleep` itself will trigger the scale. Splitting into two explicit steps ("Step A: deploy sleep pods" → "Step B: exec in and generate load") would help.

The KEDA operation's simulated metric server design is very clean — nice work there.

cheyang:
@gemini-code-assist /review

cheyang:
Round-2 re-review. Both blockers from the previous pass are cleared:

- `08-configuring-autoscaling-guide.md`: the RBGSA sample output and the "Expected output" bullet now both say `status.phase: Bound` in the EN and ZH versions.
- `08-configuring-autoscaling.md`: the parameter table now uses `PDDisaggregated` (lowercase `a`) consistently with the YAML samples, so copy/paste from the table no longer breaks the config.

The previously flagged `01-deploy-inference-service.md` / `03-configuring-rolling-updates.md` broken links have also been reworked into a TODO placeholder block, which is a reasonable interim solution.

Doc-only change, CI is green, no functional risk. LGTM from my side — from a docs standpoint this is ready to merge.

cheyang (commented):
The ScalingAdapter foundation checks out on a live cluster: `scalingAdapter.enable: true` creates the RBGSA (short name `rbgsa`, print columns `PHASE/REPLICAS/READY_REPLICAS/AGE`, phase `Bound`, `scaleTargetRef: {name, role}`), and `kubectl scale rbgsa --replicas=N` propagates to the RBG. Operations 1–3 look runnable. Two issues in Operation 4, flagged inline. (Both apply to the zh/ file too.)

doc/best-practice/en/08-configuring-autoscaling-guide.md
```diff
@@ -0,0 +1,880 @@
...
+
+### Step 1: Create PD-Disaggregated Inference Service (with ScalingAdapter Enabled)
+
+```bash
+cat <<'EOF' | kubectl apply -f -
+apiVersion: workloads.x-k8s.io/v1alpha2
+kind: RoleBasedGroup
+metadata:
+  name: pd-autoscaler-demo
+  namespace: inference
```
cheyang: Operation 4 applies resources into `namespace: inference`, but that namespace is never created — the only `kind: Namespace` in the guide creates `monitoring` for Operation 3. Copy-pasting Operation 4 fails on the very first apply: `Error from server (NotFound): namespaces "inference" not found`. Add a `kubectl create namespace inference` (or a Namespace manifest) before the applies.

doc/best-practice/en/08-configuring-autoscaling-guide.md
```diff
@@ -0,0 +1,880 @@
...
+
+      # Dry-run mode (observe only, no scaling executed)
+      dryRun: true
+
+      profiling:
+        image: "ghcr.io/sgl-project/rbg-profiler:latest"
+
+      metricsEndpoint:
+        metricSource: sglang
+        port: 9091
```
cheyang, line 809: The AutoScaler targets `metricsEndpoint.port: 9091`, but the sglang containers in this same operation launch with `--port 8000`, expose only `8000`, and pass no `--enable-metrics` flag — so nothing appears to listen on 9091 in the real (non-KEDA-sim) case, and the Planner would have no metrics source once `dryRun` is off. Is 9091 correct here, or copied from the Operation 3 KEDA simulation (where a Python server genuinely listens on 9091)? Note: the `AutoScaler` / `inference-extension.rolebasedgroup.io` CRD isn't part of this repo, so I couldn't verify the Planner end-to-end — flagging for someone with rbg-planner installed.

cheyang:
@gemini-code-assist /review

cheyang:
Re-reviewed the new commit (Remove namespace field, 535d8d3). Docs-only, CI is green. The one thing that needs a second look: pulling `namespace: inference` out of the Operation 4 RBG and AutoScaler manifests left them inconsistent with the surrounding commands, which still target `-n inference` (the `kubectl wait` right after apply, plus the verify/scale/cleanup blocks). As written, a reader who copies Operation 4 verbatim applies the CRs to the default namespace and then queries `inference`, so the wait hangs and the checks return nothing. Operations 1-3 already work in the default namespace, so dropping `-n inference` from Operation 4 keeps the guide self-consistent. This repeats across the EN and ZH guide plus the EN/ZH reference docs. The earlier fixes (phase casing `PDDisaggregated`, the TODO placeholder for the not-yet-created cross links) are unchanged and fine.

doc/best-practice/en/08-configuring-autoscaling-guide.md
```diff
@@ -0,0 +1,878 @@
...
+> **Note**: This operation requires GPU nodes and a real inference service, validating RBG Planner's PD-disaggregated intelligent scaling capability. If you only need to validate the AutoScaler CR configuration, set `dryRun: true`.
+
+### Step 1: Create PD-Disaggregated Inference Service (with ScalingAdapter Enabled)
+
+```bash
+cat <<'EOF' | kubectl apply -f -
+apiVersion: workloads.x-k8s.io/v1alpha2
+kind: RoleBasedGroup
+metadata:
+  name: pd-autoscaler-demo
```
cheyang, line 690: This RBG (and the AutoScaler block just below) no longer sets `namespace: inference`, so `kubectl apply -f -` creates it in whatever namespace the current context points at — usually `default`. But the rest of Operation 4 still hard-codes the namespace: the `kubectl wait ... -n inference` a few lines down, then all of the verify / scale / cleanup commands (`kubectl get autoscaler -n inference`, `kubectl logs -n inference`, `kubectl delete ... -n inference`, etc.). Following the guide top to bottom, the wait blocks until it times out and the verification commands come back empty. Operations 1-3 verify against the default namespace with no `-n`, so the consistent fix here is to drop `-n inference` from the Operation 4 commands too (rather than re-adding the namespace field). Same mismatch is in the AutoScaler manifest below, and mirrored in `doc/best-practice/zh/08-configuring-autoscaling-guide.md`; the reference docs (`08-configuring-autoscaling.md`, EN and ZH) have the same `-n inference` verify commands against now-namespaceless manifests.

cheyang (approved):
/lgtm
/approve

PR #393, opened 2026-07-10
fix: handle GracePeriodSeconds not correctly working and respect UpdateStrategy.Type

### LINKED ISSUE #383: [Bug] GracePeriodSeconds not honored and UpdateStrategy.Type ignored in Stateful mode
### Checklist

### Bug 1: `gracePeriodSeconds` ignored by RoleInstance controller

When `rollingUpdate.inPlaceUpdateStrategy.gracePeriodSeconds` is configured, the RoleInstance controller's `GetUpdateOptions()` always returns `GracePeriodSeconds = 0`, causing in-place updates to patch the container image immediately without the configured drain delay.

The root cause is that `GetUpdateOptions()` constructs an empty `UpdateOptions{}` and never populates `GracePeriodSeconds` from the configured value. Additionally, `GracePeriodSeconds` is only defined on `RoleInstanceSet.Spec.UpdateStrategy.InPlaceUpdateStrategy` and is not propagated to the RoleInstance, so the RoleInstance controller has no way to read it.

### Bug 2: Stateful mode ignores `UpdateStrategy.Type`

In Stateful mode, `applyTargetUpdate` calls `inPlaceUpdateInstance` unconditionally — it never checks `UpdateStrategy.Type`. As a result:

- `RecreatePod`: in-place update is still attempted when only the image changes, violating the user's explicit intent to always recreate Pods.
- `InPlaceOnly`: behaves identically to `InPlaceIfPossible`, falling back to recreate instead of failing.

The Stateless mode had a partial check (`== InPlaceIfPossibleUpdateStrategyType`) but it also mishandled `InPlaceOnly` — it skipped in-place entirely, falling through to specified-delete.

### Bug 1

1. Create an RBG with `rollingUpdate.type: InPlaceIfPossible` and `inPlaceUpdateStrategy.gracePeriodSeconds: 30`.
2. Wait for all Pods to become Ready.
3. Patch the container image to trigger an update.
4. Observe that Pods are updated within ~1 second — the 30-second grace period is not honored.

### Bug 2

1. Create an RBG with `rollingUpdate.type: RecreatePod`.
2. Wait for all Pods to become Ready.
3. Patch only the container image.
4. Observe that Pods are updated in-place (RESTARTS increases, AGE unchanged) instead of being deleted and recreated.

### Environment

- RBG Controller: v0.8.0-alpha1
- Kubernetes: >= 1.24
- Cluster: Kind

NoobDream2568:

fixes #383

## Checklist

### DIFF
--- a/pkg/reconciler/roleinstance/core/instance_core.go
+++ b/pkg/reconciler/roleinstance/core/instance_core.go
@@ -19,6 +19,7 @@ package core
 import (
 	"encoding/json"
 	"fmt"
+	"strconv"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -26,6 +27,7 @@ import (
 	"k8s.io/klog/v2"
 	kubecontroller "k8s.io/kubernetes/pkg/controller"
 
+	"sigs.k8s.io/rbgs/api/workloads/constants"
 	workloadsv1alpha2 "sigs.k8s.io/rbgs/api/workloads/v1alpha2"
 	inplaceupdatepod "sigs.k8s.io/rbgs/pkg/inplace/pod"
 	podinplaceupdate "sigs.k8s.io/rbgs/pkg/inplace/pod/inplaceupdate"
@@ -163,6 +165,11 @@ func (c *commonControl) setComponentPodIdentity(component *workloadsv1alpha2.Rol
 func (c *commonControl) GetUpdateOptions() *podinplaceupdate.UpdateOptions {
 	opts := &podinplaceupdate.UpdateOptions{}
 	podinplaceupdate.SetOptionsDefaults(opts)
+	if v, ok := c.Annotations[constants.RoleInplaceUpdateGracePeriodSecondsKey]; ok {
+		if seconds, err := strconv.Atoi(v); err == nil {
+			opts.GracePeriodSeconds = int32(seconds)
+		}
+	}
 	return opts
 }
 
--- a/pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control.go
+++ b/pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control.go
@@ -744,9 +744,13 @@ func (ssc *defaultStatefulInstanceSetControl) applyTargetUpdate(
 	klog.InfoS("Updating instance", "instanceSet", klog.KObj(set), "instance", klog.KObj(target),
 		"from", getInstanceRevision(target), "to", updateRevision.Name,
 		"isSurge", isSurgeSlot, "free", isFree)
-	inplacing, err := ssc.inPlaceUpdateInstance(set, target, updateRevision, revisions)
-	if err != nil {
-		return false, err
+	var inplacing bool
+	var err error
+	if set.Spec.UpdateStrategy.Type != workloadsv1alpha2.RecreatePodUpdateStrategyType {
+		inplacing, err = ssc.inPlaceUpdateInstance(set, target, updateRevision, revisions)
+		if err != nil {
+			return false, err
+		}
 	}
 	transitioned := inplacing
 	if !inplacing {
--- a/pkg/reconciler/roleinstanceset/statelessmode/sync/update.go
+++ b/pkg/reconciler/roleinstanceset/statelessmode/sync/update.go
@@ -179,7 +179,7 @@ func (rc *realControl) updateInstance(set *workloadsv1alpha2.RoleInstanceSet, co
 	instance *workloadsv1alpha2.RoleInstance,
 ) (time.Duration, error) {
 
-	if set.Spec.UpdateStrategy.Type == workloadsv1alpha2.InPlaceIfPossibleUpdateStrategyType {
+	if set.Spec.UpdateStrategy.Type != workloadsv1alpha2.RecreatePodUpdateStrategyType {
 		var oldRevision *apps.ControllerRevision
 		for _, r := range revisions {
 			if utils.EqualToRevisionHash("", instance, r.Name) {
--- a/pkg/reconciler/roleinstanceset/statelessmode/sync/update_test.go
+++ b/pkg/reconciler/roleinstanceset/statelessmode/sync/update_test.go
@@ -1,0 +1,156 @@
+/*
+Copyright 2026 The RBG Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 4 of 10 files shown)


cheyang:
@gemini-code-assist /review

cheyang:
The two fixes are logically correct and follow established patterns in the codebase (annotation propagation mirrors how scheduling-related annotations already work). However, none of the behavioral changes are backed by unit tests. For a fix that modifies controller update paths, please add at least basic test cases covering: (1) `GetUpdateOptions()` reading grace period from the new annotation, (2) `RecreatePod` strategy correctly skipping in-place update in both stateful and stateless modes, and (3) `InPlaceOnly` strategy now triggering in-place update in stateless mode.

pkg/reconciler/roleinstance/core/instance_core.go
```diff
@@ -163,6 +165,11 @@ func (c *commonControl) setComponentPodIdentity(component *workloadsv1alpha2.Rol
 func (c *commonControl) GetUpdateOptions() *podinplaceupdate.UpdateOptions {
 	opts := &podinplaceupdate.UpdateOptions{}
```
cheyang, line 166: The updated `GetUpdateOptions()` now reads `RoleInplaceUpdateGracePeriodSecondsKey` from annotations and parses it into `GracePeriodSeconds`. This behavioral change lacks unit test coverage. Consider adding tests that verify: (1) when the annotation is present with a valid integer, `GracePeriodSeconds` is set correctly, (2) when the annotation is absent, the default is used, (3) when the annotation contains a non-numeric value, parsing fails gracefully.

pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control.go
```diff
@@ -744,9 +744,13 @@ func (ssc *defaultStatefulInstanceSetControl) applyTargetUpdate(
 	klog.InfoS("Updating instance", "instanceSet", klog.KObj(set), "instance", klog.KObj(target),
 		"from", getInstanceRevision(target), "to", updateRevision.Name,
 		"isSurge", isSurgeSlot, "free", isFree)
-	inplacing, err := ssc.inPlaceUpdateInstance(set, target, updateRevision, revisions)
-	if err != nil {
-		return false, err
+	var inplacing bool
+	var err error
+	if set.Spec.UpdateStrategy.Type != workloadsv1alpha2.RecreatePodUpdateStrategyType {
+		inplacing, err = ssc.inPlaceUpdateInstance(set, target, updateRevision, revisions)
```
cheyang, line 750: The change from unconditionally calling `inPlaceUpdateInstance` to gating on `Type != RecreatePodUpdateStrategyType` is a significant behavioral change. A unit test should confirm that when `UpdateStrategy.Type` is `RecreatePod`, the method skips directly to delete-and-recreate, and when it is `InPlaceOnly`, in-place update is attempted.

pkg/reconciler/roleinstanceset/statelessmode/sync/update.go
```diff
@@ -179,7 +179,7 @@ func (rc *realControl) updateInstance(set *workloadsv1alpha2.RoleInstanceSet, co
 	instance *workloadsv1alpha2.RoleInstance,
 ) (time.Duration, error) {
 
-	if set.Spec.UpdateStrategy.Type == workloadsv1alpha2.InPlaceIfPossibleUpdateStrategyType {
+	if set.Spec.UpdateStrategy.Type != workloadsv1alpha2.RecreatePodUpdateStrategyType {
```
cheyang, line 182: The condition change from `== InPlaceIfPossibleUpdateStrategyType` to `!= RecreatePodUpdateStrategyType` broadens in-place update eligibility to include `InPlaceOnly`. This is the correct semantic fix but should have a corresponding test to prevent regression.

pkg/reconciler/roleinstanceset_reconciler.go
```diff
@@ -233,10 +234,16 @@ func (r *RoleInstanceSetReconciler) constructRoleInstanceSetApplyConfiguration(
 			updateStrategyConfig = updateStrategyConfig.WithInPlaceUpdateStrategy(
 				workloadsv1alpha2client.RoleInstanceSetInPlaceUpdateStrategy().WithGracePeriodSeconds(rollingUpdate.InPlaceUpdateStrategy.GracePeriodSeconds),
 			)
+			// Propagate grace period as annotation so RoleInstance controller can read it.
+			roleInstanceSetAnnotation[constants.RoleInplaceUpdateGracePeriodSecondsKey] = strconv.Itoa(int(rollingUpdate.InPlaceUpdateStrategy.GracePeriodSeconds))
 		}
```
cheyang, line 245: The first `WithAnnotations` at line 202 is overwritten by the second at line 240 when the grace period annotation is appended. Consider moving the grace period annotation population before the initial `WithAnnotations` call so annotations are set exactly once. Alternatively, move the initial `WithAnnotations` below the strategy block. Not a correctness issue, purely a readability improvement.

cheyang:
@gemini-code-assist /review

pkg/reconciler/roleinstanceset/statelessmode/sync/update_test.go
```diff
@@ -0,0 +1,156 @@
...
+func (f *fakeStatelessInplaceControl) Refresh(_ *workloadsv1alpha2.RoleInstance, _ *inplaceupdate.UpdateOptions) inplaceupdate.RefreshResult {
+	return inplaceupdate.RefreshResult{}
+}
+
+func TestUpdateInstanceRecreatePodSkipsInPlace(t *testing.T) {
+	scheme := runtime.NewScheme()
+	_ = workloadsv1alpha2.AddToScheme(scheme)
+
+	instance := &workloadsv1alpha2.RoleInstance{
+		ObjectMeta: metav1.ObjectMeta{
```
cheyang, line 61: The new `TestUpdateInstanceRecreatePodSkipsInPlace` verifies the negative branch (`RecreatePod` skips in-place), which is exactly the behavior the earlier review asked for and it looks great. As a small follow-up, consider adding a mirror case that constructs a set with `UpdateStrategy.Type = InPlaceOnly` and asserts that `CanUpdateInPlace`/`Update` on the inplace control **are** invoked. That would make the coverage matrix explicit for the `!= RecreatePodUpdateStrategyType` condition on line 182 of `update.go` and lock in the positive path against future regressions. This is a nit — not blocking.

cheyang (approved):
/lgtm
/approve

Issue #380, opened 2026-07-07
[RFC] Improve RoleInstance restart policy: recovery state with exponential backoff and crash-pod restart tolerance


Syspretor:
### Current Behavior

When `RestartPolicy=RecreateRoleInstanceOnPodRestart` (the default for `LeaderWorkerPattern` and `CustomComponentsPattern`), any pod crash triggers the following flow in `pkg/reconciler/roleinstance/sync/instance_scale.go:85-90`:

```go
if c.shouldRecreateInstanceGuarded(...) {
    setRestartingCondition(updateInstance)
    c.recorder.Event(updateInstance, v1.EventTypeNormal, "ReCreateInstance", ...)
    // Deletes ALL pods in the instance, including the crashed one
    return &expectationDiff{toDeleteNum: len(allPods), toDeletePod: allPods}, nil
}
```

This means **all pods** in the RoleInstance (leader + all workers) are deleted and recreated from scratch.

### Problems (Issue #363)

**1. Loss of crash observability**

The crashed pod is deleted before operators can run `kubectl logs -p` or inspect the failure. No crash context (exit code, reason, which pod/container) is preserved after deletion.

**2. Infinite restart loop**

If the crash cause persists (e.g., misconfiguration, hardware fault), the recreate cycle repeats indefinitely with no backoff and no retry limit, wasting resources and destabilizing the cluster.

---

## Design Goals

1. **Preserve crash context**: Keep the crashed pod alive so its previous container logs remain accessible via `kubectl logs -p`
2. **Recovery state with lock semantics**: Enter a well-defined recovery state that prevents cascading sibling rebuilds
3. **One rebuild + hold strategy**: Rebuild siblings once on recovery entry, then hold them stable while waiting for the crashed pod to self-heal or force-recreate
4. **Break infinite loops with exponential backoff**: Configurable base/max delay with exponential growth for force-recreate, plus a limit on how many kubelet-driven restarts of the crashed pod are tolerated
5. **Backward compatible**: Existing users should not perceive behavior changes unless they opt into new configuration

---

### 1. Recovery State: One Rebuild + Hold

When a pod crashes, the RoleInstance enters a **Recovery state** (tracked by `RecoveryStatus` in `RoleInstanceStatus`). Recovery and restart are the same concept — there is no separate "Restarting" mechanism.

#### Recovery Entry Criteria

The RoleInstance enters recovery when:
- The instance was previously Ready (stable state)
- A **faulty pod** is detected:
[...]

PR #394, opened 2026-07-10
feat: add exponential backoff for restart policy

### LINKED ISSUE #363: [Enhancement] RecreateRoleInstanceOnPodRestart destroys observability and causes infinite recreate loops
**Problem:**

I found when any container's RestartCount > 0, the controller immediately deletes all pods. This causes two issues in practice:

  1. No time to investigate: The pod disappears before operators can run kubectl logs -p or check what went wrong. No crash info
  (exit code, reason) is saved anywhere after deletion.
  2. Infinite restart loop: If the crash cause doesn't go away, the recreate cycle just keeps repeating with no slowdown.

**Solution:**

I think there are some possible directions to improve this:

- Save crash info (exit code, reason) into RoleInstance.Status before deleting the pod, but only crash info might not be enough
- Only trigger recreation after N consecutive restarts, not on the very first one
- Add backoff between recreation attempts

Syspretor:
## Summary

Implements the simplified restart policy enhancement from #380:

- **Preserves existing restart behavior**: When `RestartPolicy=RecreateRoleInstanceOnPodRestart` triggers, all pods (including the crashed one) are deleted and recreated — unchanged.
- **Adds exponential backoff delay**: Configurable `baseDelaySeconds` and `maxDelaySeconds` control the pacing between restart attempts. The delay grows as `min(base * 2^restartCount, max)`, preventing infinite rapid restart loops.
- **Tracks restart history**: `RestartCount` and `LastRestartTime` in `RoleInstanceStatus` persist across reconciles to compute backoff.

### New API fields

| Field | Location | Default | Description |
|-------|----------|---------|-------------|
| `baseDelaySeconds` | LeaderWorkerPattern, CustomComponentsPattern | 30 | Base delay between restart attempts |
| `maxDelaySeconds` | LeaderWorkerPattern, CustomComponentsPattern | 600 | Cap on exponential backoff |

### Example

```yaml
spec:
  roles:
  - name: worker
    leaderWorkerPattern:
      restartPolicy: 
        type: RecreateRoleInstanceOnPodRestart
        baseDelaySeconds: 30    # first retry after 30s
        maxDelaySeconds: 600    # cap at 10 minutes
```

### Backoff progression (defaults)

```
Restart 0: 30s delay
Restart 1: 60s delay
Restart 2: 120s delay
Restart 3: 240s delay
Restart 4: 480s delay
Restart 5+: 600s (capped)
```

## Test plan

- [x] Unit tests for `calculateRestartDelay` (exponential formula, overflow protection, cap)
- [x] Unit tests for `checkRestartBackoff` (no backoff on first restart, delay elapsed/not elapsed, RestartPolicy=None)
- [x] All existing unit tests pass (`make test`)
- [x] Lint passes (`make lint`)
- [ ] E2E tests (requires Kind cluster)

## Related

- Closes #363 (partial — infinite loop problem)
- RFC: #380


cheyang:
@gemini-code-assist /review

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -318,6 +338,77 @@ func (c *realControl) shouldRecreateInstanceGuarded(ctx context.Context, instanc
...
+	// Prevent overflow for large restart counts
+	if restartCount > 30 {
+		return maxDelaySeconds
+	}
+	delay := baseDelaySeconds * (1 << restartCount)
+	if maxDelaySeconds > 0 && delay > maxDelaySeconds {
+		return maxDelaySeconds
+	}
+	return delay
+}
```
cheyang, line 495: This overflows int32 when `restartCount` is in the ~27-30 range (e.g., `30 * (1<<27) = 4B > int32 max`). The current `> 30` guard doesn't cover this window. The wrapped negative value then bypasses the max-cap check and gets returned as-is, effectively disabling backoff under sustained crashes.

Suggestion: do the shift in int64, cap, then convert back:
```go
delay := int64(baseDelaySeconds) * (int64(1) << restartCount)
if maxDelaySeconds > 0 && delay > int64(maxDelaySeconds) {
    return maxDelaySeconds
}
if delay > math.MaxInt32 {
    return maxDelaySeconds
}
return int32(delay)
```

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -67,12 +74,23 @@ type expectationDiff struct {
...
 func (c *realControl) calculateDiffsWithExpectation(ctx context.Context, updateInstance *workloadsv1alpha2.RoleInstance,
 	currentRevision, updateRevision *apps.ControllerRevision,
 	revisions []*apps.ControllerRevision, pods []*v1.Pod, inactivePods []*v1.Pod) (*expectationDiff, error) {
 
+	// Check restart backoff BEFORE any pod processing (including inactive pod cleanup).
+	// If the delay from the last restart hasn't elapsed, skip all scaling and requeue.
+	// This preserves crashed pods during the backoff window for log observability.
+	if requeue := c.checkRestartBackoff(ctx, updateInstance, pods, inactivePods); requeue > 0 {
+		return &expectationDiff{requeueAfter: requeue}, nil
+	}
```
cheyang, line 101: nit (non-blocking): `RestartCount` only grows and is never reset. After the instance has been healthy for a long stretch (say, `2 * maxDelaySeconds`), consider zeroing it out so a future one-off crash doesn't immediately land at the maximum backoff. Could be a follow-up PR.

api/workloads/v1alpha2/roleinstance_types.go
```diff
@@ -44,6 +53,28 @@ type RoleInstanceReadinessGate struct {
 	ConditionType RoleInstanceConditionType `json:"conditionType"`
 }
 
+// Default values for restart policy delay configuration on RoleInstance.
```
cheyang: nit: these duplicate `DefaultBaseDelaySeconds` / `DefaultMaxDelaySeconds` from `helper.go` with different names but identical values. Minor, but sharing one pair reduces the chance they drift out of sync.

cheyang:
@gemini-code-assist /review

cheyang:
Both previous blocking issues resolved (int32 overflow fixed with int64 arithmetic, e2e tests for backoff added). Implementation is solid. The unrelated CI failure in `[RoleInstanceSet] surge maintains full readiness during rolling update` is not caused by this PR. LGTM.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -318,6 +338,81 @@ func (c *realControl) shouldRecreateInstanceGuarded(ctx context.Context, instanc
...
+	return 0
+}
+
+// calculateRestartDelay computes the exponential backoff delay for a restart attempt.
+// Formula: min(baseDelay * 2^restartCount, maxDelay)
+// Following client-go's workqueue.ItemExponentialFailureRateLimiter convention.
+// When maxDelaySeconds is 0, no cap is applied (unbounded backoff).
+func calculateRestartDelay(baseDelaySeconds, maxDelaySeconds, restartCount int32) int32 {
+	if baseDelaySeconds == 0 {
```
cheyang: nit (non-blocking): With `restartCount >= 63`, `int64(1) << restartCount` wraps to `math.MinInt64`, and the multiplication overflows int64 -- bypassing both cap checks. Practically unreachable with default maxDelay=600 (cap triggers at restartCount=5) plus the reset mechanism, but a one-liner `if restartCount >= 62 { restartCount = 62 }` before the shift would seal it completely.

cheyang:
@gemini-code-assist /review

pkg/reconciler/roleinstance/instance_reconciler.go
```diff
@@ -192,8 +195,29 @@ func (r *reconciler) syncInstance(ctx context.Context, instance *workloadsv1alph
 				break
 			}
 		}
+		// Propagate restart tracking only when updateRestartTracking modified
+		// the values during Scale (i.e., a restart-policy-triggered deletion
+		// was triggered). ApplyRevision produces zero-value status fields, so
+		// unconditional propagation would clobber the correct values from the
+		// live object during the pod creation phase.
+		if updateInstance.Status.RestartCount > 0 {
```
cheyang, line 203: This push introduces the conditional propagation `if updateInstance.Status.RestartCount > 0 { newStatus.RestartCount = updateInstance.Status.RestartCount }` and the matching guard for `LastRestartTime`. It only fires while `scaling == true`, which is only true on the reconcile where `updateRestartTracking` actually incremented the counter. On the very next reconcile (pods already deleted / being recreated, or Ready again) `scaling` is `false`, so we skip the block entirely and `UpdateInstanceStatus` writes the freshly-initialized `newStatus.RestartCount = 0` / `LastRestartTime = nil` back to the API server. The counter is effectively zeroed one reconcile after each restart, which breaks backoff exactly when it needs to kick in.

CI on this SHA reflects the regression:

- envtest `TestRestartPolicy` – three new `Exponential Backoff` cases time out at 60s:
  - `should track RestartCount and LastRestartTime after recreation` (line 867)
  - `should delay second recreation due to backoff` (line 930)
  - `should reset RestartCount after stable period` (line 1060)
- e2e `restart policy stability` – `RecreateRoleInstanceOnPodRestart with backoff delays second recreation` fails after ~1s at line 345 (pod gets recreated during the backoff window). Repeats deterministically in both `e2e-test` and `e2e-test-manifest` matrix jobs.

These are the tests this PR itself added, so the failure is a direct signal that the new propagation path is losing state, not that the tests are flaky. The right fix is likely to either propagate RestartCount / LastRestartTime unconditionally from the live instance (or from `updateInstance` after ApplyRevision) even when `scaling == false`, or to seed `newStatus` from the current `instance.Status` for these two fields before Scale runs. Please restore persistence and re-run envtest+e2e locally before pushing again.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -318,6 +338,81 @@ func (c *realControl) shouldRecreateInstanceGuarded(ctx context.Context, instanc
...
+
+	delay := calculateRestartDelay(baseDelay, maxDelay, instance.Status.RestartCount)
+	if delay == 0 {
+		return 0
+	}
+
+	elapsed := time.Since(instance.Status.LastRestartTime.Time)
+	remaining := time.Duration(delay)*time.Second - elapsed
+	if remaining > 0 {
+		klog.Infof("Restart backoff: instance %s waiting %v (restartCount=%d, delay=%ds)",
```
cheyang: The backoff-waiting message is now logged at Info instead of V(2). For a role instance in a genuine crash loop this reconciles on every backoff tick, so the operator log will get one line per instance per requeue. Not a correctness issue, but worth double-checking whether Info is the right level here — V(1) or V(2) is closer to the surrounding conventions and avoids drowning out other events during long backoff windows.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -327,6 +422,9 @@ func (c *realControl) isAlreadyRestarting(ctx context.Context, instance *workloa
 		return true
 	}
 	// Slow path: read fresh from API server (bypasses informer cache)
+	if c.apiReader == nil {
```
cheyang, line 506: The added `if c.apiReader == nil { return false }` guard is a sensible defensive check and matches the pattern used elsewhere for unit-testable controls. No action needed.

cheyang (commented):
Thanks for the PR — the direction is right, and introducing backoff addresses the "instant furious restart" problem. Nice touches too: the int64 overflow guard and following client-go's `ItemExponentialFailureRateLimiter` convention.

Against issue #363, I left a few inline suggestions below. Three that don't map to a specific line:

- **No durable crash forensics for the unattended case**: preserving the pod during the backoff window is great for live `kubectl logs -p`, but once the window closes the evidence is gone. Consider recording podName/node/container/exitCode/reason/finishedAt into status (e.g. `LastCrashInfo`) before the rebuild, so a 3am crash can still be diagnosed the next morning.
- **Consider grouping the API fields**: `baseDelaySeconds`/`maxDelaySeconds` are flat on both patterns; if more recovery-related knobs get added later, a `RestartPolicyConfig` sub-struct would be easier to extend and wouldn't touch the `RestartPolicy` field type.
- **The defaults change recovery behavior**: `base=30/max=600` turns "rebuild immediately on crash" into "wait 30s first, escalating to 10min". Worth calling this default change out in the PR description, or consider a more conservative default (e.g. backoff off by default / opt-in).

Nice starting point overall.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -363,7 +517,12 @@ func shouldRecreateInstance(instance *workloadsv1alpha2.RoleInstance, pods []*v1
 	// Only trigger when Instance was previously Ready (stable state)
 	// and is NOT in the middle of any update.
 	// This avoids triggering recreate during initial creation, scaling up, or rolling/in-place updates.
-	if !wasInstanceReady(instance) || instance.Generation != instance.Status.ObservedGeneration {
+	// Exception: if LastRestartTime is set, a previous restart-policy recreation occurred,
+	// meaning the instance WAS stable before. The status update during backoff may have set
+	// Ready=False (because pods are Failed), so wasInstanceReady() returns false. We bypass
+	// this check to allow the recreation to proceed after backoff expires.
+	if (!wasInstanceReady(instance) && instance.Status.LastRestartTime == nil) ||
```
cheyang, line 551: 🔴 **This weakens the Ready invariant of `shouldRecreateInstance` and may regress.**

Once `LastRestartTime` has been set even once, the instance will trigger a rebuild **even if it was never Ready**. For example, a pod that keeps crashing during init and never reaches Ready would keep satisfying the recreate condition.

Suggest not using `LastRestartTime != nil` to permanently bypass the Ready gate. Instead, use an explicit "recovering" state (the `Restarting` condition): relax the Ready check only while that state is active, and clear it once recovery completes.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -318,6 +355,120 @@ func (c *realControl) shouldRecreateInstanceGuarded(ctx context.Context, instanc
...
+	// Read fresh from API server for ALL restart-related fields. The informer
+	// cache may be stale for RestartCount, LastRestartTime, AND the Restarting
+	// condition. Stale Restarting=True (from a previous restart that has since
+	// completed) would incorrectly block both backoff and recreation.
+	restartCount := instance.Status.RestartCount
+	lastRestartTime := instance.Status.LastRestartTime
+	restarting := isInstanceRestarting(instance)
+	if c.apiReader != nil {
+		fresh := &workloadsv1alpha2.RoleInstance{}
+		if err := c.apiReader.Get(ctx, client.ObjectKeyFromObject(instance), fresh); err == nil {
```
cheyang: 🔴 **State model doesn't fit the reconcile loop well: several `apiReader.Get` direct reads.**

Here (and at lines 112, 485) you read straight from the API server to work around the informer cache, potentially multiple times per reconcile. That's usually a sign the state has more than one source of truth, and it adds apiserver load.

Suggest anchoring the backoff on the `Restarting` condition's `LastTransitionTime` (naturally visible within reconcile), combined with the in-memory `restartingCache` for informer lag, so you don't need a fresh read every round.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -318,6 +355,120 @@ func (c *realControl) shouldRecreateInstanceGuarded(ctx context.Context, instanc
...
+// Formula: min(baseDelay * 2^restartCount, maxDelay)
+// Following client-go's workqueue.ItemExponentialFailureRateLimiter convention.
+// When maxDelaySeconds is 0, no cap is applied (unbounded backoff).
+func calculateRestartDelay(baseDelaySeconds, maxDelaySeconds, restartCount int32) int32 {
+	if baseDelaySeconds == 0 {
+		return 0
+	}
+	// Use int64 arithmetic to prevent overflow for large restart counts.
+	// Example: baseDelay=30, restartCount=27 → 30 * 2^27 = 4B > int32 max.
+	delay := int64(baseDelaySeconds) * (int64(1) << restartCount)
```
cheyang: 🟡 **Overflow at very large `restartCount`.**

If `restartCount` grows to 63+: `int64(1) << 63` goes negative and `<< 64` becomes 0, so the computed `delay` turns negative or zero → backoff silently disappears.

Suggest moving the comparison before the shift: if `restartCount` is already large enough that `base << count` must exceed `max`, return `max` directly; or cap the shift count (e.g. via `bits.Len`).

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -509,6 +668,32 @@ func setRestartingCondition(instance *workloadsv1alpha2.RoleInstance) {
...
+// accumulated maximum.
+func updateRestartTracking(instance *workloadsv1alpha2.RoleInstance) {
+	// Reset counter if the instance has been stable for a long period.
+	if instance.Status.LastRestartTime != nil {
+		maxDelay := instance.Spec.GetMaxDelaySeconds()
+		stableThreshold := time.Duration(maxDelay) * 2 * time.Second
+		if stableThreshold < 10*time.Minute {
+			stableThreshold = 10 * time.Minute
+		}
+		if time.Since(instance.Status.LastRestartTime.Time) > stableThreshold {
```
cheyang, line 715: 🟡 **The `RestartCount` reset semantics are a bit fragile.**

This only resets "when the next restart happens", based on "time since the last restart", rather than "the instance has stayed Ready". A crashloop where each rebuild takes longer than `maxDelay*2` before crashing again would get reset every round.

Suggest gating the reset on `wasInstanceReady` plus a sustained stability window (with hysteresis), so a flapping recovery can't keep zeroing the counter.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -318,6 +355,120 @@ func (c *realControl) shouldRecreateInstanceGuarded(ctx context.Context, instanc
...
+			if p.Status.Phase == v1.PodFailed && p.DeletionTimestamp == nil {
+				hasFailedPods = true
+				break
+			}
+		}
+		if !hasFailedPods {
+			return 0
+		}
+		// No previous restart tracking from informer — check fresh data below
+		if instance.Status.LastRestartTime == nil {
```
cheyang: ⚪ **Empty dead code.**

This `if` body is empty with only a comment — no actual behavior. Suggest removing it, or filling in the intended fresh-read logic.

cheyang:
@gemini-code-assist /review

pkg/reconciler/roleinstance/sync/instance_scale.go
cheyang: The envtest `RestartPolicy Controller Integration > Exponential Backoff > should delay second recreation due to backoff` at `test/envtest/testcase/restart_policy/restart_policy_test.go:952` is still red on this SHA. The `Consistently` block that expects the pod to survive the backoff window fails, which means recreation proceeds immediately on the second crash. The three fresh-API-reader workarounds added this round did not close the gap opened by the round-3 persistence-gating change. Given this is the same failure surface as round 3 and the fix is now sprawling across two methods, please reconsider the persistence model itself (write `RestartCount` / `LastRestartTime` unconditionally at the moment of restart, before any `wasInstanceReady` / `Ready`-condition side effect) rather than layering more informer-fallback reads.

cheyang: Lint fails on this file with `SA9003: empty branch`. The block `if instance.Status.LastRestartTime == nil { // Also check fresh from API server below }` has no body — it's just a comment. It doesn't affect control flow, so the lint fail is legitimate. Either drop the block entirely and rely on the fresh-read logic below, or hoist the comment into a doc block above the fresh-read section.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -67,12 +74,23 @@ type expectationDiff struct {
...
 	toScaleRoleIDS map[string]sets.Set[int32]
+
+	// requeueAfter is set when the reconcile should be requeued after a delay
+	// without performing any scaling actions (used by restart backoff).
+	requeueAfter time.Duration
 }
 
 func (c *realControl) calculateDiffsWithExpectation(ctx context.Context, updateInstance *workloadsv1alpha2.RoleInstance,
 	currentRevision, updateRevision *apps.ControllerRevision,
 	revisions []*apps.ControllerRevision, pods []*v1.Pod, inactivePods []*v1.Pod) (*expectationDiff, error) {
```
cheyang, line 85: Lint also fails on gocyclo complexity 33 in `calculateDiffsWithExpectation`. Each new informer-fallback path added this round bumped cyclomatic count over the linter budget. Extracting the fresh-restart-tracking sync into a helper (e.g. `refreshRestartTrackingFromAPI(ctx, ri)`) would drop this method under 30 and keep the call-site behaviour equivalent to today's inline code.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -318,6 +355,120 @@ func (c *realControl) shouldRecreateInstanceGuarded(ctx context.Context, instanc
...
+			}
+		}
+		if !hasFailedPods {
+			return 0
+		}
+		// No previous restart tracking from informer — check fresh data below
+		if instance.Status.LastRestartTime == nil {
+			// Also check fresh from API server below
+		}
+	}
```
cheyang, line 418: `calculateDiffsWithExpectation` and `checkRestartBackoff` each now do their own `apiReader.Get` for the RoleInstance when the informer's `LastRestartTime` is nil. On the hot restart path this doubles direct API-server reads and creates two independent points that can diverge. Consider reading once at reconcile entry and threading the fresh copy through, or centralising the fresh read in a helper reused by both call sites.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -363,7 +517,12 @@ func shouldRecreateInstance(instance *workloadsv1alpha2.RoleInstance, pods []*v1
 	// Only trigger when Instance was previously Ready (stable state)
 	// and is NOT in the middle of any update.
 	// This avoids triggering recreate during initial creation, scaling up, or rolling/in-place updates.
-	if !wasInstanceReady(instance) || instance.Generation != instance.Status.ObservedGeneration {
+	// Exception: if LastRestartTime is set, a previous restart-policy recreation occurred,
+	// meaning the instance WAS stable before. The status update during backoff may have set
+	// Ready=False (because pods are Failed), so wasInstanceReady() returns false. We bypass
```
cheyang: `shouldRecreateInstance` now bypasses the `wasInstanceReady` check whenever `LastRestartTime != nil`. Because `LastRestartTime` is a durable status field, once set the bypass applies forever — including cases where the instance never actually became Ready between restarts (e.g. crash-loop from the very first pod-Ready flip). Consider clearing `LastRestartTime` on a stable-Ready observation (Gemini raised the same shape of concern for the exponential-decay reset case) so this bypass is bounded by 'instance was stable at some recent point'.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -318,6 +355,120 @@ func (c *realControl) shouldRecreateInstanceGuarded(ctx context.Context, instanc
...
+	if delay == 0 {
+		return 0
+	}
+
+	elapsed := time.Since(lastRestartTime.Time)
+	remaining := time.Duration(delay)*time.Second - elapsed
+	if remaining > 0 {
+		klog.Infof("Restart backoff: instance %s waiting %v (restartCount=%d, delay=%ds)",
+			klog.KObj(instance), remaining.Round(time.Second), restartCount, delay)
+		return remaining
```
NoobDream2568, line 462: Nit: If delay is large (e.g., 10h), could it cause the actual wait to be off by minutes? Given the current use cases, high-precision timing may not be necessary, but it might be worth confirming whether this drift actually occurs, since controller-runtime RateLimitingQueue isn't designed for long-duration delays.

pkg/reconciler/roleinstance/sync/instance_scale.go
cheyang: The int64 switch fixed the int32 case, but the cap is still applied *after* the multiply, so it doesn't protect against int64 overflow itself:

```go
delay := int64(baseDelaySeconds) * (int64(1) << restartCount)
if maxDelaySeconds > 0 && delay > int64(maxDelaySeconds) { ... }
```

With `base=30`, `30 * 2^59` already exceeds `math.MaxInt64` and wraps negative; at `restartCount >= 64` the shift yields 0. In both cases `delay > maxDelaySeconds` is false, the cap is skipped, and `calculateRestartDelay` returns a garbage/zero value. `checkRestartBackoff` then computes a non-positive `remaining` and backoff is silently disabled — i.e. we fall back to furious restarts exactly when a workload has been crashlooping the longest. With the default `maxDelay=600` (~10 min/restart) this is reachable overnight (~59 restarts).

Suggest short-circuiting on `restartCount` *before* shifting:

```go
if maxDelaySeconds > 0 && (restartCount >= 31 || int64(baseDelaySeconds)<<restartCount > int64(maxDelaySeconds)) {
    return maxDelaySeconds
}
if restartCount >= 31 { // unbounded case
    return 0x7FFFFFFF
}
```

The unit case labelled `"overflow boundary (27)"` doesn't actually hit the boundary — the real one is ~59. Please add a case for it.

pkg/reconciler/roleinstance/instance_status.go
```diff
@@ -412,8 +425,25 @@ func (r *realStatusUpdater) updateStatus(ctx context.Context, instance *workload
...
+		// but the current reconcile read a lower value from the stale cache.
+		// We must never roll back RestartCount or LastRestartTime.
+		liveRestartCount := clone.Status.RestartCount
+		liveLastRestartTime := clone.Status.LastRestartTime
 		clone.Status = *newStatus
 		clone.Status.Conditions = append(clone.Status.Conditions, liveCustomConditions...)
+		// Preserve live restart tracking if it's ahead of our precomputed
+		// newStatus. This prevents a stale-cache reconcile from clobbering
+		// a RestartCount that was already incremented by a prior reconcile.
+		if liveRestartCount > newStatus.RestartCount {
```
cheyang: `updateRestartTracking` resets `RestartCount` to 0 (then 1) after a long stable period, but this monotonic guard unconditionally keeps the larger live value:

```go
if liveRestartCount > newStatus.RestartCount {
    clone.Status.RestartCount = liveRestartCount
}
```

So if the count reached e.g. 5, the instance was stable, then crashes again: `updateRestartTracking` produces `newStatus.RestartCount = 1`, but `liveRestartCount (5) > 1` restores it to 5. The reset never takes effect and backoff jumps straight back to the cap instead of restarting from the base.

The `LastRestartTime` guard just below uses a different, correct condition (only preserve live when it is *newer*). Suggest treating the pair as one timestamp-versioned tuple: only preserve the live count when the live timestamp is newer than `newStatus`; if `newStatus.LastRestartTime` is newer, allow the smaller count through.

The `should reset RestartCount after stable period` envtest only exercises `1 -> 1`, where `1 > 1` is false, so it can't catch this. Please add a `5 -> 1` case.

test/envtest/testcase/restart_policy/restart_policy_test.go
```diff
@@ -806,4 +806,297 @@ var _ = Describe("RestartPolicy Controller Integration", func() {
...
+			Consistently(func() bool {
+				podList := listRolePods(rbgName, roleName)
+				active := getActivePods(podList)
+				for _, p := range active {
+					if p.UID == secondUID {
+						return true // Original pod still exists — backoff working
+					}
+				}
+				return false
+			}, backoffDelay-time.Second, interval).Should(BeTrue(),
```
cheyang: `[It] should delay second recreation due to backoff` fails in CI (14 passed / 1 failed). The second crash is recreated without honoring the delay window, which points at the backoff decision reading state that hasn't been persisted/synced yet. This needs to be green before merge — it is the core behavior the PR adds.

pkg/reconciler/roleinstance/sync/instance_scale.go
cheyang:
```go
if instance.Status.LastRestartTime == nil {
    // Also check fresh from API server below
}
```

This branch is a no-op and is failing `staticcheck` (SA9003). Please remove it.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -67,12 +74,23 @@ type expectationDiff struct {
...
 }
 
 func (c *realControl) calculateDiffsWithExpectation(ctx context.Context, updateInstance *workloadsv1alpha2.RoleInstance,
 	currentRevision, updateRevision *apps.ControllerRevision,
 	revisions []*apps.ControllerRevision, pods []*v1.Pod, inactivePods []*v1.Pod) (*expectationDiff, error) {
 
+	// Check restart backoff BEFORE any pod processing (including inactive pod cleanup).
+	// If the delay from the last restart hasn't elapsed, skip all scaling and requeue.
+	// This preserves crashed pods during the backoff window for log observability.
+	if requeue := c.checkRestartBackoff(ctx, updateInstance, pods, inactivePods); requeue > 0 {
```
cheyang: `calculateDiffsWithExpectation` is at gocyclo 33 (limit 30) and fails lint. The inlined fresh-read/backoff handling is most of the added branching — extracting the "read fresh restart tracking from API server" block (which is duplicated with `checkRestartBackoff`) into a helper would fix the lint and remove the duplication at the same time.

api/workloads/v1alpha2/roleinstance_types.go
```diff
@@ -36,6 +36,15 @@ type RoleInstanceSpec struct {
 
 	// ReadinessGates is an optional list of PodReadinessGates for the whole RoleInstance.
 	ReadinessGates []RoleInstanceReadinessGate `json:"readinessGates,omitempty"`
+
+	// BaseDelaySeconds is the base delay between restart attempts (seconds).
+	// Subsequent attempts use exponential backoff: delay = min(base * 2^(round-1), maxDelaySeconds).
+	// +optional
+	BaseDelaySeconds *int32 `json:"baseDelaySeconds,omitempty"`
```
cheyang: Unlike `LeaderWorkerPattern`/`CustomComponentsPattern`, the `RoleInstanceSpec` fields have no validation:

```go
// +optional
BaseDelaySeconds *int32 `json:"baseDelaySeconds,omitempty"`
// +optional
MaxDelaySeconds *int32 `json:"maxDelaySeconds,omitempty"`
```

A directly-created RoleInstance with a negative `baseDelaySeconds` flows into `calculateRestartDelay`, yields a negative duration, and `checkRestartBackoff` then returns "no wait" — backoff is bypassed. Please add `+kubebuilder:validation:Minimum=0` here to match the parent patterns, and consider a CEL rule for `maxDelaySeconds >= baseDelaySeconds`.

cheyang:
@gemini-code-assist /review

cheyang (commented):
Re-ran the verification harness against `ad7cd462` after the `RestartPolicyConfig` refactor. Summary: **B4 fixed ✅; B1 partially fixed ⚠️; B2 and B5 still reproduce ❌.** Details inline.

Full harness + reproductions (incl. the B2 instrumented trace) are on my fork branch `cheyang/rbg@verify/pr394-restart-backoff` under `docs/verification/pr394-restart-backoff/`; `scripts/re-verify.sh <ref>` reproduces all of it against any commit.

api/workloads/v1alpha2/rolebasedgroup_types.go
```diff
@@ -159,6 +159,29 @@ const (
...
+	// +optional
+	Type RestartPolicyType `json:"type,omitempty"`
+
+	// BaseDelaySeconds is the base delay between restart attempts (seconds).
+	// Subsequent attempts use exponential backoff: delay = min(base * 2^restartCount, maxDelaySeconds).
+	// The first recreation after a crash is immediate (no backoff) because LastRestartTime is nil.
+	// Default is 30.
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	BaseDelaySeconds *int32 `json:"baseDelaySeconds,omitempty"`
```
cheyang, line 177: ✅ **B4 (negative-delay validation) resolved.** Because `RoleInstanceSpec.RestartPolicy` now uses this shared `RestartPolicyConfig`, the `Minimum=0` here applies to RoleInstance too — the apiserver now rejects a negative `baseDelaySeconds` on RoleInstance, not just RBG:
```
spec.restartPolicy.baseDelaySeconds in body should be greater than or equal to 0
```
The old validation asymmetry is gone. (Minor: `checkRestartBackoff` still doesn't clamp a negative value in code — it returns `0s` = no backoff — but the CRD gate makes that path unreachable via the API. Optional to add a defensive clamp.)

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -318,6 +348,145 @@ func (c *realControl) shouldRecreateInstanceGuarded(ctx context.Context, instanc
...
+// Following client-go's workqueue.ItemExponentialFailureRateLimiter convention.
+// When maxDelaySeconds is 0, no cap is applied (unbounded backoff).
+func calculateRestartDelay(baseDelaySeconds, maxDelaySeconds, restartCount int32) int32 {
+	if baseDelaySeconds == 0 {
+		return 0
+	}
+	// Guard against shift overflow: int64(1) << 63 wraps to negative.
+	// With default maxDelay=600, the max-cap triggers at restartCount≈5, so
+	// this branch is only reachable when maxDelay is 0 (unbounded backoff).
+	if restartCount >= 62 {
```
cheyang: ⚠️ **B1 (int64 overflow) is only partially fixed.** This guard closes `rc≥62`, but the premise in the comment above (“with maxDelay>0 the cap triggers first, so this branch is only reachable when maxDelay==0”) doesn't hold: the cap check `delay > maxDelay` needs `delay` to still be **positive**, but `int64(base) << rc` wraps negative once `base·2^rc ≥ 2^63`. For the default `base=30` that happens at `rc=59` — three counts **below** this guard:
```
calculateRestartDelay(30, 600, 59) = 0   // want 600
calculateRestartDelay(30, 600, 60) = 0
calculateRestartDelay(30, 600, 61) = 0
calculateRestartDelay(30, 600, 62) = 600 // guard kicks in here
calculateRestartDelay(10,   0, 60) = 0   // uncapped path too
```
A `0` return makes `checkRestartBackoff` treat it as “no backoff”, so a crashloop that reaches ~59 restarts silently loses backoff. The guard only covers `base≤3`; any `base≥4` has a live window. Suggest capping on `restartCount` (or testing the multiply for overflow) **before** the shift, e.g.:
```go
if maxDelaySeconds > 0 && (restartCount >= 31 || int64(baseDelaySeconds)<<restartCount > int64(maxDelaySeconds)) {
    return maxDelaySeconds
}
```

pkg/reconciler/roleinstance/instance_status.go
```diff
@@ -412,8 +425,37 @@ func (r *realStatusUpdater) updateStatus(ctx context.Context, instance *workload
...
+		// deletion phase) may have already persisted a higher RestartCount,
+		// but the current reconcile read a lower value from the stale cache.
+		// We must never roll back RestartCount or LastRestartTime UNLESS a
+		// reset was explicitly triggered by updateRestartTracking.
+		liveRestartCount := clone.Status.RestartCount
+		liveLastRestartTime := clone.Status.LastRestartTime
+		// Detect if a reset occurred during this reconcile. updateRestartTracking
+		// resets RestartCount to 0 then 1 after a stable period. If newStatus
+		// has a lower count than the original instance, it's a reset.
+		isReset := newStatus.RestartCount < instance.Status.RestartCount
```
cheyang: ❌ **B2 (stable-period reset) still reproduces.** This `isReset` guard fires correctly for the reconcile that performs the reset — the counter does reset `5→1` and persist. But a stale-cache reconcile ~6ms later clobbers it back. Instrumented envtest trace (seed `RestartCount=5`, `LastRestartTime` 11min ago, then crash):
```
updateRestartTracking: count 5 -> 1                                     # reset works
updateStatus: newStatus=1 instance=5 liveClone=5 isReset=true  -> final=1  # persisted OK
updateStatus: newStatus=5 instance=5 liveClone=1 isReset=false -> final=5  # <-- clobbered
updateStatus: newStatus=5 instance=5 liveClone=5 isReset=false -> final=5  # stuck at 5
```
In the 2nd reconcile `newStatus.RestartCount` is still `5` (computed from a lagging informer) while the freshly-read `clone` (API) is already `1`. Then `isReset := 5 < 5 => false`, and the next line `clone.Status = *newStatus` copies the stale `5` over the fresh `1`; the `liveRestartCount > newStatus` rescue below is also false (`1 > 5`). So the reset isn't durable — any reconcile firing inside the informer-sync window re-persists the pre-reset count, and `RestartCount` stays `5` for the whole window (120/120 polls). `isReset` only guards the rollback direction (stale-low over fresh-high); it doesn't protect a fresh-low reset from a stale-high `newStatus`. A robust fix likely needs to base the reset decision on the fresh API value and only overwrite `RestartCount` when it actually changed this reconcile, rather than blindly `clone.Status = *newStatus`.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -509,6 +688,32 @@ func setRestartingCondition(instance *workloadsv1alpha2.RoleInstance) {
...
+		maxDelay := instance.Spec.GetMaxDelaySeconds()
+		stableThreshold := time.Duration(maxDelay) * 2 * time.Second
+		if stableThreshold < 10*time.Minute {
+			stableThreshold = 10 * time.Minute
+		}
+		if time.Since(instance.Status.LastRestartTime.Time) > stableThreshold {
+			instance.Status.RestartCount = 0
+		}
+	}
+	instance.Status.RestartCount++
```
cheyang, line 719: ❌ **B5 (first backoff = 2×base) unchanged.** `RestartCount` is incremented to ≥1 here **before** the first backoff check runs, so the smallest realized delay is `calculateRestartDelay(base,max,1) = 2×base`; the documented “first retry after `base`” never occurs at runtime. Either use `1 << (restartCount-1)` in `calculateRestartDelay`, or update the doc/PR table to say `2×base`.

cheyang (changes requested):
Thanks for the `RestartPolicyConfig` refactor — it's the right shape and it cleanly resolves the validation issue (B4). Requesting changes on the three that still reproduce against `ad7cd462` (verified with the harness; details in the inline comments):

**Blocking**
- **B1 — overflow guard is at the wrong threshold.** `restartCount >= 62` is above the actual overflow point: for `base=30`, `int64(base)<<rc` wraps negative at `rc=59`, so `calculateRestartDelay(30,600,{59,60,61})=0` → backoff silently disabled. Cap on `restartCount` (or test the multiply) *before* the shift.
- **B2 — stable-period reset is not durable.** The new `isReset` guard works for the reconcile that resets `5→1`, but a stale-informer reconcile ~6ms later re-persists the old count via `clone.Status = *newStatus`, so `RestartCount` never returns to 1 (reproduced: stuck at 5 for the full window). The fix needs to key off the fresh API value and only overwrite `RestartCount` when it actually changed this reconcile.

**Should fix (or document)**
- **B5 — first realized backoff is `2×base`, not `base`.** `RestartCount` is incremented before the first check, so the documented "first retry after base" never happens. Use `1<<(rc-1)` or correct the docs.

**Resolved** ✅
- **B4** — RoleInstance now rejects negative `baseDelaySeconds` via the shared `Minimum=0`.

Reproduction harness + the B2 instrumented trace are on `cheyang/rbg@verify/pr394-restart-backoff` (`docs/verification/pr394-restart-backoff/`); `scripts/re-verify.sh <ref>` re-runs everything and prints a per-finding verdict, so it's easy to confirm once these are addressed. Happy to re-verify after the next push.

cheyang:
@gemini-code-assist /review

cheyang:
Thanks for pushing this through another round — the two blockers from the last review are both cleared now. Lint is green (the SA9003 empty-branch warning and the gocyclo=33 complexity are gone), and the envtest backoff suite passes, including the RestartCount-reset-after-stable-period case that was flaky before.

The exponential-backoff logic still reads correctly: int64 arithmetic with the shift-overflow guard (restartCount >= 62) and the int32 cap, and the first recreation being immediate when LastRestartTime is nil now matches the updated doc comment. Scoping the delay propagation to explicitly-set (non-nil) values in constructRoleInstanceSetApplyConfiguration is a clean fix — it keeps the default 30/600 from perturbing the spec hash and blocking recreation. The Minimum=0 validation on the RoleInstanceSpec delay fields and the reset-detection guard in updateStatus also address the earlier concerns. E2e coverage for the feature is in place under test/e2e/testcase/v1alpha2/restart_policy_stability.go.

The only thing still red is e2e-test / e2e-test-manifest, and it isn't your code: both jobs fail-fast on the unrelated "component ordering (startAfter)" case because the runner can't pull registry-ap-southeast-1-vpc.ack.aliyuncs.com/acs/pause:3.5 (dial tcp ... i/o timeout, then ImagePullBackOff). The backoff e2e cases never get a chance to run. That's a CI/registry infra hiccup rather than a defect here, so it just needs an e2e rerun once the registry is reachable. I've flagged it to the maintainer to re-trigger the job — from a code standpoint this looks ready.

cheyang:
Thanks for the quick turnaround on `245884dd` — I re-ran the verification harness
against it. **3 of the 4 open issues are resolved (B1, B5, B4b); B2 still
reproduces.**

### ✅ B1 (int64 overflow) — fixed
Checking the overflow *before* the shift is exactly right. The old guard only
covered `rc≥62`, but `base·2^rc` wrapped negative below that; the new
`rc>=62 || base > 1<<(62-rc)` closes the whole window. Verified:
```
calculateRestartDelay(30, 600, 59) = 600   # was 0
calculateRestartDelay(30, 600, 60) = 600   # was 0
calculateRestartDelay(30, 600, 61) = 600   # was 0
calculateRestartDelay(10,   0, 60) = 2147483647   # uncapped saturates, was 0
```

### ✅ B5 (first backoff = base, not 2×base) — fixed
`base·2^(restartCount-1)` with the `restartCount<=0 => 0` guard makes the first
realized backoff equal `baseDelaySeconds` (30s), matching the docs. I flipped my
canary into a contract test asserting `first delay == base`, and it's green on
`245884dd`.

### ✅ B4b (negative delay via API) — still fixed
RoleInstance continues to reject negative `baseDelaySeconds` through the shared
`RestartPolicyConfig` `Minimum=0`. (B4a note: `checkRestartBackoff` still returns
`0s` for a negative base rather than clamping — pure defense-in-depth now that the
CRD gate makes it unreachable via the API; fine to leave, or add a clamp.)

### ❌ B2 (stable-period reset) — still reproduces
The `isReset` → timestamp-based rewrite fixed the round-2 symptom (a stale-*low*
count clobbering a fresh-*high* one), but the reset-to-1 is still not durable.
Instrumented envtest trace (seed `RestartCount=5`, `LastRestartTime` 11 min ago,
then crash):
```
[22] newStatus.RC=1 informer.RC=5 clone.RC=5 changed=true  -> final.RC=1   # reset persisted, API=1
[24] newStatus.RC=5 informer.RC=5 clone.RC=1 changed=true  -> final.RC=5   # CLOBBER
[26] newStatus.RC=1 informer.RC=1 clone.RC=5 changed=false -> final.RC=5   # preserve-live re-cements 5
```
Line 24 is the clobber: a racing reconcile carries `RestartCount=5` (a stale-high
value — `syncRestartTrackingFromAPI` only ever pulls the count *upward*,
`if fresh.RC > instance.RC`) together with a *fresh* `LastRestartTime` (re-stamped by
`updateRestartTracking`, which didn't reset because its synced timestamp was already
recent). `restartTrackingChanged` only inspects the timestamp, so it concludes "trust
newStatus" and writes 5 over the API's 1. Then at line 26 a reconcile that *does* have
the correct `newStatus.RC=1` takes the `changed=false` "preserve the live value"
branch — but the live value it reads back is the clobbered 5, so it re-cements it.
Net: `RestartCount` stays 5 for the whole window (120/120 polls).

The underlying tension is that "reset 5→1" contradicts two preserve-the-maximum
mechanisms that are still in place: (1) `syncRestartTrackingFromAPI` is monotonic-up
only, so it structurally cannot carry a reset downward; and (2) the `updateStatus`
fallback preserves the (cache-backed) live count. A fresh timestamp is being used as a
proxy for "this count is authoritative," but the count and the timestamp can come from
different sources within a reconcile.

Suggestion: treat `(RestartCount, LastRestartTime)` as one timestamp-versioned pair —
**newest timestamp wins, not highest count wins** — in both `syncRestartTrackingFromAPI`
(adopt the fresh count *and* time together when the fresh timestamp is newer, even if
the count is lower) and `updateStatus` (only overwrite RestartCount when the value
paired with the newer timestamp changed; never restore a count whose paired timestamp
is older than newStatus's).

---
Verification harness (tests + scripts + this trace) lives on my fork branch
`cheyang/rbg@verify/pr394-restart-backoff` under
`docs/verification/pr394-restart-backoff/`; `scripts/re-verify.sh` reproduces all of
the above against any ref (`results/b2-clobber-round3-trace.log` for the B2 trace).

pkg/reconciler/roleinstance/instance_status.go
```diff
@@ -412,8 +425,36 @@ func (r *realStatusUpdater) updateStatus(ctx context.Context, instance *workload
...
+		// reconcile intentionally modified restart tracking (via updateRestartTracking).
+		liveRestartCount := clone.Status.RestartCount
+		liveLastRestartTime := clone.Status.LastRestartTime
+		// Detect whether updateRestartTracking ran this reconcile by checking if
+		// newStatus.LastRestartTime differs from the informer's value. When
+		// updateRestartTracking runs, it sets LastRestartTime = metav1.Now(), which
+		// always differs from the informer's (stale or not) value. When it didn't
+		// run, newStatus.LastRestartTime equals the informer's value (initialized
+		// from instance.Status at reconcile start).
+		restartTrackingChanged := newStatus.LastRestartTime != nil &&
```
cheyang, line 441: **B2 (stable-period reset still clobbered).** This timestamp check treats a fresh `LastRestartTime` as proof that `newStatus.RestartCount` is authoritative — but within a reconcile the count and the timestamp can come from different sources. A racing reconcile carries a stale-high count (5) together with a freshly re-stamped `LastRestartTime`, so `restartTrackingChanged` is `true` and the stale `5` is written over the reset `1`. Conversely, when `newStatus.RestartCount` *is* the correct reset value but `changed=false`, the `else` branch below preserves the (cache-backed) live value — which by then is the clobbered `5`.

Instrumented envtest trace (seed count=5, LastRestartTime 11 min ago, then crash) shows the count stuck at 5 for the whole window:
```
newStatus.RC=1 informer.RC=5 clone.RC=5 changed=true  -> final.RC=1   # reset persisted
newStatus.RC=5 informer.RC=5 clone.RC=1 changed=true  -> final.RC=5   # CLOBBER
newStatus.RC=1 informer.RC=1 clone.RC=5 changed=false -> final.RC=5   # preserve-live re-cements 5
```
Consider deciding reset-vs-preserve from the `(RestartCount, LastRestartTime)` pair — newest timestamp wins — rather than from the timestamp difference alone.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -318,6 +348,147 @@ func (c *realControl) shouldRecreateInstanceGuarded(ctx context.Context, instanc
...
+// fields (e.g., checkRestartBackoff needs the Restarting condition).
+func (c *realControl) syncRestartTrackingFromAPI(ctx context.Context, instance *workloadsv1alpha2.RoleInstance) *workloadsv1alpha2.RoleInstance {
+	if c.apiReader == nil {
+		return nil
+	}
+	fresh := &workloadsv1alpha2.RoleInstance{}
+	if err := c.apiReader.Get(ctx, client.ObjectKeyFromObject(instance), fresh); err != nil {
+		return nil
+	}
+	if fresh.Status.RestartCount > instance.Status.RestartCount {
```
cheyang: **Related to B2.** This pull is monotonic-up only, so it can never carry a reset (5→1) downward — a reconcile seeded from a stale-high read keeps `5` and (via `updateRestartTracking`) re-stamps it with a fresh timestamp, which then defeats the `restartTrackingChanged` guard in `updateStatus`. Between this and that guard, the stable-period reset is not durable.

If you treat the fresh `(RestartCount, LastRestartTime)` as one versioned pair here — adopt both the fresh count and time when the fresh timestamp is newer, even if the count is lower — the reset can survive.

cheyang:
Verified the B2 fix — it works. I re-ran the harness against `422b9394`; the
stable-period reset now survives (seed `RestartCount=5`, `LastRestartTime` 11 min
ago, crash → count resets to 1 and stays 1 for the whole window, where it was
previously stuck at 5).

The two-part fix matches the failure mode exactly:
- `syncRestartTrackingFromAPI` now adopts `(RestartCount, LastRestartTime)` as a
  versioned pair — newest timestamp wins — so a reset is carried downward instead of
  being blocked by the old monotonic-up guard.
- `updateStatus` now requires the fresh timestamp to differ from *both* the informer
  and the live value, so a `syncRestartTrackingFromAPI` pass-through is no longer
  mistaken for a real `updateRestartTracking` and can't re-clobber the reset.

Status of the four issues I raised: **B1, B5, B2, B4b all resolved.** The only
leftover is B4a — `checkRestartBackoff` still returns `0s` for a negative
`baseDelaySeconds` rather than clamping — but that's pure defense-in-depth now that
B4b rejects negatives at the CRD, so it's unreachable through the API. Fine to leave
as-is, or add a one-line clamp if you want belt-and-suspenders.

Nothing else in the delta stands out. Thanks for working through all of these.

(One micro-note, non-blocking: `syncRestartTrackingFromAPI` compares timestamps with
`After`, and `metav1.Time` is second-granularity — so a hypothetical reset landing in
the *same second* as the previously-known `LastRestartTime` wouldn't be adopted by the
pair branch. In practice the reset threshold is ≥10 min, so the timestamps are always
well separated; noting it only for completeness.)

---
Verification harness (tests + scripts + traces) lives on my fork branch
`cheyang/rbg@verify/pr394-restart-backoff` under
`docs/verification/pr394-restart-backoff/`; `scripts/re-verify.sh` reproduces the
current all-green-except-B4a state against any ref.

cheyang (approved):
LGTM — approving.

Re-verified at head d0db8af0: **B1, B5, B2, B4b all fixed**, and re-run of my harness reproduces the all-green state. The only leftover is B4a (checkRestartBackoff returns 0s for a negative baseDelaySeconds instead of clamping), but that is pure defense-in-depth now that B4b rejects negatives at the CRD, so it is unreachable via the API — non-blocking, fine to leave or add a one-line clamp later.

This supersedes my earlier changes-requested (2026-07-17), which predated the B2 fix. CI is green across build/lint/unit/envtest/e2e. Thanks for working through all of these.

Verification harness (tests + scripts + traces) lives on my fork branch cheyang/rbg@verify/pr394-restart-backoff under docs/verification/pr394-restart-backoff/.

api/workloads/v1alpha2/rolebasedgroup_types.go
cheyang: Confirmed fixed — re-verified at head d0db8af0: the shared RestartPolicyConfig Minimum=0 now applies to RoleInstance too, and the envtest check shows the apiserver rejects a negative baseDelaySeconds on RoleInstance. Resolving.

pkg/reconciler/roleinstance/sync/instance_scale.go
cheyang: Confirmed fixed — re-verified at head d0db8af0: the delay formula is now base·2^(restartCount-1), so calc(base,max,1)=base (first realized delay = base, not 2×base). Unit contract test passes. Resolving.

pkg/reconciler/roleinstance/instance_status.go
cheyang: Confirmed fixed — re-verified at head d0db8af0: syncRestartTrackingFromAPI now adopts (RestartCount, LastRestartTime) as a versioned pair (newest timestamp wins) and updateStatus distinguishes a real update from a pass-through, so the stable-period reset 5→1 persists for the full window instead of being clobbered. Envtest contract test passes. Resolving.

PR #395, opened 2026-07-14
feat: add admission webhook for rbg

NoobDream2568:
## Ⅰ. Motivation

This PR introduces a validation mechanism through an admission validating webhook to improve user-facing feedback.

It covers two invalid user operations: creating a RoleBasedGroup with a name that is not a DNS label, and manually changing role replicas when the role is managed by ScalingAdapter.

## Ⅱ. Modifications

- Add RoleBasedGroup create webhook validation to require `metadata.name` to be a DNS label.
- Reject RoleBasedGroup names containing `.` with a clear validation message.
- Add RoleBasedGroup update webhook validation for role replicas changes.
- Query the corresponding ScalingAdapter named `<rbgName>-<roleName>` when `scalingAdapter.enable=true`.
- Reject the update if the requested role replicas value differs from `ScalingAdapter.spec.replicas`.
- Keep existing rolling update validation unchanged.
- Exclude webhook helper structs from DeepCopy generation.
- Add unit tests for allowed and rejected validation scenarios.
- Add end-to-end tests covering the validating webhook:
  - Reject RoleBasedGroup creation when the name contains `.`.
  - Reject role replicas updates while ScalingAdapter is enabled, but allow non-replicas updates such as adding a new environment variable.

## Ⅲ. Does this pull request fix one issue?

NONE.

## Ⅳ. List the added test cases (unit test/integration test) if any, please explain if no tests are needed.

Added unit tests for RoleBasedGroup admission validation:

- Allow RoleBasedGroup creation with a DNS label name.
- Reject RoleBasedGroup creation when `metadata.name` contains `.`.
- Reject manual role replicas changes when ScalingAdapter is enabled and has a different `spec.replicas`.
- Allow replicas changes when the ScalingAdapter has the same replicas value.
- Allow replicas changes when ScalingAdapter is disabled.
- Allow updates when role replicas are unchanged.
- Allow updates before the corresponding ScalingAdapter is created.
- Allow updates when the corresponding ScalingAdapter has no `spec.replicas` set.

Added end-to-end tests against the real API server and validating webhook:

- Reject RoleBasedGroup creation when `metadata.name` contains `.`.
- Reject role replicas updates when `scalingAdapter.enable=true` and the requested replicas differ from the corresponding `ScalingAdapter.spec.replicas`.
- Allow non-replicas updates (e.g. adding an environment variable) on the same RoleBasedGroup while ScalingAdapter is enabled.

## Ⅴ. Describe how to verify it

- Create a RoleBasedGroup with `scalingAdapter.enable=true`.
- Set the corresponding ScalingAdapter with `spec.replicas=1`.
- Update RoleBasedGroup role replicas to `3`.
- The update was correctly rejected by the validating webhook.

## VI. Special notes for reviews

The name validation applies on RoleBasedGroup creation. The replicas validation only applies when the role replicas value is changed, `scalingAdapter.enable=true`, the corresponding ScalingAdapter exists, `ScalingAdapter.spec.replicas` is set, and the requested role replicas value differs from `ScalingAdapter.spec.replicas`.


api/workloads/v1alpha2/rolebasedgroup_admission.go
```diff
@@ -0,0 +1,92 @@
...
+// +kubebuilder:object:generate=false
+type RoleBasedGroupValidator struct {
+	Client client.Client
+}
+
+var _ admission.CustomValidator = &RoleBasedGroupValidator{}
+
+// ValidateCreate validates a RoleBasedGroup on creation.
+func (v *RoleBasedGroupValidator) ValidateCreate(_ context.Context, obj runtime.Object) (admission.Warnings, error) {
+	rbg, ok := obj.(*RoleBasedGroup)
```
diw-zw, line 43: Could you also include the RBG naming validation?

NoobDream2568: Sure. This is exactly what we're gonna do.

api/workloads/v1alpha2/rolebasedgroup_validation.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	if ru.Partition != nil {
+		partition, err := intstr.GetScaledValueFromIntOrPercent(ru.Partition, int(replicas), true)
+		if err != nil {
+			errs = append(errs, fmt.Errorf(
+				"spec.roles[%d].rolloutStrategy.rollingUpdate.partition: invalid value %q",
+				index, ru.Partition.String(),
+			))
+		} else if int32(partition) >= replicas {
```
diw-zw, line 89: RoleSpec.Replicas allows 0 (+kubebuilder:validation:Minimum=0) and RollingUpdate.Partition defaults to 0 (+kubebuilder:default=0). 
When a role has an explicit rolloutStrategy.rollingUpdate and is scaled to replicas: 0 (e.g. manual scale-down, or the scaling adapter driving replicas to 0), the API server has already defaulted partition=0, so the check int32(partition) >= replicas becomes 0 >= 0 and the update/create is rejected with "partition: 0 must be less than replicas 0". Since the webhook is failurePolicy=fail, this hard-blocks a configuration the CRD schema explicitly permits. With replicas == 0 there is nothing to roll out, so the partition check should not run.

cheyang, line 89: When a role has an explicit rollingUpdate and replicas is 0, partition has already been defaulted to 0, so this becomes `0 >= 0` and the request is rejected with "partition: 0 must be less than replicas 0". That's a config the CRD schema allows (Replicas Minimum=0, Partition default=0), and with failurePolicy=Fail it gets hard-blocked. There's nothing to roll out at replicas 0, so the partition check should be skipped when replicas == 0.

api/workloads/v1alpha2/rolebasedgroup_admission.go
```diff
@@ -0,0 +1,100 @@
...
+
+// Default applies default values to the incoming RoleBasedGroup.
+func (d *RoleBasedGroupDefaulter) Default(ctx context.Context, obj runtime.Object) error {
+	return nil
+}
+
+// RoleBasedGroupValidator implements admission.CustomValidator for RoleBasedGroup.
+// Wired into the manager by SetupWebhookWithManager.
+//
+// +kubebuilder:webhook:path=/validate-workloads-x-k8s-io-v1alpha2-rolebasedgroup,mutating=false,failurePolicy=fail,sideEffects=None,groups=workloads.x-k8s.io,resources=rolebasedgroups,verbs=create;update;delete,versions=v1alpha2,name=vrolebasedgroup.kb.io,admissionReviewVersions=v1
```
cheyang: The validator is registered for `delete`, but ValidateDelete is a no-op. Combined with failurePolicy=Fail, every RBG deletion now depends on the webhook server being reachable and gets blocked when it isn't, with no validation in return. Unless a delete-time check is planned, I'd drop `delete` from the verbs here and from the webhook manifests.

api/workloads/v1alpha2/rolebasedgroup_validation.go
```diff
@@ -0,0 +1,158 @@
...
+		newRole := &newRBG.Spec.Roles[i]
+		if newRole.ScalingAdapter == nil || !newRole.ScalingAdapter.Enable {
+			continue
+		}
+		oldRole, ok := oldRoles[newRole.Name]
+		if !ok || roleReplicasEqual(oldRole.Replicas, newRole.Replicas) {
+			continue
+		}
+
+		adapterName := fmt.Sprintf("%s-%s", newRBG.Name, newRole.Name)
```
cheyang: This re-derives the adapter name inline instead of reusing `scale.GenerateScalingAdapterName(rbg.Name, role.Name)`, which the controller already uses for the same object. If the naming convention ever changes, the controller will follow it but this check won't, and a mismatch just falls into the NotFound branch and silently skips validation. Reusing the shared helper keeps them in lockstep.

api/workloads/v1alpha2/rolebasedgroup_validation.go
NoobDream2568: Good catch! The partition validation is unnecessary when replicas is 0. I'll skip this check in that case. Thanks for flagging this.

api/workloads/v1alpha2/rolebasedgroup_admission.go
NoobDream2568: I'll remove both the delete validator and the unused defaulter from the webhook registration and manifests, since they're unused.

api/workloads/v1alpha2/rolebasedgroup_validation.go
NoobDream2568: OK, I didn't realize there was already such a helper actually.

api/workloads/v1alpha2/rolebasedgroup_validation.go
NoobDream2568: The same comment as mentioned above. I'll correct this.

api/workloads/v1alpha2/rolebasedgroup_validation.go
NoobDream2568: Directly importing pkg/scale from the API package would create an import cycle because pkg/scale already depends on api/workloads/v1alpha2. I moved the naming helper to the v1alpha2 API package `helper.go` and kept pkg/scale.GenerateScalingAdapterName as a wrapper, so both the webhook validation and controller code share the same underlying naming logic without introducing a cycle.

cheyang:
@gemini-code-assist /review

deploy/helm/rbgs/templates/mutating-webhook-configuration.yaml
```diff
@@ -0,0 +1,36 @@
...
+- name: mrolebasedgroup.kb.io
+  admissionReviewVersions:
+  - v1
+  sideEffects: None
+  failurePolicy: Fail
+  clientConfig:
+    service:
+      name: rbgs-webhook-service
+      namespace: {{ .Release.Namespace }}
+      path: /mutate-workloads-x-k8s-io-v1alpha2-rolebasedgroup
```
cheyang: This template deploys `MutatingWebhookConfiguration/rbgs-mutating-webhook-configuration` with `failurePolicy: Fail`, pointing at `/mutate-workloads-x-k8s-io-v1alpha2-rolebasedgroup`. Nothing in the code base registers a handler at that path: the kubebuilder marker in `rolebasedgroup_admission.go` is `mutating=false`, `SetupWebhookWithManager` in `rolebasedgroup_webhook.go` only calls `WithValidator(...)`, and `grep -R "mutate-workloads-x-k8s-io"` on the repo returns no hits.

As soon as anyone installs this chart, the API server will POST every RoleBasedGroup CREATE/UPDATE to the webhook server, which returns 404, and the `Fail` policy turns that into a permanent admission rejection. That is a regression against the PR's stated goal of "improve user-facing feedback" — it will block every rbg mutation, not just the invalid ones.

Either drop this template entirely (there is no defaulter in this PR and `deploy/kubectl/manifests.yaml` correctly omits the mutating config) or wire up a real `admission.CustomDefaulter` and register it via `WithDefaulter(...)` alongside the validator.

api/workloads/v1alpha2/rolebasedgroup_admission.go
```diff
@@ -0,0 +1,86 @@
...
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// RoleBasedGroupValidator implements admission.CustomValidator for RoleBasedGroup.
+// Wired into the manager by SetupWebhookWithManager.
+//
+// +kubebuilder:webhook:path=/validate-workloads-x-k8s-io-v1alpha2-rolebasedgroup,mutating=false,failurePolicy=fail,sideEffects=None,groups=workloads.x-k8s.io,resources=rolebasedgroups,verbs=create;update,versions=v1alpha2,name=vrolebasedgroup.kb.io,admissionReviewVersions=v1
```
cheyang, line 33: The change adds a first-class ValidatingWebhookConfiguration wired end-to-end (cert bootstrap, caBundle patcher, webhook server, admission plugin) but only exercises `ValidateCreate`/`ValidateUpdate` with fake clients in `helper_test.go`. That gives no signal about whether the webhook actually reaches the API server: cert bootstrap, caBundle patch on the VWC, TLS handshake, and `failurePolicy: Fail` semantics are all skipped.

Our convention for `feat:` PRs is that new user-facing behaviour needs a `test/e2e/` case that drives the feature through the kube API (create RBG with a `.` in the name → expect API-server rejection; enable ScalingAdapter, then `kubectl edit` role replicas → expect rejection). Existing e2e suites like `test/e2e/testcase/v1alpha2/scaling_adapter.go` are the right neighbourhood.

Happy to be talked out of this if envtest-level coverage against the real webhook manager is planned in a follow-up, but as it stands the integration path is untested.

api/workloads/v1alpha2/rolebasedgroup_webhook.go
```diff
@@ -20,10 +20,12 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
-// SetupWebhookWithManager sets up the conversion webhook for RoleBasedGroup with the Manager.
+// SetupWebhookWithManager sets up the conversion and validating webhooks for
+// RoleBasedGroup with the Manager.
 func (r *RoleBasedGroup) SetupWebhookWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewWebhookManagedBy(mgr).
 		For(r).
+		WithValidator(&RoleBasedGroupValidator{Client: mgr.GetClient()}).
```
cheyang, line 28: `SetupWebhookWithManager` wires the validator with `mgr.GetClient()`, which is the informer-cached client. `ValidateScalingAdapterReplicas` then reads `RoleBasedGroupScalingAdapter` through that cache to decide whether to reject an rbg update.

Admission is a blocking API-server hop, but the cache lags real writes by an unbounded amount (informer sync delay). Right after an operator/HPA updates `ScalingAdapter.spec.replicas`, an rbg update that lines up with the *new* replicas can still be rejected because the cached adapter shows the *old* replicas, and vice versa.

A cheap fix is to pass `mgr.GetAPIReader()` here (or store both and use the direct reader from `ValidateScalingAdapterReplicas`) so validation decisions are grounded in a live read. Copilot flagged the same thing on this line — agreed.

deploy/helm/rbgs/templates/validating-webhook-configuration.yaml
```diff
@@ -0,0 +1,37 @@
...
+      path: /validate-workloads-x-k8s-io-v1alpha2-rolebasedgroup
+  rules:
+  - apiGroups:
+    - workloads.x-k8s.io
+    apiVersions:
+    - v1alpha2
+    operations:
+    - CREATE
+    - UPDATE
+    - DELETE
```
cheyang: The rules include `DELETE`, and `deploy/kubectl/manifests.yaml` / `config/webhook/manifests.yaml` (and the kubebuilder marker `verbs=create;update`) all agree on CREATE+UPDATE only. `ValidateDelete` in `rolebasedgroup_admission.go` just returns `nil, nil`, so the extra DELETE registration adds no validation but keeps the webhook on the delete hot-path — with `failurePolicy: Fail`, a webhook outage will now also block `kubectl delete rbg`.

Worth dropping DELETE from the helm template so all three deployment surfaces line up.

api/workloads/v1alpha2/rolebasedgroup_admission.go
```diff
@@ -0,0 +1,86 @@
...
+// +kubebuilder:object:generate=false
+type RoleBasedGroupValidator struct {
+	Client client.Client
+}
+
+var _ admission.CustomValidator = &RoleBasedGroupValidator{}
```
cheyang, line 39: Even setting aside the e2e blocker above, an envtest that boots the webhook server via `SetupWebhookWithManager` and posts an `AdmissionReview` (or uses the controller-runtime `admissiontest` helpers) would catch things the fake-client unit test can't — e.g. the CustomValidator wiring, the JSON decoder, and the `ValidateDelete` no-op contract.

api/workloads/v1alpha2/rolebasedgroup_webhook.go
NoobDream2568: Our current behavior only queries the ScalingAdapter when replicas actually changes and the adapter is enabled, so as long as replica updates aren't frequent, the performance overhead of a direct read should be acceptable. I'll switch to GetAPIReader() to eliminate the cache lag issue.

cheyang (approved):
/lgtm
/approve

PR #397, opened 2026-07-15
chore(rbgs): update Helm chart and default image versions to 0.8.0-al…

cheyang:
…pha.2

fixes #XXXX

## Checklist

### DIFF
--- a/deploy/helm/rbgs/Chart.yaml
+++ b/deploy/helm/rbgs/Chart.yaml
@@ -15,10 +15,10 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.8.0-alpha.1
+version: 0.8.0-alpha.2
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
-appVersion: "0.8.0-cea2a47"
+appVersion: "0.8.0-69fe55d"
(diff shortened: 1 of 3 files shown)


cheyang:
Straightforward version bump. I checked all four files plus the rest of the tree: the controller image tag moves from `v0.8.0-cea2a47` to `v0.8.0-69fe55d` consistently across kustomization, values.yaml (both the controller and crdUpgrade images), and the rendered manifests.yaml, and no stale `cea2a47` references are left anywhere. The chart `version` is correctly bumped to `0.8.0-alpha.2` and `appVersion` matches the new image tag. No leftover `0.8.0-alpha.1`. Nothing blocking here.

deploy/helm/rbgs/Chart.yaml
```diff
@@ -15,10 +15,10 @@ type: application
...
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.8.0-alpha.1
+version: 0.8.0-alpha.2
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
-appVersion: "0.8.0-cea2a47"
+appVersion: "0.8.0-69fe55d"
```
cheyang, line 24: Minor, and pre-existing rather than introduced here: appVersion is `0.8.0-69fe55d` (no `v`) while every image tag uses `v0.8.0-69fe55d`. The manager/crd-upgrade templates render the tag as `.Values.image.tag | default .Chart.AppVersion`, so this only matters if someone clears `image.tag` in values, in which case the fallback would resolve to a `v`-less tag that likely isn't published. Not a blocker for this bump, just noting it in case you want to normalize the two at some point.

PR #400, opened 2026-07-16
doc: add doc best-practice/configuring-coordinated-policy

NoobDream2568:

NONE.

## Checklist

### DIFF
--- a/doc/best-practice/en/07-configuring-coordinated-policy.md
+++ b/doc/best-practice/en/07-configuring-coordinated-policy.md
@@ -1,0 +1,387 @@
+# Configuring Coordinated Policy
+
+## Overview
+
+In multi-role inference services (such as PD-disaggregated architecture), each role operates independently, but their deployment and upgrades need to remain coordinated. If Prefill and Decode create or update at different rates, it may lead to version mismatches, resource contention, or service unavailability.
+
+`CoordinatedPolicy` is a standalone CRD that defines cross-role coordination strategies. It supports two coordination scenarios:
+
++ **Coordinated scaling**: Controls creation progress synchronization across multiple roles during initial deployment, ensuring replica counts grow proportionally and progressively
++ **Coordinated upgrade**: Controls update progress synchronization across multiple roles during rolling updates, ensuring consistent update progress across roles
+
+`CoordinatedPolicy` **applies to the same-named RBG in the current namespace**, associating roles by referencing role names in the RBG. It is decoupled from the RBG lifecycle — creating or deleting a CoordinatedPolicy does not affect the RBG itself.
+
+## Prerequisites
+
++ Kubernetes cluster version >= 1.24
++ RBG Controller installed (see [Installation Guide](https://github.com/sgl-project/rbg))
++ A RoleBasedGroup with multiple roles deployed
+
+---
+
+## Background: Why Multi-Role Coordination Is Needed
+
+In PD-disaggregated architecture, Prefill (prompt encoding) and Decode (token generation) are two independent roles, each with its own replica count and update strategy. When they are deployed or upgraded simultaneously, the following problems arise without coordination:
+
+### Initial Deployment Scenario
+
+#### Uncoordinated Initial Deployment
+
+| Role | Deployment Progress | Description |
+| --- | --- | --- |
+| Prefill | `0 → 6` (quickly created) | All instances ready |
+| Decode | `0 → 2` (only half created) | Instances clearly insufficient |
+
+Problems:
+
+- Prefill instances are ready but Decode is insufficient, unable to form an effective inference pipeline
+- Prefill instances created earlier sit idle, wasting GPU resources
+
+### Upgrade Scenario
+
+#### Uncoordinated Upgrade
+
+| Role | Update Progress | Description |
+| --- | --- | --- |
+| Prefill | Fully updated to new version | All new instances online |
+| Decode | Only half updated | Old instances still serving |
+
+Problems:
+
+- New-version Prefill and old-version Decode may have protocol incompatibilities
+- Prefill has many unavailable instances while Decode is still serving normally
+
+`CoordinatedPolicy` uses the `maxSkew` parameter to control the progress difference between roles, ensuring they advance in lockstep.
+
+---
+
+## CoordinatedPolicy Basic Structure
+
+`CoordinatedPolicy` defines a set of coordination rules via `spec.policies`. Each rule specifies a group of roles and the corresponding strategy:
+
+```yaml
+apiVersion: workloads.x-k8s.io/v1alpha2
+kind: CoordinatedPolicy
+metadata:
+  name: <policy-name>             # Same name as the RBG to bind
+  namespace: <ns-name>            # Same namespace as the RBG to bind
+spec:
+  policies:
+    - name: <rule-name>           # Rule name, unique within the CoordinatedPolicy
+      roles:                       # List of role names to coordinate
+        - <role-1>
+        - <role-2>
+      strategy:                    # Coordination strategy
+        scaling:                   # Coordinated scaling (optional)
+          ...
+        rollingUpdate:             # Coordinated upgrade (optional)
+          ...
+```
+
+### Parameter Description
+
+| Parameter | Type | Required | Description |
+| --- | --- | --- | --- |
+| `spec.policies[].name` | string | Yes | Rule name, unique within the CoordinatedPolicy |
+| `spec.policies[].roles` | []string | Yes | List of role names to coordinate, must match role names in the RBG |
+| `spec.policies[].strategy.scaling` | object | No | Coordinated scaling strategy (only affects initial deployment) |
+| `spec.policies[].strategy.rollingUpdate` | object | No | Coordinated upgrade strategy |
+
+> **Note**: `scaling` and `rollingUpdate` can be configured together or used independently. `scaling` only takes effect during the initial deployment phase and does not affect subsequent autoscaling behavior after deployment completes. A single CoordinatedPolicy can contain multiple rules, each applying to different role groups.
+>
+
+---
+
+## Scenario 1: Coordinated Scaling (Initial Deployment)
+
+During initial deployment of a PD-disaggregated architecture, Prefill and Decode replicas need to be created proportionally and progressively. Without coordination, one role may finish creating quickly while the other is still queuing, causing the ready role to sit idle. The `scaling` strategy of `CoordinatedPolicy` uses `maxSkew` to control the creation progress difference between roles and `progression` to control the creation pace, ensuring the multi-role replica ratio always meets expectations.
+
+> **Note**: Coordinated scaling only affects the replica creation rate during the initial deployment phase and does not affect runtime autoscaling behavior. Once initial deployment completes, this strategy no longer takes effect.
+>
+
+### Configuration Example
+
+```yaml
+apiVersion: workloads.x-k8s.io/v1alpha2
+kind: RoleBasedGroup
+metadata:
+  name: pd-inference
+spec:
+  roles:
+    - name: prefill
+      replicas: 2
+      standalonePattern:
+        template:
+          spec:
+            containers:
+              - name: engine
+                image: lmsysorg/sglang:v0.5.9
+                ports:
+                  - containerPort: 8000
+                resources:
+                  requests:
+                    nvidia.com/gpu: "1"
+                  limits:
+                    nvidia.com/gpu: "1"
+
+    - name: decode
+      replicas: 4
+      standalonePattern:
+        template:
+          spec:
+            containers:
+              - name: engine
+                image: lmsysorg/sglang:v0.5.9
+                ports:
+                  - containerPort: 8000
+                resources:
+                  requests:
+                    nvidia.com/gpu: "1"
+                  limits:
+                    nvidia.com/gpu: "1"
+
+---
+# Coordinated scaling strategy (initial deployment)
+apiVersion: workloads.x-k8s.io/v1alpha2
+kind: CoordinatedPolicy
+metadata:
+  name: pd-inference
+spec:
+  policies:
+    - name: prefill-decode-scaling
+      roles:
+        - prefill
+        - decode
+      strategy:
+        scaling:
+          maxSkew: "10%"
+          progression: OrderScheduled
+```
+
+### Parameter Description (Coordinated Scaling)
+
+| Parameter | Type | Required | Default | Description |
+| --- | --- | --- | --- | --- |
+| `strategy.scaling.maxSkew` | intOrString | No | `"100%"` | Allowed deployment progress deviation between roles. Can be an absolute value (e.g., `2`) or a percentage (e.g., `"10%"`) |
+| `strategy.scaling.progression` | string | No | `OrderScheduled` | Creation pace: `OrderScheduled` or `OrderReady` |
+
+### How maxSkew Works
+
+`maxSkew` controls the creation progress difference between roles. The Controller calculates each role's deployment progress (created replicas / target replicas) and ensures the progress difference between any two roles does not exceed `maxSkew`.
+
+Assume Prefill has a target replica count of 5 and Decode has a target replica count of 10, with `maxSkew: "10%"`:
+
+#### Coordinated Scaling Process (maxSkew: 10%)
+
+| Batch | Prefill Progress | Decode Progress | Progress Difference | Within maxSkew |
+| --- | --- | --- | --- | --- |
+| 1 | `0 → 1` (20%) | `0 → 1` (10%) | 10% | ≤ 10%, allowed to continue |
+| 2 | `1 → 2` (40%) | `1 → 3` (30%) | 10% | ≤ 10%, allowed to continue |
+
+And so on — the two roles always maintain proportional progressive creation. If Prefill creates too fast and the progress difference exceeds 10%, the Controller pauses Prefill creation and waits for Decode to catch up.
+
+#### Choosing maxSkew
+
+| maxSkew Value | Behavior | Applicable Scenario |
+| --- | --- | --- |
+| `"1%"` | Nearly synchronous creation, minimal progress difference | Scenarios requiring extremely high inter-role consistency |
+| `"10%"` | Allows small progress differences | Most production environments |
+| `"50%"` | Allows larger progress differences | Prioritizes deployment speed over consistency |
+| `"100%"` | No progress difference restriction (default) | No coordination needed, roles create independently |
+
+### progression Mode Comparison
+
+| Mode | Behavior | Applicable Scenario |
+| --- | --- | --- |
+| `OrderScheduled` | A Pod is considered complete once scheduled (node assigned), then the next batch proceeds | Sufficient GPU resources, prioritizing deployment speed |
+| `OrderReady` | All Pods must be fully Ready before proceeding to the next batch | Ensures new instances are available before continuing, safer |
+
+> **Note**: In `OrderReady` mode, if a Pod takes too long to become ready due to model loading time, it blocks subsequent role creation. It is recommended to configure appropriate initial delays with `startupProbe` / `readinessProbe`.
+>
+
+---
+
+## Scenario 2: Coordinated Upgrade
+
+When multiple roles of an RBG trigger rolling updates simultaneously (e.g., updating the inference engine image version), the `rollingUpdate` strategy of `CoordinatedPolicy` ensures consistent update progress across roles, avoiding version mismatches.
+
+### Configuration Example (Coordinated Upgrade)
+
+```yaml
+apiVersion: workloads.x-k8s.io/v1alpha2
+kind: RoleBasedGroup
+metadata:
+  name: pd-inference
+spec:
+  roles:
+    - name: prefill
+      replicas: 7
+      rolloutStrategy:
+        type: RollingUpdate
+        rollingUpdate:
+          maxUnavailable: 1
+      standalonePattern:
+        template:
+          spec:
+            containers:
+              - name: engine
+                image: lmsysorg/sglang:v0.6.0   # New version
+
+    - name: decode
+      replicas: 3
+      rolloutStrategy:
+        type: RollingUpdate
+        rollingUpdate:
+          maxUnavailable: 1
+      standalonePattern:
+        template:
+          spec:
+            containers:
+              - name: engine
+                image: lmsysorg/sglang:v0.6.0   # New version
+
+---
+# Coordinated upgrade strategy
+apiVersion: workloads.x-k8s.io/v1alpha2
+kind: CoordinatedPolicy
+metadata:
+  name: pd-inference
+spec:
+  policies:
+    - name: prefill-decode-rollout
+      roles:
+        - prefill
+        - decode
+      strategy:
+        rollingUpdate:
+          maxSkew: "10%"
+          maxUnavailable: "10%"
+```
+
+### Parameter Description (Coordinated Upgrade)
+
+| Parameter | Type | Required | Default | Description |
+| --- | --- | --- | --- | --- |
+| `strategy.rollingUpdate.maxSkew` | intOrString | No | `"100%"` | Allowed update progress deviation between roles |
+| `strategy.rollingUpdate.maxUnavailable` | intOrString | No | - | Maximum proportion of unavailable instances across all roles during coordinated update |
+| `strategy.rollingUpdate.partition` | intOrString | No | - | Global partition value for coordinated update (shared by all roles) |
+
+### How Coordinated Upgrade Works
+
+Assume Prefill has 7 instances and Decode has 3 instances, with `maxSkew: "10%"`:
+
+1. **Calculate update progress**: Each role's update progress = updated instances / total instances
+2. **Control progress difference**: The Controller ensures the update progress difference between Prefill and Decode does not exceed 10%
+3. **Coordinate update order**: If Prefill updates too fast, the Controller pauses Prefill's update and waits for Decode to catch up
+
+#### Coordinated Upgrade Process (maxSkew: 10%)
+
+| Role | Instances | Update Status | Update Progress |
+| --- | --- | --- | --- |
+| Prefill | 7 | 4/7 updated | 57% |
+| Decode | 3 | 1/3 updated | 33% |
+
+Progress difference: `57% - 33% = 24%`, exceeds `10%`. The Controller pauses Prefill's update and prioritizes Decode's update.
+
+### Choosing maxSkew (Coordinated Upgrade)
+
+| maxSkew Value | Behavior | Applicable Scenario |
+| --- | --- | --- |
+| `"1%"` | Nearly synchronous update, minimal progress difference | Scenarios requiring extremely high version consistency |
+| `"10%"` | Allows small progress differences | Most production environments |
+| `"50%"` | Allows larger progress differences | Prioritizes update speed over consistency |
+| `"100%"` | No progress difference restriction (default) | No coordination needed, roles update independently |
+
+> **Note**: The smaller the `maxSkew`, the slower the update, but the higher the version consistency across roles. Choose an appropriate value based on business requirements and cluster size.
+>
+
+---
+
+## Scenario 3: Configuring Both Scaling and Upgrade Coordination
+
+`scaling` and `rollingUpdate` can be configured simultaneously in the same CoordinatedPolicy without conflict:
+
+```yaml
+apiVersion: workloads.x-k8s.io/v1alpha2
+kind: CoordinatedPolicy
+metadata:
+  name: pd-full-policy
+spec:
+  policies:
+    - name: prefill-decode
+      roles:
+        - prefill
+        - decode
+      strategy:
+        # Coordinated scaling (initial deployment)
+        scaling:
+          maxSkew: "10%"
+          progression: OrderScheduled
+        # Coordinated upgrade
+        rollingUpdate:
+          maxSkew: "10%"
+          maxUnavailable: "10%"
+```
+
+> **Note**: The `maxSkew` for scaling and upgrade can be set to different values. For example, allow a larger deviation (`"50%"`) during initial deployment for speed, while requiring strict synchronization (`"1%"`) during upgrades to ensure version consistency.
+>
+
+---
+
+## Multi-Rule Configuration
+
+A single CoordinatedPolicy can contain multiple rules, each applying to different role groups. This is useful in complex inference services, such as a Router + Prefill + Decode three-role architecture:
+
+```yaml
+apiVersion: workloads.x-k8s.io/v1alpha2
+kind: CoordinatedPolicy
+metadata:
+  name: multi-rule-policy
+spec:
+  policies:
+    # Rule 1: Prefill and Decode coordinated deployment
+    - name: engine-scaling
+      roles:
+        - prefill
+        - decode
+      strategy:
+        scaling:
+          maxSkew: "10%"
+          progression: OrderReady
+
+    # Rule 2: Prefill and Decode coordinated upgrade
+    - name: engine-rollout
+      roles:
+        - prefill
+        - decode
+      strategy:
+        rollingUpdate:
+          maxSkew: "5%"
+```
+
+---
+
+## Verification
+
+```bash
+# Check CoordinatedPolicy status
+kubectl get cpolicy
+
+# View CoordinatedPolicy details
+kubectl get cpolicy <policy-name> -o yaml
+
+# View RBG role replica counts (deployment progress coordination)
+kubectl get rbg <rbg-name> -o jsonpath='{range .spec.roles[*]}{.name}{"="}{.replicas}{"\n"}{end}'
+
+# View role Pod update status (upgrade coordination)
+kubectl get pods -l rbg.workloads.x-k8s.io/group-name=<rbg-name> -o wide
+```
+
+## Related Documents
+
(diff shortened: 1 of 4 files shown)


cheyang:
@gemini-code-assist /review

doc/best-practice/zh/07-configuring-coordinated-policy.md
```diff
@@ -0,0 +1,379 @@
...
+
+## 背景：为什么需要多角色协作
+在 PD 分离架构中，Prefill（Prompt 编码）和 Decode（Token 生成）是两个独立角色，各自拥有独立的副本数和更新策略。当它们同时进行部署或升级时，如果不加以协调，会出现以下问题：
+
+### 首次部署场景
+
+**不协调的首次部署**
+
+| 角色 | 部署进度 | 说明 |
+| --- | --- | --- |
```
cheyang, line 28: The `markdown-check` job is red on this PR with 18 findings across both files. Main categories:

- `MD024/no-duplicate-heading`: `参数说明`, `配置示例`, `maxSkew 的选择`, `预期行为`, `验证`, `清理`, `步骤 1：创建 CoordinatedPolicy 和多角色...` are repeated. Consider disambiguating (e.g. `参数说明（scaling）`, `参数说明（rollingUpdate）`) or nesting them under distinct H3 sections so their anchors differ.
- `MD036/no-emphasis-as-heading`: bold-only paragraphs like `**不协调的首次部署**` (line 28), `**不协调的升级**` (line 42), `**协作伸缩过程（maxSkew: 10%）**` (line 178), `**协作升级过程（maxSkew: 10%）**` (line 281) should be real headings (`####`) or inline table captions.
- `MD001/heading-increment`: line 81 jumps from `## CoordinatedPolicy 基础结构` straight to `#### 参数说明`. Use `###` to match the same section in later scenes.
- `MD042/no-empty-links`: the four `[…](#)` entries under `## 相关文档`.

Making these lint clean should unblock the failing job.

doc/best-practice/zh/07-configuring-coordinated-policy-guide.md
```diff
@@ -0,0 +1,247 @@
...
+    - name: prefill
+      replicas: 4
+      rolloutStrategy:
+        type: RollingUpdate
+      standalonePattern:
+        template:
+          spec:
+            containers:
+              - name: engine
+                    image: alpine:3.23.5
```
cheyang: In operation two the container spec is misaligned:

```yaml
            containers:
              - name: engine
                    image: alpine:3.23.5   # over-indented, YAML parse error
                imagePullPolicy: IfNotPresent
                command: ["sleep", "3600"]
```

Copy-pasting into `kubectl apply -f -` fails with `mapping values are not allowed in this context`. Please align `image:` with `imagePullPolicy:` (16 spaces).

doc/best-practice/zh/07-configuring-coordinated-policy.md
```diff
@@ -0,0 +1,379 @@
...
+
+# 查看各角色 Pod 更新状态（升级协作）
+kubectl get pods -l rbg.workloads.x-k8s.io/group-name=<rbg-name> -o wide
+
+# 查看 CoordinatedPolicy 的 Conditions
+kubectl get cpolicy <policy-name> -o jsonpath='{.status.conditions}'
+```
+
+## 相关文档
++ [使用 RBG 部署推理服务](#)
```
cheyang: All four entries under `## 相关文档` currently link to `#`, which is why `MD042/no-empty-links` fires and the Gemini bot flagged the same. Either replace with the real paths (e.g. `./05-port-allocation-and-service-discovery.md`) or drop the section until those docs are written.

doc/best-practice/zh/07-configuring-coordinated-policy.md
```diff
@@ -0,0 +1,379 @@
+# 配置多角色协作策略
```
cheyang, line 1: The existing best-practice pairs (`02-using-role-templates*.md`, `05-port-allocation-and-service-discovery*.md`, `10-deploy-mooncake-store-with-rbg*.md`) all ship a matching file under `doc/best-practice/en/`. Please add `doc/best-practice/en/07-configuring-coordinated-policy.md` and `…-guide.md`, or note in the PR description that an EN version will follow so this doesn't fall through.

doc/best-practice/zh/07-configuring-coordinated-policy.md
```diff
@@ -0,0 +1,379 @@
...
+| 参数 | 类型 | 是否必填 | 默认值 | 说明 |
+| --- | --- | --- | --- | --- |
+| `strategy.scaling.maxSkew` | intOrString | 否 | `"100%"` | 允许的角色间部署进度偏差。可以是绝对值（如 `2`）或百分比（如 `"10%"`） |
+| `strategy.scaling.progression` | string | 否 | `OrderScheduled` | 创建节奏：`OrderScheduled` 或 `OrderReady` |
+
+
+### maxSkew 的工作原理
+`maxSkew` 控制各角色的创建进度差异。Controller 计算每个角色的部署进度（已创建的副本数 / 目标副本数），确保任意两个角色的进度差异不超过 `maxSkew`。
+
+假设 Prefill 目标副本数为 6，Decode 目标副本数为 8，`maxSkew: "10%"`：
```
cheyang: The narrative says Prefill target = 6 and Decode target = 8, then the table shows Prefill `0→1 (17%)` / Decode `0→1 (13%)`, `1→2 (33%)` / `1→3 (38%)`. `1/8 = 12.5%` and `3/8 = 37.5%` — the numbers are close but rounded inconsistently, and picking targets like 5 and 10 (or 6 and 6) would make the illustration land cleaner.

doc/best-practice/zh/07-configuring-coordinated-policy-guide.md
cheyang: 这段 YAML 缩进有误：`image:` 比同级的 `- name: engine` 多缩进了 4 个空格，`kubectl apply` 会直接失败：`error converting YAML to JSON: yaml: line 17: mapping values are not allowed in this context`。用户照抄操作二会卡在第一步。把 `image:` 对齐到与下面的 `imagePullPolicy:` / `command:` 相同的缩进即可（同一块里 decode 角色的缩进是正确的，可对照）。

doc/best-practice/zh/07-configuring-coordinated-policy.md
```diff
@@ -0,0 +1,379 @@
...
+kubectl get cpolicy <policy-name> -o yaml
+
+# 查看 RBG 各角色副本数（部署进度协作）
+kubectl get rbg <rbg-name> -o jsonpath='{range .spec.roles[*]}{.name}{"="}{.replicas}{"\n"}{end}'
+
+# 查看各角色 Pod 更新状态（升级协作）
+kubectl get pods -l rbg.workloads.x-k8s.io/group-name=<rbg-name> -o wide
+
+# 查看 CoordinatedPolicy 的 Conditions
+kubectl get cpolicy <policy-name> -o jsonpath='{.status.conditions}'
```
cheyang: 实测 CoordinatedPolicy 的 `.status.conditions` 一直为空（`kubectl get cpolicy` 也只显示 NAME/AGE），这条验证命令不会有任何输出，容易让用户误以为出错。建议去掉这一步，或改为查看 RBG 各角色副本数 / Pod 更新状态来验证协作效果。

doc/best-practice/zh/07-configuring-coordinated-policy.md
```diff
@@ -0,0 +1,379 @@
...
+```yaml
+apiVersion: workloads.x-k8s.io/v1alpha2
+kind: RoleBasedGroup
+metadata:
+  name: pd-inference
+spec:
+  roles:
+    - name: prefill
+      replicas: 2
+      scalingAdapter:
```
cheyang: 协作伸缩并不需要 `scalingAdapter`（操作文档里未配置也能正常分批创建）。在“协作伸缩”示例中带上 `scalingAdapter.enable: true` 容易让读者误以为是前提条件，建议去掉以免混淆。

cheyang:
@gemini-code-assist /review

cheyang:
Re-review of the coordinated-policy docs. All four blockers from the previous round are addressed:

- markdown-check CI is green now (the 18 lint errors are gone).
- The YAML indentation in the operations guide is correct across both language versions.
- The empty related-doc links were replaced with plain-text entries plus a TODO note, which is what cleared the lint failures.
- The English counterparts are in place — both the concept doc and the operations guide now have en/ versions.

I cross-checked the documented API against `api/workloads/v1alpha2/coordinatedpolicy_types.go` and it lines up: `apiVersion: workloads.x-k8s.io/v1alpha2`, `kind: CoordinatedPolicy`, the `cpolicy` short name, and the field set (`strategy.scaling.maxSkew` / `progression` with the `OrderScheduled`/`OrderReady` enum, and `strategy.rollingUpdate.maxSkew` / `maxUnavailable` / `partition`, all IntOrString). The worked maxSkew examples are internally consistent.

Content looks good to merge. Left one non-blocking note about the Related Documents section — a few of the linked docs already exist and could be wired up now instead of deferred. Handing off to a maintainer for the lgtm/approved labels.

doc/best-practice/zh/07-configuring-coordinated-policy.md
```diff
@@ -0,0 +1,374 @@
...
+# 查看 RBG 各角色副本数（部署进度协作）
+kubectl get rbg <rbg-name> -o jsonpath='{range .spec.roles[*]}{.name}{"="}{.replicas}{"\n"}{end}'
+
+# 查看各角色 Pod 更新状态（升级协作）
+kubectl get pods -l rbg.workloads.x-k8s.io/group-name=<rbg-name> -o wide
+```
+
+## 相关文档
+
+<!-- TODO: 以下文档尚未创建，待文档完成后统一添加链接 -->
```
cheyang: The TODO here says these docs haven't been created yet, but a few of them already live alongside this file: `01-deploy-inference-service.md` (使用 RBG 部署推理服务), `03-configuring-rolling-updates.md` (配置滚动更新策略), and `04-configuring-inplace-update-and-scheduling-policies.md` (原地升级与原地调度). You could wire up relative links for those now and keep the TODO only for the autoscaling doc that genuinely doesn't exist yet. Non-blocking.

doc/best-practice/en/07-configuring-coordinated-policy.md
```diff
@@ -0,0 +1,387 @@
...
+# View RBG role replica counts (deployment progress coordination)
+kubectl get rbg <rbg-name> -o jsonpath='{range .spec.roles[*]}{.name}{"="}{.replicas}{"\n"}{end}'
+
+# View role Pod update status (upgrade coordination)
+kubectl get pods -l rbg.workloads.x-k8s.io/group-name=<rbg-name> -o wide
+```
+
+## Related Documents
+
+<!-- TODO: The following documents have not been created yet. Links will be added after documentation is complete. -->
```
cheyang: Same as the zh version: `01-deploy-inference-service.md` and `03-configuring-rolling-updates.md` already exist under `doc/best-practice/en/`, so "Deploying Inference Services with RBG" and "Configuring Rolling Update Strategies" can be linked now. The in-place update and autoscaling entries don't have EN counterparts yet, so the TODO still makes sense for those. Non-blocking.

cheyang:
@gemini-code-assist /review

doc/best-practice/en/07-configuring-coordinated-policy.md
cheyang: The links here now resolve to real files (verified 01-deploy-inference-service.md and 03-configuring-rolling-updates.md both exist), so this is good to go. Small nit: this TODO still says the listed docs "have not been created yet," but the first two are already linked. The zh version just dropped the TODO line entirely. Either trim this comment so it only covers the two items that genuinely lack an EN page (Autoscaling, In-Place Update), or drop it to match zh. Not blocking.

doc/best-practice/zh/07-configuring-coordinated-policy.md
```diff
@@ -0,0 +1,372 @@
...
+      strategy:
+        scaling:
+          maxSkew: "10%"
+          progression: OrderScheduled
+```
+
+### 参数说明（协作伸缩）
+| 参数 | 类型 | 是否必填 | 默认值 | 说明 |
+| --- | --- | --- | --- | --- |
+| `strategy.scaling.maxSkew` | intOrString | 否 | `"100%"` | 允许的角色间部署进度偏差。可以是绝对值（如 `2`）或百分比（如 `"10%"`） |
```
cheyang: The table offers `2` as a valid `scaling.maxSkew`, but `parsePercentage` in `pkg/coordination/coordinationscaling/scaler.go:253-257` requires a trailing `%`. The error travels up through `CalculateScalingForAllCoordination` to `Reconcile`, which returns it, so reconciliation stops before any role is created.

The consequence is bigger than coordinated scaling being skipped. On a live cluster (3-node, v1.36.1-aliyun.1, controller v0.8.0-47cfe17) applying a two-role RBG with `maxSkew: 2` gives:

```
RBG ready : False
pods      : 0
Warning  FailedCalculateScaling  rolebasedgroup/c  failed to create coordination
  scaler: failed to parse maxSkew: percentage string must end with '%'
```

Changing only that value to `"100%"` gives `READY=True` and all four pods Running. Nothing warns the user at apply time, because the CRD is `x-kubernetes-int-or-string`: `kubectl apply` succeeds and the only clue is that event.

Worth flagging that the Go doc comment on `coordinatedpolicy_types.go:86-87` and the generated CRD description make the same absolute-value promise, so the doc is faithfully describing an intent the code does not implement. Changing `scaling.maxSkew` to accept `intstr.GetScaledValueFromIntOrPercent`, as `rollingUpdate.maxSkew` already does, would make this table correct as written, and that may be the better fix. If the docs are the thing to change instead, this row needs to say percentage-only and warn about the failure mode.

doc/best-practice/en/07-configuring-coordinated-policy.md
```diff
@@ -0,0 +1,387 @@
...
+        scaling:
+          maxSkew: "10%"
+          progression: OrderScheduled
+```
+
+### Parameter Description (Coordinated Scaling)
+
+| Parameter | Type | Required | Default | Description |
+| --- | --- | --- | --- | --- |
+| `strategy.scaling.maxSkew` | intOrString | No | `"100%"` | Allowed deployment progress deviation between roles. Can be an absolute value (e.g., `2`) or a percentage (e.g., `"10%"`) |
```
cheyang: Mirror of the Chinese row above. `Can be an absolute value (e.g., 2)` describes something that bricks the group, so whatever wording you land on for the zh table needs to land here too.

doc/best-practice/zh/07-configuring-coordinated-policy.md
```diff
@@ -0,0 +1,372 @@
...
+
+> **说明**：`scaling` 和 `rollingUpdate` 可以同时配置，也可以单独使用。`scaling` 仅在首次部署阶段生效，部署完成后不影响后续的弹性伸缩行为。一个 CoordinatedPolicy 可以包含多条规则，分别作用于不同的角色组。
+>
+
+---
+
+## 场景一：协作伸缩（首次部署）
+在 PD 分离架构首次部署时，Prefill 和 Decode 的副本需要按比例渐进式创建。如果不加协调，一个角色可能快速创建完毕而另一个角色还在排队，导致先就绪的角色空闲等待。`CoordinatedPolicy` 的 `scaling` 策略通过 `maxSkew` 控制各角色的创建进度差异，通过 `progression` 控制创建的节奏，确保多角色副本数比例始终符合预期。
+
+> **说明**：协作伸缩仅影响首次部署阶段的副本创建速度，不影响线上运行时的弹性伸缩行为。首次部署完成后，该策略不再生效。
```
cheyang: There is no first-deployment gate anywhere in the code path. `CalculateScalingForAllCoordination` runs on every reconcile, and `CalculateTargetReplicas` only compares each role's `CurrentReplicas / DesiredReplicas` ratio, with nothing distinguishing an initial rollout from a later one. The resulting targets overwrite role replicas at `rolebasedgroup_controller.go:532`.

Measured from a steady state of prefill 5/5 and decode 10/10, raising the targets to 20/40 with `maxSkew: 10%` yields `{prefill: 7, decode: 14}` on the first step, not 20/40.

This matters most for anyone pairing the policy with an HPA or a ScalingAdapter: scale-out gets throttled, and the note here tells them it cannot be the cause. The same limitation appears at lines 82 and 86, and at en:87/90/99.

doc/best-practice/zh/07-configuring-coordinated-policy-guide.md
```diff
@@ -0,0 +1,247 @@
...
+                imagePullPolicy: IfNotPresent
+                command: ["sleep", "3600"]
+EOF
+```
+
+### 预期行为（初始部署）
+
+- 创建 4 个 prefill Pod 和 4 个 decode Pod
+- RBG 状态 Ready 后，可以同时触发两个角色的更新
+- `rollingUpdate.maxSkew: 1`：两个角色之间的更新进度差异最多为 1 个实例
```
cheyang: `rolebasedgroup_controller.go:1487` converts the value with `intstr.GetScaledValueFromIntOrPercent(&maxSkew, 100, true)`. The base is the literal `100`, so an integer is read as a percentage: `1` is identical to `"1%"`, and the surrounding function comment says as much (`s = allowed percentage max-skew`).

A table over four values confirms that `"1"`/`"1%"`, `"10"`/`"10%"` and `"100"`/`"100%"` produce identical bounds. In the guide's own 4-replica scenario, `maxSkew: 1` collapses to a plus-or-minus-zero lockstep window, so the YAML at line 134 does not demonstrate the behaviour the text describes, and `1`, `1%` and `25%` all give the same next step there.

Using a percentage in the example (`maxSkew: "25%"`) would avoid the ambiguity entirely. Line 196 and en:184/198 need the same correction.

doc/best-practice/zh/07-configuring-coordinated-policy.md
```diff
@@ -0,0 +1,372 @@
...
+`maxSkew` 控制各角色的创建进度差异。Controller 计算每个角色的部署进度（已创建的副本数 / 目标副本数），确保任意两个角色的进度差异不超过 `maxSkew`。
+
+假设 Prefill 目标副本数为 5，Decode 目标副本数为 10，`maxSkew: "10%"`：
+
+#### 协作伸缩过程（maxSkew: 10%）
+
+| 批次 | Prefill 进度 | Decode 进度 | 进度差异 | 是否符合 maxSkew |
+| --- | --- | --- | --- | --- |
+| 1 | `0 → 1`（20%） | `0 → 1`（10%） | 10% | ≤ 10%，允许继续 |
+| 2 | `1 → 2`（40%） | `1 → 3`（30%） | 10% | ≤ 10%，允许继续 |
```
cheyang: Batch 1 is right. Batch 2 is not: starting from `{prefill: 1/5, decode: 1/10}` with `maxSkew: 10%`, the algorithm returns `{prefill: 1, decode: 2}`, not `{prefill: 2, decode: 3}`.

The reason is worth showing in the table, because it is the interesting part of the design. `minProgress` is taken from the slowest role that has not finished (decode at 0.1), so `maxAllowedProgress` is 0.2. Prefill is already at `1/5 = 0.2`, hits `scaler.go:150`, and holds at its current replica count while decode catches up. A three-row walkthrough, where both start, prefill pauses, then prefill resumes, would describe the real behaviour and explain the mechanism at the same time.

en:176-179 carries the same numbers.

doc/best-practice/zh/07-configuring-coordinated-policy-guide.md
```diff
@@ -0,0 +1,247 @@
...
+> coordinated-scaling-demo-decode-0    1/1     Running   0          19s
+> coordinated-scaling-demo-decode-1    1/1     Running   0          7s
+> coordinated-scaling-demo-prefill-0   1/1     Running   0          18s
+> coordinated-scaling-demo-prefill-1   1/1     Running   0          18s
+> coordinated-scaling-demo-prefill-2   0/1     Running   0          7s
+> coordinated-scaling-demo-prefill-3   0/1     Running   0          7s
+```
+
+**预期输出：**
+- Pod 不是单一角色一次性全部创建完成，而是 prefill 和 decode 按接近进度逐步创建，每次创建 1 decode 和 2 prefill
```
cheyang, line 103: For the guide's scenario (prefill 4, decode 2, `maxSkew: "25%"`, `progression: OrderReady`) the batches are `{0,0}` to `{1,1}` to `{2,1}` to `{3,2}` to `{4,2}`. Prefill advances by one every batch; decode advances every other batch. No batch creates two prefill.

The sample `kubectl get pods` output at lines 94-99 shows prefill-2 and prefill-3 both at `0/1` simultaneously, which does not fit that cadence either. It would be worth recapturing that block from a real run rather than adjusting it by hand, since status lag can make the real output look different again.

en:105 and en:94-100 mirror this.

doc/best-practice/zh/07-configuring-coordinated-policy.md
```diff
@@ -0,0 +1,372 @@
...
+        scaling:
+          maxSkew: "10%"
+          progression: OrderScheduled
+```
+
+### 参数说明（协作伸缩）
+| 参数 | 类型 | 是否必填 | 默认值 | 说明 |
+| --- | --- | --- | --- | --- |
+| `strategy.scaling.maxSkew` | intOrString | 否 | `"100%"` | 允许的角色间部署进度偏差。可以是绝对值（如 `2`）或百分比（如 `"10%"`） |
+| `strategy.scaling.progression` | string | 否 | `OrderScheduled` | 创建节奏：`OrderScheduled` 或 `OrderReady` |
```
cheyang: The field carries `+optional` and an Enum but no `+kubebuilder:default`, so the live CRD has no default and an omitted `progression` is `""`. `canProceedToNextBatch` switches on only `OrderScheduled` and `OrderReady` (`scaler.go:227-238`), so `""` matches no case and batch gating is skipped altogether. That is measurably different from `OrderScheduled`, which does check `ScheduledReplicas < CurrentReplicas`. `getProgressionType()` only falls back to the default when `Strategy.Scaling` is nil.

Measured: omitting the field gives `{prefill: 2, decode: 4}` where `OrderScheduled` gives `{prefill: 2, decode: 2}`.

Adding `+kubebuilder:default=OrderScheduled` to the API would make this row true as written. Otherwise the default column should say the gate is simply not applied.

doc/best-practice/zh/07-configuring-coordinated-policy.md
```diff
@@ -0,0 +1,372 @@
...
+        rollingUpdate:
+          maxSkew: "10%"
+          maxUnavailable: "10%"
+```
+
+### 参数说明（协作升级）
+| 参数 | 类型 | 是否必填 | 默认值 | 说明 |
+| --- | --- | --- | --- | --- |
+| `strategy.rollingUpdate.maxSkew` | intOrString | 否 | `"100%"` | 允许的角色间更新进度偏差 |
+| `strategy.rollingUpdate.maxUnavailable` | intOrString | 否 | - | 协作更新过程中，所有角色合计允许不可用的最大实例比例 |
```
cheyang: `rolebasedgroup_controller.go:1284-1286` assigns the value straight onto each role's strategy inside the `for _, role := range rbg.Spec.Roles` loop, and it lands in `strategyRollingUpdate[role.Name]`. So `"10%"` means 10% per role and overrides each role's own `rolloutStrategy.rollingUpdate.maxUnavailable`. The total unavailable ceiling is roughly 10% times the number of roles, not 10% overall. The e2e case at `test/e2e/testcase/v1alpha2/coordinated_policy.go:49,55` sets it per role too. Same wording at en:266.

doc/best-practice/zh/07-configuring-coordinated-policy.md
```diff
@@ -0,0 +1,372 @@
...
+| `strategy.rollingUpdate.partition` | intOrString | 否 | - | 协作更新的全局分区值（所有角色共享） |
```
cheyang: `CalculatePartitionReplicas` (`pkg/utils/utils.go:105-125`) scales the value against that role's own `replicas`, so `"50%"` becomes 4 on a 7-replica role and 2 on a 3-replica one. It is one policy with per-role arithmetic rather than one shared ordinal. It also acts as a floor only: the partition derived from `maxSkew` wins when it is larger (`controller.go:1327-1331`). Same wording at en:267.

doc/best-practice/zh/07-configuring-coordinated-policy.md
```diff
@@ -0,0 +1,372 @@
...
+假设 Prefill 目标副本数为 5，Decode 目标副本数为 10，`maxSkew: "10%"`：
+
+#### 协作伸缩过程（maxSkew: 10%）
+
+| 批次 | Prefill 进度 | Decode 进度 | 进度差异 | 是否符合 maxSkew |
+| --- | --- | --- | --- | --- |
+| 1 | `0 → 1`（20%） | `0 → 1`（10%） | 10% | ≤ 10%，允许继续 |
+| 2 | `1 → 2`（40%） | `1 → 3`（30%） | 10% | ≤ 10%，允许继续 |
+
+以此类推，两个角色始终保持按比例渐进式创建。如果 Prefill 创建过快，进度差异超过 10%，Controller 会暂停 Prefill 的创建，等待 Decode 跟上。
```
cheyang: The direction is right, but "暂停 Prefill 的创建" reads as though something is rolled back or actively suspended. What happens is quieter: `scaler.go:149-153` pins the fast role's target to its current replica count, so nothing new is created and nothing already created is deleted. Saying that explicitly removes the worry that pods might be reclaimed. Line 265 and en:181/275 phrase it the same way.

doc/best-practice/zh/07-configuring-coordinated-policy.md
```diff
@@ -0,0 +1,372 @@
+# 配置多角色协作策略
```
cheyang, line 1: This lands on 07 while main has 01-04 and 10, leaving 05 and 06 empty. #402 takes 05, so there is no collision, but if 06 is not spoken for it might be worth using it. Unrelated: the guide's link text hard-codes the ordinal (`7. 配置 CoordinatedPolicy`), which means renumbering later touches two places.

doc/best-practice/zh/07-configuring-coordinated-policy.md
```diff
@@ -0,0 +1,372 @@
...
+          maxUnavailable: "10%"
+```
+
+### 参数说明（协作升级）
+| 参数 | 类型 | 是否必填 | 默认值 | 说明 |
+| --- | --- | --- | --- | --- |
+| `strategy.rollingUpdate.maxSkew` | intOrString | 否 | `"100%"` | 允许的角色间更新进度偏差 |
+| `strategy.rollingUpdate.maxUnavailable` | intOrString | 否 | - | 协作更新过程中，所有角色合计允许不可用的最大实例比例 |
+| `strategy.rollingUpdate.partition` | intOrString | 否 | - | 协作更新的全局分区值（所有角色共享） |
+
```
cheyang, line 259: Outside this PR's scope and not documented here, so nothing to change. Noting it while it is in view: `controller.go:1449-1453` compares two `maxUnavailable` values with base 100, which can order an absolute against a percentage in a way that surprises. Only reachable when several rules both carry `rollingUpdate`, which the doc does not cover.

cheyang (changes requested):
Thanks for this. The structure is clear, the zh/en pair is genuinely in sync, every YAML example passes server-side validation, and all the relative links resolve. That is better than average for a doc of this size, so most of what follows is about claims that do not match the controller rather than about the writing.

I verified each of these against the code and, where it mattered, against a live cluster (3-node, v1.36.1-aliyun.1, controller v0.8.0-47cfe17). Reproductions and captured output are on `verify/pr400-coordinated-policy-doc` at `cheyang/rbg@660da2bc`, under `docs/verification/pr400-coordinated-policy-doc/`.

## What blocks merge

The parameter table offers an absolute `scaling.maxSkew` (`2`). Applying exactly that leaves the RoleBasedGroup with **zero pods** and `READY=False`, with the only clue in a `FailedCalculateScaling` event, because `kubectl apply` succeeds: the CRD is `x-kubernetes-int-or-string`. Changing that one value to `"100%"` brings all four pods up. Details inline at zh:158 and en:165.

Four further claims describe behaviour the controller does not have. Coordinated scaling is said to affect only the first deployment, but there is no such gate and scale-out is throttled too. `rollingUpdate.maxSkew: 1` is said to mean one instance, when it is always a percentage, so `1` equals `"1%"`. And the batch numbers in both the worked example and the guide do not match the algorithm's output.

## Worth deciding before you edit

Several of these are arguably implementation bugs rather than documentation bugs, and I would rather flag that than have you write around them:

`scaling.maxSkew`: the Go doc comment at `coordinatedpolicy_types.go:86-87` and the generated CRD description also promise absolute values, so the doc is faithfully describing an intent `parsePercentage` never implemented. Making `scaling.maxSkew` use `intstr.GetScaledValueFromIntOrPercent`, as `rollingUpdate.maxSkew` already does, would make your table correct as written.

`progression`: it has no `+kubebuilder:default`, so omitting it gives `""`, which matches no case in `canProceedToNextBatch` and skips batch gating entirely rather than behaving like `OrderScheduled`. Adding the default to the API would likewise validate the table.

Either direction is fine by me. It just changes what you write, and I am happy to re-run the harness against whichever you pick.

## Not verified

The `kubectl get pods` sample output in the guide I could only check by algorithm, not by a full live rollout, so it would be worth recapturing from a real run. I also have not confirmed how coordinated scaling interacts with `RoleBasedGroupScalingAdapter` specifically, which affects how strongly to word the HPA caveat.

doc/best-practice/zh/07-configuring-coordinated-policy.md
NoobDream2568: 暂停 -> 暂缓。

NoobDream2568: 暂停 -> 暂缓。

doc/best-practice/zh/07-configuring-coordinated-policy-guide.md
NoobDream2568: Adjust to `maxSkew: "50%"`. And it works correctly in this example.

cheyang (approved):
Re-reviewed at `cc623b20`. The blocker is fixed, and fixed better than I asked for.

I had suggested changing the `2` to `"100%"`. You did something more useful — you documented the failure mode instead of quietly routing around it:

> `strategy.scaling.maxSkew` … 仅支持百分比格式（如 `"10%"`）；不支持整数（如 `2`），设置整数会导致解析失败，所有角色的副本都不会被创建

That is exactly right, and it is the sentence a reader needs. Someone who tries an absolute value now finds out from the doc why every replica stayed at zero, instead of from a `FailedCalculateScaling` event. The default is `"100%"` in both languages.

The four behavioural claims are also corrected. The one I most wanted fixed reads well now:

> 协作伸缩对首次部署和后续扩容均生效……策略持续生效，不会在首次部署完成后失效。缩容不受该策略限制。

`markdown-check` is green (22s), so the 18 findings are gone too.

Two open threads remain and neither needs anything from you:

- **zh:28** — my markdown-check comment is stale now that the job passes. Resolving it.
- **zh:259** — that one was flagged at the time as outside this PR's scope with nothing to change; leaving it for whoever picks up the multi-rule `maxUnavailable` comparison later.

Approving.

One follow-up that belongs in the API rather than in your doc, so please don't hold this PR for it: `scaling.maxSkew`'s Go doc comment at `coordinatedpolicy_types.go:86-87` and the generated CRD description still promise absolute values, which is what sent you down this path originally. Either `parsePercentage` should grow `intstr.GetScaledValueFromIntOrPercent` like `rollingUpdate.maxSkew` already has, or those descriptions should be corrected. Same for `progression` having no `+kubebuilder:default`. Worth a separate issue — if the API changes, your table gets simpler and this paragraph can go away.

cheyang (commented):
Re-reviewed after the latest push. Thanks for the follow-ups.

The stale `<!-- TODO -->` note in the EN "Related Documents" section is gone now, which was the only thing outstanding from the last round.

I also went through the technical corrections in `07-configuring-coordinated-policy.md` (EN and zh) against the controller code, and they line up:

- `scaling.maxSkew` being percentage-only, with a bare integer failing to parse: matches `parsePercentage` in `pkg/coordination/coordinationscaling/scaler.go`, which requires the `%` suffix and errors out otherwise. The previous "can be an absolute value (e.g. `2`)" wording was misleading, so this is a good fix.
- `scaling.progression` defaulting to no batch gating when omitted: matches `canProceedToNextBatch` — an empty progression hits neither the `OrderScheduled` nor `OrderReady` branch, so nothing gates the next batch. Recommending an explicit value is the right call.
- `scaling` applying to scale-outs (not just first deployment) and leaving scale-down alone: consistent with `CalculateTargetReplicas`, which works off desired-vs-current generically and short-circuits roles already at/over desired.
- `rollingUpdate.partition` percentages scaling against each role's own replica count (4 on a 7-replica role, 2 on a 3-replica one): matches the `GetScaledValueFromIntOrPercent(..., role.Replicas, roundUp)` call in the stateful reconciler.

The rework of the maxSkew batch table (batch 2 holding Prefill at the cap while Decode catches up) is also consistent with the `minProgress + maxSkew` cap logic.

Docs-only change, so it's exempt from e2e; markdown-check and the other required checks are green. From my side this is ready to merge — flagging to the maintainer for the final call.

PR #402, opened 2026-07-17
doc: add doc best-practice/port-allocation-and-service-discovery

NoobDream2568:

fixes #XXXX

## Checklist

### DIFF
--- a/doc/best-practice/en/05-port-allocation-and-service-discovery.md
+++ b/doc/best-practice/en/05-port-allocation-and-service-discovery.md
@@ -1,0 +1,666 @@
+# Port Allocation and Service Discovery
+
+## Overview
+
+In distributed inference services, multiple roles and components need to communicate with each other. RBG provides a three-layer service discovery mechanism, covering different scenarios from simple to complex:
+
+- **Layer 1: Headless Service + DNS**
+  - Automatically creates Headless Services
+  - Each Pod has a stable DNS name
+  - Use case: Fixed ports, known topology
+- **Layer 2: Environment Variables + ConfigMap**
+  - Controller automatically injects environment variables (role name, index, leader address, etc.)
+  - Automatically generates a global ConfigMap (address and port topology for all roles)
+  - ConfigMap mounted at `/etc/rbg/config.yaml`
+  - Use case: Runtime cluster topology retrieval
+- **Layer 3: Port Allocation + Component Discovery**
+  - Dynamic port allocation (avoids port conflicts in hostNetwork scenarios)
+  - Component discovery annotation (component-discovery) to obtain addresses and ports of other components
+  - Use case: CustomComponentsPattern + hostNetwork + RDMA scenarios
+
+## Prerequisites
+
++ Kubernetes cluster version >= 1.24
++ RBG Controller installed (see [Installation Guide](https://github.com/sgl-project/rbg))
++ Port allocation feature requires enabling `--enable-port-allocator` (disabled by default)
+
+---
+
+## Layer 1: Headless Service and DNS Service Discovery
+
+### Automatically Created Headless Service
+
+RBG Controller automatically creates a Headless Service (`ClusterIP: None`) for each role, with the naming convention:
+
+```plain
+s-{rbgName}-{roleName}
+```
+
+For example, if the RBG is named `pd-inference` and the role is named `prefill`, the automatically created Service is named `s-pd-inference-prefill`.
+
+Headless Service configuration:
+
+| Property | Value | Description |
+| --- | --- | --- |
+| `clusterIP` | `None` | Headless, no virtual IP assigned |
+| `publishNotReadyAddresses` | `true` | Publishes DNS records even when Pods are not ready |
+| `selector` | `rbg.workloads.x-k8s.io/group-name: <rbgName>, rbg.workloads.x-k8s.io/role-name: <roleName>` | Automatically matches all Pods of the role |
+
+
+### Pod DNS Naming Convention
+
+Each Pod receives a stable DNS name through the Headless Service:
+
+```plain
+{rbgName}-{roleName}-{index}.{serviceName}.{namespace}.svc.cluster.local
+```
+
+Examples:
+
+```plain
+# The 0th Pod of the Prefill role
+pd-inference-prefill-0.s-pd-inference-prefill.default.svc.cluster.local
+
+# The 2nd Pod of the Decode role
+pd-inference-decode-2.s-pd-inference-decode.default.svc.cluster.local
+```
+
+### Usage
+
+Pods can directly access Pods of other roles via DNS names, without manually configuring Services or IP addresses:
+
+```yaml
+# Access the 0th Prefill instance from a Decode Pod
+containers:
+  - name: engine
+    env:
+      - name: PREFILL_ADDR
+        value: "pd-inference-prefill-0.s-pd-inference-prefill.default.svc.cluster.local:8000"
+```
+
+> **Note**: The `{index}` in the DNS name is the Pod's ordinal index (starting from 0). In StatefulSet and RoleInstanceSet modes, Pod names are stable, and DNS addresses remain unchanged after Pod recreation.
+>
+
+---
+
+## Layer 2: Environment Variables and ConfigMap
+
+### Automatically Injected Environment Variables
+
+Controller automatically injects the following environment variables into each Pod, allowing inference engines to obtain their own role information and cluster topology at runtime:
+
+#### Basic Variables (All Roles)
+| Variable | Value | Description |
+| --- | --- | --- |
+| `RBG_GROUP_NAME` | RBG name | Name of the owning RoleBasedGroup |
+| `RBG_ROLE_NAME` | Role name | Name of the role the current Pod belongs to |
+
+
+#### Stateful Roles (StatefulSet / LeaderWorkerSet / RoleInstanceSet)
+| Variable | Value | Source |
+| --- | --- | --- |
+| `RBG_ROLE_INDEX` | Pod ordinal index | Downward API (StatefulSet: `pod-index` label, RoleInstanceSet: `role-instance-index` label) |
+
+
+#### RoleInstanceSet-Specific Variables
+| Variable | Value | Source |
+| --- | --- | --- |
+| `RBG_ROLE_INSTANCE_NAME` | RoleInstance name | Downward API (`role-instance-name` label) |
+| `RBG_COMPONENT_NAME` | Component name | Downward API (`component-name` label) |
+| `RBG_COMPONENT_INDEX` | Component index | Downward API (`component-id` label) |
+
+
+#### LeaderWorkerPattern-Specific Variables
+| Variable | Value | Description |
+| --- | --- | --- |
+| `RBG_LWP_LEADER_ADDRESS` | Leader Pod FQDN | Computed value: `$(RBG_ROLE_INSTANCE_NAME)-0.{svcName}.{namespace}` |
+| `RBG_LWP_WORKER_INDEX` | Current Worker index | Downward API (`component-index` label) |
+| `RBG_LWP_GROUP_SIZE` | Total Pods in group | Downward API (`component-size` label) |
+
+
+> **Note**: Size-related environment variables (such as `RBG_LWP_GROUP_SIZE`) change during scaling, but Controller intentionally does not inject these variables in non-LWP scenarios to avoid triggering Pod recreation when replica counts change.
+>
+
+### ConfigMap Cluster Topology
+
+Controller automatically creates an RBG-level ConfigMap (with the same name as the RBG) containing address and port information for all roles. The ConfigMap is automatically mounted into every Stateful role's Pods.
+
+#### ConfigMap Structure
+```yaml
+# ConfigMap: pd-inference (key: config.yaml)
+group:
+  name: pd-inference
+  size: 2
+  roles:
+    - prefill
+    - decode
+roles:
+  prefill:
+    size: 2
+    instances:
+      - address: pd-inference-prefill-0.s-pd-inference-prefill
+        ports:
+          http: 8000
+      - address: pd-inference-prefill-1.s-pd-inference-prefill
+        ports:
+          http: 8000
+  decode:
+    size: 4
+    instances:
+      - address: pd-inference-decode-0.s-pd-inference-decode
+        ports:
+          http: 8000
+      - address: pd-inference-decode-1.s-pd-inference-decode
+        ports:
+          http: 8000
+      - address: pd-inference-decode-2.s-pd-inference-decode
+        ports:
+          http: 8000
+      - address: pd-inference-decode-3.s-pd-inference-decode
+        ports:
+          http: 8000
+```
+
+#### ConfigMap Mount Configuration
+| Property | Value | Description |
+| --- | --- | --- |
+| Volume name | `rbg-cluster-config` | Automatically injected Volume |
+| Mount path | `/etc/rbg` | Automatically mounted in all containers |
+| File name | `config.yaml` | ConfigMap key |
+| Permissions | Read-only | Pods cannot modify |
+
+
+Inference engines can directly read `/etc/rbg/config.yaml` to obtain the entire cluster's topology information:
+
+```python
+# Inference engine code example: reading cluster topology
+import yaml
+
+with open("/etc/rbg/config.yaml") as f:
+    config = yaml.safe_load(f)
+
+# Get all Prefill instance addresses
+prefill_instances = config["roles"]["prefill"]["instances"]
+for inst in prefill_instances:
+    print(f"Prefill: {inst['address']}:{inst['ports']['http']}")
+
+# Get all Decode instance addresses
+decode_instances = config["roles"]["decode"]["instances"]
+for inst in decode_instances:
+    print(f"Decode: {inst['address']}:{inst['ports']['http']}")
+```
+
+#### Port Information Source
+
+Port information in the ConfigMap comes from the `servicePorts` definition in the role Spec:
+
+```yaml
+spec:
+  roles:
+    - name: prefill
+      replicas: 2
+      servicePorts:
+        - name: http
+          port: 8000
+          protocol: TCP
+      standalonePattern:
+        template:
+          spec:
+            containers:
+              - name: engine
+                image: lmsysorg/sglang:v0.5.9
+                ports:
+                  - containerPort: 8000
+```
+
+Port names are converted to lowercase with `-` replaced by `_` (e.g., `http-api` → `http_api`). Unnamed ports use the `port{number}` format (e.g., `port8000`).
+
+> **Note**: The ConfigMap is automatically maintained by the Controller. When role replica counts change (scaling) or ServicePorts are modified, the ConfigMap updates automatically. Since the ConfigMap is mounted as a Volume, files inside Pods are automatically synced by kubelet without requiring Pod restart.
+>
+
+---
+
+## Layer 3: Port Allocation and Component Discovery
+
+### Background: Why Dynamic Port Allocation Is Needed
+
+In RDMA inference scenarios using `hostNetwork: true`, multiple Pods on the same node share the host's network namespace. If two Pods' containers listen on the same port, a port conflict occurs.
+
+```plain
+┌──────────────────────────────────────────────────────────────────┐
+│  hostNetwork Port Conflict Problem                                │
+│                                                                  │
+│  Node A                                                          │
+│  ┌──────────────────────────────────────────────────────┐        │
+│  │  Prefill Pod-0        Prefill Pod-1                  │        │
+│  │  containerPort: 8000   containerPort: 8000  ← Conflict! │     │
+│  │                                                      │        │
+│  │  Shared host network namespace → cannot bind same port│        │
+│  └──────────────────────────────────────────────────────┘        │
+│                                                                  │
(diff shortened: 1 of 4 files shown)


cheyang:
@gemini-code-assist /review

cheyang:
This is a solid start on documenting the three tiers of service discovery in RBG. Before this can move out of WIP two things need to happen: (1) the Markdown Quality Check job is red on this branch because of 10 MD024 duplicate-heading violations in the guide file — please make each `预期行为` / `验证` / `清理` heading unique or configure `siblings_only`; (2) the project convention is bilingual under `doc/best-practice/{en,zh}/`, but only the `zh/` variants are added here — please add the English counterparts. A few smaller things (leftover TODO in `相关文档`, unnecessary `hostNetwork` on the `router` sample, install-guide link pointing at the repo root, label-key inconsistency between the concept doc and the guide) are called out inline. Happy to re-review once CI is green and the English pair lands.

doc/best-practice/zh/05-port-allocation-and-service-discovery-guide.md
```diff
@@ -0,0 +1,499 @@
...
+              - name: engine
+                image: alpine:3.23.5
+                imagePullPolicy: IfNotPresent
+                command: ["sleep", "3600"]
+EOF
+```
+
+### 预期行为
+
+- 每个 Pod 自动注入 `RBG_GROUP_NAME`、`RBG_ROLE_NAME` 等环境变量
```
cheyang, line 160: markdownlint (MD024/no-duplicate-heading) reports 10 duplicate H3 headings in this guide: `预期行为` (lines 160, 321, 465), `验证` (lines 165, 224, 327, 470), and `清理` (lines 247, 354, 489). The Markdown Quality Check job is red because of this. Two options: (a) enable `siblings_only: true` in `.markdownlint.jsonc` if the repo does not already, or (b) qualify each duplicated heading so it is unique in the document, e.g. `### 验证：Headless Service`, `### 验证：ConfigMap`, `### 清理（操作一）`, etc. Option (b) is the cheapest fix and also improves TOC navigation. Please rerun `markdownlint` locally before pushing the next revision.

doc/best-practice/zh/05-port-allocation-and-service-discovery.md
```diff
@@ -0,0 +1,664 @@
+# 端口分配与服务发现
```
cheyang, line 1: Every other best-practice entry ships as a bilingual pair — `doc/best-practice/en/<n>-<slug>.md` and `doc/best-practice/zh/<n>-<slug>.md` (see 02 and 10). This PR only adds the Chinese variants. Please add the corresponding English files under `doc/best-practice/en/` so the documentation set stays consistent. If English is not ready yet that is fine to defer, but the WIP label should stay on until it lands.

doc/best-practice/zh/05-port-allocation-and-service-discovery.md
```diff
@@ -0,0 +1,664 @@
...
+
+例如，RBG 名为 `pd-inference`，角色名为 `prefill`，自动创建的 Service 名称为 `s-pd-inference-prefill`。
+
+Headless Service 的配置：
+
+| 属性 | 值 | 说明 |
+| --- | --- | --- |
+| `clusterIP` | `None` | Headless，不分配虚拟 IP |
+| `publishNotReadyAddresses` | `true` | 即使 Pod 未就绪也发布 DNS 记录 |
+| `selector` | `group-name: <rbgName>, role-name: <roleName>` | 自动匹配角色的所有 Pod |
```
cheyang: The concept doc describes the Headless Service selector as `group-name: <rbgName>, role-name: <roleName>`, and the verification snippet at line 636 uses `workloads.x-k8s.io/group-name=<rbg-name>`. The companion guide (and the other best-practice docs, e.g. `02-using-role-templates-guide.md`) consistently use the fully-qualified key `rbg.workloads.x-k8s.io/group-name`. Please make the concept doc show the full label keys, and cross-check that line 636 uses the same prefix as the guide so readers can copy/paste either doc without editing.

doc/best-practice/zh/05-port-allocation-and-service-discovery.md
```diff
@@ -0,0 +1,664 @@
...
+  - 适用：需要运行时获取集群拓扑的场景
+- **第 3 层：端口分配 + 组件发现**
+  - 动态分配端口（hostNetwork 场景避免端口冲突）
+  - 组件发现注解（component-discovery）获取其他组件的地址和端口
+  - 适用：CustomComponentsPattern + hostNetwork + RDMA 场景
+
+## 前提条件
+
++ Kubernetes 集群版本 >= 1.24
++ 已安装 RBG Controller（参考 [安装指南](https://github.com/sgl-project/rbg)）
```
cheyang, line 24: `[安装指南](https://github.com/sgl-project/rbg)` sends readers to the repo home instead of the actual install document. Please point it to the real installation guide (e.g. a relative link under `doc/getting-started/` or wherever the canonical install doc lives) so the prerequisite is actionable.

doc/best-practice/zh/05-port-allocation-and-service-discovery.md
```diff
@@ -0,0 +1,664 @@
...
+# 查看端口分配结果（PodScoped 场景）
+kubectl exec -it <pod-name> -- env | grep -E 'LEADER_GRPC_PORT|WORKER_GRPC_PORT'
+
+# 查看 RoleInstance 注解中的端口值
+kubectl get roleinstance <name> -o jsonpath='{.metadata.annotations}'
+```
+
+## 相关文档
+
+<!-- TODO: 以下文档尚未创建，待文档完成后统一添加链接 -->
```
cheyang: The trailing `<!-- TODO: 以下文档尚未创建，待文档完成后统一添加链接 -->` block ships plain-text titles with no links. Either (a) drop the section until the referenced docs exist, or (b) add relative links to the pages that do exist today so the section is not misleading. Leaving a TODO in a merged doc tends to become permanent.

doc/best-practice/zh/05-port-allocation-and-service-discovery.md
```diff
@@ -0,0 +1,664 @@
...
+| **CustomComponentsPattern 多组件通信** | 端口分配 + 组件发现 + 依赖管理 | 组件间需要发现彼此的地址和动态端口 |
+
+
+---
+
+## 验证
+
+```bash
+# 查看自动创建的 Headless Service
+kubectl get svc -l workloads.x-k8s.io/group-name=<rbg-name>
```
cheyang: 这里的 label 少了 `rbg.` 前缀。实际 label 是 `rbg.workloads.x-k8s.io/group-name`（见 `api/workloads/constants/label.go` 与 `constants.go` 中的 RBGPrefix）。实测用 `workloads.x-k8s.io/group-name` 作为选择器查不到任何 Service（返回 No resources found）。操作文档里用的是正确的 key，建议这里同步改为 `rbg.workloads.x-k8s.io/group-name`。

doc/best-practice/zh/05-port-allocation-and-service-discovery.md
```diff
@@ -0,0 +1,664 @@
...
+--port-allocate-strategy=random  # 分配策略（目前仅支持 random）
+--start-port=30000               # 端口范围起始值
+--port-range=5000                # 端口范围大小（30000~34999）
+```
+
+Helm 部署配置：
+
+```yaml
+# values.yaml
+portAllocator:
```
cheyang: 这个顶层 `portAllocator:` 在 chart v0.8.0 中已失效。values 已重组到 `controller.features.portAllocator`（见 `deploy/helm/rbgs/values.yaml` 与 `manager.yaml`），且 chart 的 `NOTES.txt` 明确警告顶层 `portAllocator.*` 会被静默忽略。用户照此配置部署，端口分配器不会开启，操作三/四会毫无报错地失败、极难排查。建议改为：

```yaml
controller:
  features:
    portAllocator:
      enabled: true
      strategy: random
      startPort: 30000
      portRange: 5000
```

cheyang:
@gemini-code-assist /review

doc/best-practice/en/05-port-allocation-and-service-discovery.md
```diff
@@ -0,0 +1,668 @@
...
+# View port allocation results (PodScoped scenario)
+kubectl exec -it <pod-name> -- env | grep -E 'LEADER_GRPC_PORT|WORKER_GRPC_PORT'
+
+# View port values in RoleInstance annotations
+kubectl get roleinstance <name> -o jsonpath='{.metadata.annotations}'
+```
+
+## Related Documents
+
+<!-- TODO: The following documents have not been created yet; links will be added after documentation is complete -->
```
cheyang: The markdown-check, the missing English counterpart, and the earlier prefix/link/hostNetwork nits are all cleared now, so this reads well end to end. One small leftover: this TODO says the related docs don't exist yet, but `01-deploy-inference-service.md` and `03-configuring-rolling-updates.md` are already sitting in this same `en/` directory. You could turn the first two bullets into real relative links (same style you use in the guide's header at line 3) and keep only the in-place update entry as a TODO. Not a blocker.

cheyang:
@gemini-code-assist /review

doc/best-practice/en/05-port-allocation-and-service-discovery.md
```diff
@@ -0,0 +1,666 @@
...
+# View port allocation results (PodScoped scenario)
+kubectl exec -it <pod-name> -- env | grep -E 'LEADER_GRPC_PORT|WORKER_GRPC_PORT'
+
+# View port values in RoleInstance annotations
+kubectl get roleinstance <name> -o jsonpath='{.metadata.annotations}'
+```
+
+## Related Documents
+
++ [Deploying Inference Services with RBG](./01-deploy-inference-service.md)
```
cheyang, line 664: This was the last thing I'd flagged and it's sorted now - the TODO placeholder is gone and both links resolve to files that exist in this directory. I also confirmed the EN list intentionally leaves "In-Place Update and In-Place Scheduling" as plain text: there's no English 04-configuring-inplace-update-and-scheduling-policies.md yet (only the zh/ side has it), so keeping it unlinked avoids a dead link rather than being an oversight. Nothing else outstanding from my side.

doc/best-practice/zh/05-port-allocation-and-service-discovery.md
```diff
@@ -0,0 +1,663 @@
...
+| --- | --- | --- | --- |
+| `name` | string | 是 | 端口逻辑名称，用于被其他组件引用 |
+| `env` | string | 是 | 注入到容器的环境变量名 |
+| `scope` | string | 否 | 作用域：`PodScoped`（默认）或 `RoleScoped` |
+| `annotationKey` | string | 否 | 同时注入到 Pod 注解的 key（可选） |
+
+
+### 引用其他组件的端口
+
+通过 `references` 字段，一个组件可以获取同角色内其他组件的已分配端口：
```
cheyang, line 315: The sentence and the table below it place no constraint on scope, and the complete example further down allocates the leader's port as `"scope": "RoleScoped"`. Read together, they suggest `references: prefill.leader.leader-grpc` will resolve. It will not.

`pkg/port-allocator/manager.go:125-143` builds only a PodScoped key (`FormatPodScopedPortKey(refPodName, portName)`) and returns `referenced port not found` when it misses. There is no RoleScoped fallback, unlike `component-discovery`'s `resolvePortRef`, which tries both keys. The error surfaces through `instance_scale.go:275-277`, so the whole component fails to start:

```
RoleScoped target : err=referenced port not found: prefill.leader.leader-grpc
                    (key: pd-server-prefill-0-leader-0.leader-grpc)
PodScoped target  : err=<nil>  env=[{Name:LEADER_GRPC_PORT Value:32423}]
```

Two things would help: state that `from` must target a PodScoped port and always resolves against the target component's pod 0 (`GetReferencePodName` hard-codes id 0), and point readers at `component-discovery`'s `portRefs` when they need a RoleScoped port. The guide version at line 322 already gets the pod-0 part right and uses a PodScoped target in its example, so it is only the main document that needs this.

doc/best-practice/zh/05-port-allocation-and-service-discovery.md
```diff
@@ -0,0 +1,663 @@
...
+      portRange: 5000
+```
+
+### 端口作用域
+
+端口分配支持两种作用域，通过 `scope` 参数控制：
+
+| 作用域 | 说明 | 适用场景 |
+| --- | --- | --- |
+| `PodScoped` | 每个 Pod 分配独立的端口值 | 同一组件的多个 Pod 需要不同端口（hostNetwork 场景） |
```
cheyang, line 278: `PodScoped` is described as giving每个 Pod an independent port, and the third-layer opening (lines 222-245) frames dynamic allocation as the fix for hostNetwork collisions. The implementation disclaims exactly that guarantee, in its own comment at `pkg/port-allocator/random.go:28-31`:

> It is best-effort only: uniqueness is guaranteed within a single AllocateBatch call, but NOT across multiple calls or multiple instances.

PodScoped allocation calls `AllocateBatch` once per pod (`manager.go:243-253`), which is precisely the across-calls case. `Release()` is also a no-op (`random.go:81-83`), so freed ports are never accounted for, and line 249 notes that `random` is the only strategy available.

A deterministic demonstration: 9 PodScoped ports drawn from a range of 8 produce duplicates by pigeonhole (`{30003:[2,7,8], 30004:[1,4], 30006:[0,6]}`); with 33 out of 32, eleven values repeat.

This is the finding most likely to bite someone in production, so it seems worth a short warning: uniqueness holds only within one batch, set `--port-range` well above the expected concurrent pod count, and have the engine retry on bind failure.

doc/best-practice/zh/05-port-allocation-and-service-discovery.md
```diff
@@ -0,0 +1,663 @@
...
+| `RBG_ROLE_NAME` | `prefill` | 角色名称 |
+
+
+**worker-0 Pod：**
+
+| 环境变量 | 示例值 | 说明 |
+| --- | --- | --- |
+| `WORKER_GRPC_PORT` | `31205` | PodScoped 端口（每个 worker 不同） |
+| `LEADER_GRPC_PORT` | `30142` | 从 leader 的 RoleScoped 端口引用 |
+| `LEADER_ADDR` | `pd-server-prefill-0-...s-pd-server-prefill.default.svc.cluster.local` | leader 的 FQDN 地址 |
```
cheyang, line 597: The ellipsis swallows `leader-0` and the dot that separates it, so the sample reads `pd-server-prefill-0-...s-pd-server-prefill.default.svc.cluster.local`. Pasted together that is not a resolvable name. In the router table the same truncation collapses `LEADER_ADDR`, `WORKER_0_ADDR` and `WORKER_1_ADDR` into three identical strings, removing exactly the part that distinguishes them.

From a real cluster:

```
LEADER_ADDR=pd-prefill-0-leader-0.s-pd-prefill.<ns>.svc.cluster.local
```

Spelling them out in full, with `-leader-0.`, `-worker-0.` and `-worker-1.` respectively, would make the table useful for copy-paste. Lines 606/614/616/618 and en:600/609/617/619/621 are the same.

doc/best-practice/zh/05-port-allocation-and-service-discovery-guide.md
```diff
@@ -0,0 +1,499 @@
...
+kubectl get cm -n default pd-inference -o yaml
+```
+
+```bash
+# 查看 Pod 中挂载的 ConfigMap 内容
+kubectl exec -n default pd-inference-prefill-0 -- cat /etc/rbg/config.yaml
+
+> group:
+>   name: pd-inference
+>   size: 2
```
cheyang, line 185: `config_builder.go:91` serialises with `sigs.k8s.io/yaml`, which sorts keys, so the real file is `name`, `roles`, `size` and then `instances`, `size`, with `decode` before `prefill` in the roles map. This block prints `size` before `roles`. The scale-out block later in the same guide (lines 228-239) already uses the sorted order, so the two disagree with each other.

From a live cluster:

```
group:
  name: a
  roles:
  - prefill
  size: 1
roles:
  prefill:
    instances:
    - address: a-prefill-0.s-a-prefill
    size: 2
```

The content matches the doc exactly; only the ordering differs. en:184-209 too.

doc/best-practice/zh/05-port-allocation-and-service-discovery.md
```diff
@@ -0,0 +1,663 @@
...
++ Kubernetes 集群版本 >= 1.24
++ 已安装 RBG Controller（参考 [安装指南](https://github.com/sgl-project/rbg)）
++ 端口分配功能需要启用 `--enable-port-allocator`（默认关闭）
+
+---
+
+## 第 1 层：Headless Service 与 DNS 服务发现
+### 自动创建的 Headless Service
+
+RBG Controller 为每个角色自动创建一个 Headless Service（`ClusterIP: None`），命名规则为：
```
cheyang, line 32: `pkg/utils/service_utils.go:30-45` keeps the old name when a Service already exists as `{rbgName}-{roleName}` without the `s-` prefix, and only uses the new form otherwise. Names longer than 63 characters are also truncated with any trailing `-` removed (`api/workloads/v1alpha2/helper.go:106-116`). Both matter to someone upgrading from an older release and looking for their service. A footnote would cover it. New installs behave exactly as documented, verified live: `s-a-prefill`, `s-pd-prefill`, both `ClusterIP: None`. en:33-39 too.

doc/best-practice/zh/05-port-allocation-and-service-discovery.md
```diff
@@ -0,0 +1,663 @@
...
+```
+
+### 端口作用域
+
+端口分配支持两种作用域，通过 `scope` 参数控制：
+
+| 作用域 | 说明 | 适用场景 |
+| --- | --- | --- |
+| `PodScoped` | 每个 Pod 分配独立的端口值 | 同一组件的多个 Pod 需要不同端口（hostNetwork 场景） |
+| `RoleScoped` | 角色内所有 Pod 共享同一个端口值 | 角色对外暴露统一端口（组件间通信） |
```
cheyang, line 279: "角色内所有 Pod 共享同一个端口值" is accurate. The RoleScoped port lives on the RoleInstanceSet annotation and is copied to every RoleInstance (`manager.go:235-240`). Every example here uses `replicas: 1`, though, and with several replicas plus hostNetwork that shared value could put two pods on one node in conflict. One sentence on when RoleScoped is the right choice would close the gap.

doc/best-practice/en/05-port-allocation-and-service-discovery.md
```diff
@@ -0,0 +1,666 @@
...
++ [Configuring Rolling Update Strategies](./03-configuring-rolling-updates.md)
++ In-Place Update and In-Place Scheduling
```
cheyang, line 666: Nothing to fix here. Recording it so it does not look like an oversight later: `In-Place Update and In-Place Scheduling` is plain text because `doc/best-practice/en/04-*` did not exist, while the zh side links it. #415 adds the English 04, so this can become a link once that merges.

cheyang (changes requested):
This is a strong document, the most accurate of the three in this batch. I checked the third layer especially closely and the core mechanics all hold up: the `s-{rbg}-{role}` naming, all nine `RBG_*` variable names character for character against `api/workloads/constants/env.go`, their Downward API label sources, the ConfigMap structure and port-name conversion rules, the component-discovery FQDN format including the PodScoped to RoleScoped fallback, the Helm values paths, and the `30000` to `34999` arithmetic. Every YAML example passes `--dry-run=server`. The four env-var sub-tables split by applicability are precisely right: a standalone role gets exactly the base plus Stateful plus RoleInstanceSet variables and no `RBG_LWP_*`, which is what I measured on real pods.

Evidence for everything below, including a live-cluster run, is on `verify/pr402-port-alloc-doc` at `cheyang/rbg@95ca3e02`, under `docs/verification/pr402-port-alloc-doc/`.

## What blocks merge

Two gaps, both in the third layer.

`references` is presented as scope-agnostic, and the complete example allocates the leader's port as `RoleScoped`. But `references` only ever builds a PodScoped key and returns `referenced port not found` otherwise, which fails the whole component's pod creation. The guide version already gets this right; it is the main document that needs the constraint spelled out.

Dynamic port allocation is framed as the answer to hostNetwork conflicts, while `random.go:28-31` explicitly disclaims uniqueness across calls, and PodScoped allocates once per pod, which is exactly that case. `Release()` is a no-op too, so freed ports are never reclaimed in the accounting. Since `random` is the only strategy available, this is the one most likely to bite someone in production, and a short warning would cover it.

## Note on the two major items

Both look like implementation gaps as much as doc gaps. A RoleScoped fallback in `references` would mirror what `component-discovery` already does, and a uniqueness-tracking allocator would make the hostNetwork framing true. If either is on the roadmap, the doc could describe the intended behaviour instead of the current limitation. Your call.

## Smaller things

The `LEADER_ADDR` samples are truncated by an ellipsis that removes `-leader-0` and its separating dot, which also collapses three distinguishable router-table rows into one identical string. The ConfigMap key order in one guide block disagrees with both the generated output and the other block in the same guide. And the Service naming rule has a legacy fallback that matters to anyone upgrading.

## Not verified

The port allocator was not enabled on the cluster when I started, so F1 and F2 are proven deterministically rather than live. F2 uses the pigeonhole principle (9 ports from a range of 8), which is exact and will not flake, and deliberately does not try to provoke a probabilistic collision in a 5000-port range. The live run allocates a single port on a single component, so please do not read it as evidence that allocation is collision-free.

cheyang (approved):
/lgtrm
/approve

PR #403, opened 2026-07-17
chore(rbgs): reorganize Helm chart templates into kubebuilder-style l…

diw-zw:
### Ⅰ. Motivation

The Helm chart templates under `deploy/helm/rbgs/templates/` were previously kept in a single flat directory, mixing manager, RBAC, webhook and CRD-upgrade resources together. As the chart grows, this flat layout makes it harder to locate and maintain resources.

This PR reorganizes the templates into a **kubebuilder-style directory layout** (`manager/`, `rbac/`, `webhook/`, `crd-upgrade/`). This is the same convention adopted by other Kubernetes projects in the ecosystem such as **LWS (LeaderWorkerSet)** and **Kueue**. Since RBG is a kubebuilder-based project, aligning its deployment/config layout with kubebuilder's own organizational logic keeps the project idiomatic, improves readability, and lowers the onboarding cost for contributors already familiar with kubebuilder conventions.

### Ⅱ. Modifications

Restructured `deploy/helm/rbgs/templates/` into kubebuilder-style subdirectories (pure moves/renames, no functional change to rendered manifests):

| Before | After |
| --- | --- |
| `manager.yaml` | `manager/manager.yaml` |
| `clusterrole.yaml` | `rbac/clusterrole.yaml` |
| `rolebinding.yaml` | `rbac/clusterrole_binding.yaml` |
| `secret-role.yaml` | `rbac/cert_role.yaml` + `rbac/cert_role_binding.yaml` |
| `serviceaccount.yaml` | `rbac/service_account.yaml` |
| `validating-webhook-configuration.yaml` | `webhook/manifests.yaml` |
| `webhook-service.yaml` | `webhook/service.yaml` |
| `upgrade/crd-upgrade.yaml` | `crd-upgrade/crd-upgrade.yaml` |

Additionally, the combined `secret-role.yaml` (which bundled both a `Role` and a `RoleBinding`) is split into two dedicated files — `rbac/secret_role.yaml` and `rbac/secret_role_binding.yaml` — following the kubebuilder convention of one resource per file.

NONE

No tests needed. This change only relocates and renames Helm template files; the rendered Kubernetes manifests remain functionally identical.

Render the chart and confirm the output is unchanged from `main`:

```bash
helm template deploy/helm/rbgs > /tmp/after.yaml
git stash && helm template deploy/helm/rbgs > /tmp/before.yaml && git stash pop
diff <(yq -P 'sort_keys(..)' /tmp/before.yaml) <(yq -P 'sort_keys(..)' /tmp/after.yaml)
```

The rendered resources should be equivalent (a `helm lint deploy/helm/rbgs` should also pass).

This is purely a directory/file reorganization aligning the Helm chart with kubebuilder's conventions (as seen in LWS and Kueue). No template logic or rendered output changes, except that the secret Role/RoleBinding are now in separate files.

## Checklist


cheyang (commented):
Thanks for the cleanup. I rendered the chart before/after locally and the output is identical apart from the `# Source:` path comments — same resource set (kind + name) and same document ordering, so the move itself is safe.

A few naming/scope notes (all non-blocking):

- **Layout** matches kubebuilder's `config/` scaffolding (`rbac/` `manager/` `webhook/`), which is a fine convention to align to. Left an inline note on the rbac file names + the coupled Makefile target.
- **`webhook/manifests.yaml`**: `manifests.yaml` is kubebuilder's name for the *generated* webhook config, but in a hand-maintained chart it's vaguer than the previous `validating-webhook-configuration.yaml`. I'd keep the descriptive name.
- **`crd-upgrade/crd-upgrade.yaml`**: the filename repeats the directory. `crd-upgrade/job.yaml` (or similar) is enough now that it's in its own folder.
- Dropping the `deploy/helm/rbgs/templates/engineruntime` line from `uninstall` is a good catch — that path doesn't exist on main, so it was a `kubectl delete -f` against a missing directory. Since it's outside the pure-move scope, worth a one-line mention in the PR description.

Makefile
```diff
@@ -60,8 +60,8 @@ help: ## Display this help.
 .PHONY: manifests
 manifests: controller-gen kustomize ## Generate WebhookConfiguration, CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) crd:allowDangerousTypes=true,crdVersions=v1,generateEmbeddedObjectMeta=true,ignoreUnexportedFields=true,maxDescLen=200 rbac:roleName=controller-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases output:rbac:artifacts:config=config/rbac
-	cp config/rbac/role.yaml deploy/helm/rbgs/templates/clusterrole.yaml
-	sed -i.bak 's/name: controller-role/name: rbgs-controller-role/' deploy/helm/rbgs/templates/clusterrole.yaml && rm -f deploy/helm/rbgs/templates/clusterrole.yaml.bak
+	cp config/rbac/role.yaml deploy/helm/rbgs/templates/rbac/role.yaml
+	sed -i.bak 's/name: controller-role/name: rbgs-controller-role/' deploy/helm/rbgs/templates/rbac/role.yaml && rm -f deploy/helm/rbgs/templates/rbac/role.yaml.bak
```
cheyang: Two things on this target.

`config/rbac/role.yaml` is actually a **ClusterRole** (kubebuilder's historical file naming), so landing it as `rbac/role.yaml` next to the namespaced `secret_role.yaml` is a bit misleading: the cluster-scoped resource gets the generic name while the namespaced cert Role carries the `secret_` qualifier. Naming by kind would read better — `rbac/clusterrole.yaml` + `rbac/clusterrolebinding.yaml`, and a dedicated cert Role for the namespaced one.

Whatever you rename it to, this `cp`/`sed` pair has to move with it. If the file is renamed but this line still writes `rbac/role.yaml`, the next `make manifests` regenerates `role.yaml` next to the renamed file, and `helm template` then renders two copies of the ClusterRole.

deploy/helm/rbgs/templates/rbac/cert_role_binding.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
cheyang, line 1: Minor: `secret_role` / `secret_role_binding` reads as a generic "Role for secrets". This specifically scopes the controller to the webhook cert secret, so `cert_role.yaml` / `cert_role_binding.yaml` would say what it is. These are hand-written (not produced by `make manifests`), so renaming them is a pure file move with no Makefile change needed.

cheyang:
@gemini-code-assist /review

cheyang:
@gemini-code-assist /review

cheyang:
Delta since the previous review (1b33e1b0) is limited to four 100%-similarity file renames (`role`/`role_binding` -> `clusterrole`/`clusterrole_binding`, `secret_role`/`secret_role_binding` -> `cert_role`/`cert_role_binding`) plus the corresponding Makefile path update. Kind/name coupling between each Role/ClusterRole and its Binding is preserved and CI is fully green on 69aded44. The new names are intentional — `clusterrole*` matches the actual `kind: ClusterRole` (clearer than kubebuilder's default `role.yaml`), and `cert_role*` was requested in the previous review round because this Role is scoped specifically to the webhook cert secret rather than "secrets" in general. Automated bot feedback (Gemini) pushing back on these names is based on the stale PR description; the code direction is correct. Two follow-ups worth doing before merge: refresh the Before/After table in the PR body so it matches the current filenames, and (optionally) align the in-cluster resource name of `cert_role` with its filename. No blockers.

cheyang (approved):
Reviewed this as a pure reorganization and confirmed it's behavior-preserving. Rendered the chart on both the merge-base and this head with `helm template` — identical after stripping the `# Source:` path comments (12 objects, unchanged). `kustomize build config/default` is byte-identical too (17 objects), so `make manifests` output doesn't drift from the `config/rbac` rename. `helm lint` passes, the RBAC `roleRef`/subject references all still resolve, and nothing outside the chart references the old flat paths. LGTM.

Two minor, non-blocking notes:
- The commit "Rename rbac resources' name: role -> clusterrole, secret_role -> cert_role" only renames *files* — the resource `metadata.name` values are unchanged (which is exactly why the render stays identical). Might be worth rewording so it isn't read as a resource-name change.
- The controller edit is comment-only (updates a file path in a doc comment) — correct, just noting it's not a behavior change.

cheyang (approved):
/lgtm
/approve

PR #406, opened 2026-07-20
refactor(helm): group values under global/controller/crdUpgrade and u…

diw-zw:
### Ⅰ. Motivation

The rbgs Helm chart `values.yaml` used a flat structure that mixed controller settings, image config, feature flags, and CRD-upgrade settings at the top level. This made the values hard to read and left no clear place for shared config. This PR reorganizes the values into logical groups and introduces a shared `global` section that individual components can override.

### Ⅱ. Modifications

**Chart values (`deploy/helm/rbgs/values.yaml`)**
- Add top-level `global.imagePullSecrets`, shared by the controller Deployment and the crd-upgrade Job, overridable per component via  `controller.imagePullSecrets` / `crdUpgrade.imagePullSecrets`.
- Nest controller settings under `controller.*`: `replicaCount`, `image`  (with `pullPolicy` moved inside), `resources`, `nodeSelector`, `tolerations`,  `podAnnotations`, `podSecurityContext`, `securityContext`, `tuning`  (renamed from`controllerTuning`), `pprof`, and a new `features` group holding  `gangScheduling.schedulerName` (was top-level `schedulerName`) and  `portAllocator`.
- Nest CRD upgrader image fields under `crdUpgrade.image.*`  (`repository` / `tag` / `pullPolicy`, previously flat  `repository` / `tag` / `imagePullPolicy`).

**Templates**
- Update `templates/manager/manager.yaml` and  `templates/crd-upgrade/crd-upgrade.yaml` to the new value paths.
- `imagePullSecrets` uses component-level value with fallback to  `global.imagePullSecrets`.

**Tooling & docs**
- `Makefile` (`helm-deploy`): `--set image.tag` → `--set controller.image.tag`.
- `tools/release/update-release.sh`: update the `sed` to match the 4-space  indented `tag:` fields (`controller.image.tag` and `crdUpgrade.image.tag`).
- Docs: `doc/install.md`, `doc/dev/how_to_develop.md` (pprof),  `doc/dev/how_to_release.md`, `doc/features/gang-scheduling.md`.

NONE

No new tests. This is a Helm chart value/field reorganization; correctness isverified by rendering the chart (`helm template`).

```bash
# Chart renders with defaults
helm template deploy/helm/rbgs

# Per-component override falls back to global correctly
helm template deploy/helm/rbgs \
  --set 'controller.imagePullSecrets[0].name=ctrl-secret' \
  --set 'global.imagePullSecrets[0].name=global-secret' | grep -A1 imagePullSecrets
# controller uses ctrl-secret; crd-upgrade Job falls back to global-secret

# Release script sed still targets both image tags
sed -E 's/^(    tag:[[:space:]]+).*/\1"vTEST"/' deploy/helm/rbgs/values.yaml | grep 'tag:'
```

This is a **breaking change** for anyone overriding chart values: existing`--set`/custom values files using the old flat keys (e.g. `image.tag`,`schedulerName`, `controllerTuning`, `portAllocator`, `crdUpgrade.tag`) must be migrated to the new nested paths.

## Checklist

### DIFF
--- a/deploy/helm/rbgs/templates/crd-upgrade/crd-upgrade.yaml
+++ b/deploy/helm/rbgs/templates/crd-upgrade/crd-upgrade.yaml
@@ -15,7 +15,7 @@ spec:
 {{- end }}
   template:
     spec:
-      {{- with .Values.imagePullSecrets }}
+      {{- with .Values.crdUpgrade.imagePullSecrets | default .Values.global.imagePullSecrets }}
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
@@ -30,8 +30,8 @@ spec:
       serviceAccountName: rbgs-crds-upgrade
       containers:
         - name: rbgs-crds-upgrade
-          image: "{{ .Values.crdUpgrade.repository }}:{{ .Values.crdUpgrade.tag | default .Chart.AppVersion }}"
-          imagePullPolicy: {{ .Values.crdUpgrade.imagePullPolicy | default "IfNotPresent" }}
+          image: "{{ .Values.crdUpgrade.image.repository }}:{{ .Values.crdUpgrade.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.crdUpgrade.image.pullPolicy | default "IfNotPresent" }}
           env:
             - name: WEBHOOK_NAMESPACE
               value: {{ .Release.Namespace | quote }}
--- a/deploy/helm/rbgs/templates/crd-upgrade/crd-upgrade.yaml
+++ b/deploy/helm/rbgs/templates/crd-upgrade/crd-upgrade.yaml
@@ -15,7 +15,7 @@ spec:
 {{- end }}
   template:
     spec:
-      {{- with .Values.crdUpgrade.imagePullSecrets | default .Values.global.imagePullSecrets }}
+      {{- with .Values.crdUpgrade.imagePullSecrets | default (.Values.global).imagePullSecrets }}
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
--- a/deploy/helm/rbgs/templates/manager/manager.yaml
+++ b/deploy/helm/rbgs/templates/manager/manager.yaml
@@ -6,47 +6,47 @@ metadata:
   labels:
     control-plane: rbgs-controller
 spec:
-  replicas: {{ .Values.replicaCount }}
+  replicas: {{ .Values.controller.replicaCount }}
   selector:
     matchLabels:
       control-plane: rbgs-controller
   template:
     metadata:
-      {{- with .Values.podAnnotations }}
+      {{- with .Values.controller.podAnnotations }}
       annotations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       labels:
         control-plane: rbgs-controller
     spec:
-      {{- with .Values.imagePullSecrets }}
+      {{- with .Values.controller.imagePullSecrets | default .Values.global.imagePullSecrets }}
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: rbgs-controller-sa
       securityContext:
-        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+        {{- toYaml .Values.controller.podSecurityContext | nindent 8 }}
       containers:
         - name: {{ .Chart.Name }}
           args:
             - --metrics-bind-address=:8443
             - --leader-elect
             - --health-probe-bind-address=:8081
-            {{- if .Values.controllerTuning }}
-            - --max-concurrent-reconciles={{ .Values.controllerTuning.maxConcurrentReconciles | default 10 }}
-            - --kube-api-qps={{ .Values.controllerTuning.kubeApiQPS | default 20 }}
-            - --kube-api-burst={{ .Values.controllerTuning.kubeApiBurst | default 30 }}
+            {{- if .Values.controller.tuning }}
+            - --max-concurrent-reconciles={{ .Values.controller.tuning.maxConcurrentReconciles | default 10 }}
+            - --kube-api-qps={{ .Values.controller.tuning.kubeApiQPS | default 20 }}
+            - --kube-api-burst={{ .Values.controller.tuning.kubeApiBurst | default 30 }}
             {{- end }}
-            {{- if .Values.portAllocator.enabled }}
+            {{- if .Values.controller.features.portAllocator.enabled }}
             - --enable-port-allocator=true
-            - --port-allocate-strategy={{ .Values.portAllocator.strategy | default "random" }}
-            - --start-port={{ .Values.portAllocator.startPort | default 30000 }}
-            - --port-range={{ .Values.portAllocator.portRange | default 5000 }}
+            - --port-allocate-strategy={{ .Values.controller.features.portAllocator.strategy | default "random" }}
+            - --start-port={{ .Values.controller.features.portAllocator.startPort | default 30000 }}
+            - --port-range={{ .Values.controller.features.portAllocator.portRange | default 5000 }}
             {{- end }}
-            - --scheduler-name={{ .Values.schedulerName | default "scheduler-plugins" }}
-            {{- if .Values.pprof.enabled }}
+            - --scheduler-name={{ .Values.controller.features.gangScheduling.schedulerName | default "scheduler-plugins" }}
+            {{- if .Values.controller.pprof.enabled }}
             - --enable-pprof=true
-            - --pprof-bind-address={{ .Values.pprof.bindAddress | default ":6060" }}
+            - --pprof-bind-address={{ .Values.controller.pprof.bindAddress | default ":6060" }}
             {{- end }}
           env:
             - name: POD_NAMESPACE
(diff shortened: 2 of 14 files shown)


cheyang:
@gemini-code-assist /review

tools/release/update-release.sh
```diff
@@ -76,8 +76,9 @@ fi
 
 VALUES_FILE="${CHARTS_DIR}/rbgs/values.yaml"
 if [[ -f "$VALUES_FILE" ]]; then
-    # Update tag field (quoted, to avoid YAML numeric/scientific-notation coercion)
-    sed -i.bak -E "s/^(  tag:[[:space:]]+).*/\1\"$TAG\"/" "$VALUES_FILE"
+    # Update tag fields (controller.image.tag and crdUpgrade.image.tag, both at
+    # 4-space indent). Quoted to avoid YAML numeric/scientific-notation coercion.
+    sed -i.bak -E "s/^(    tag:[[:space:]]+).*/\1\"$TAG\"/" "$VALUES_FILE"
```
cheyang: The regex `s/^(    tag:[[:space:]]+).*/\1"$TAG"/` will match any YAML key named `tag` at exactly 4 spaces of indentation. Today that correctly targets `controller.image.tag` and `crdUpgrade.image.tag`, but if a future values addition introduces another `tag:` at that indent level it would be silently overwritten. Consider anchoring via a two-pass approach or using `yq` for targeted edits.

deploy/helm/rbgs/templates/manager/manager.yaml
```diff
@@ -56,22 +60,23 @@ spec:
...
             - /manager
           securityContext:
-            {{- toYaml .Values.securityContext | nindent 12 }}
-          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
-          imagePullPolicy: {{ .Values.image.pullPolicy }}
-          {{- if .Values.pprof.enabled }}
+            {{- toYaml .Values.controller.securityContext | nindent 12 }}
+          image: "{{ .Values.controller.image.repository }}:{{ .Values.controller.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.controller.image.pullPolicy | default "IfNotPresent" }}
+          {{- $pprof := .Values.controller.pprof | default dict }}
```
cheyang: The template variable `$pprof` is assigned identically at line 50 and again at line 66 (`{{- $pprof := .Values.controller.pprof | default dict }}`). Because Go template `:=` is block-scoped this is technically correct, but pulling the assignment to the top of the `containers` block once would reduce duplication and make the template easier to follow.

cheyang:
Nice cleanup of the values layout. One thing slipped through the rename: `test/stress/scripts/deploy-controller.sh` (lines 55-65) isn't part of this PR and still passes the old flat `--set` paths — `image.tag`, `replicaCount`, `resources.*`, `controllerTuning.*`, `pprof.*`. Since the chart no longer reads those keys, Helm accepts them silently and drops the overrides, so the stress deploy quietly falls back to chart defaults. I confirmed this: `--set image.tag=<x>` leaves the rendered image at the default tag, and `--set controllerTuning.maxConcurrentReconciles=99` still renders `--max-concurrent-reconciles=10`. Worth repointing those to `controller.image.tag`, `controller.replicaCount`, `controller.resources.*`, `controller.tuning.*`, and `controller.pprof.*` — and note `pprof.port` was never a valid key even before this PR (the chart uses `containerPort`). The two inline comments below are a smaller instance of the same theme: a regrouped value that isn't fully guarded.

deploy/helm/rbgs/templates/manager/manager.yaml
```diff
@@ -6,47 +6,51 @@ metadata:
...
-      {{- with .Values.podAnnotations }}
+      {{- with .Values.controller.podAnnotations }}
       annotations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       labels:
         control-plane: rbgs-controller
     spec:
-      {{- with .Values.imagePullSecrets }}
+      {{- with .Values.controller.imagePullSecrets | default .Values.global.imagePullSecrets }}
```
cheyang: `.Values.global.imagePullSecrets` is dereferenced here without safe navigation. `global` is a Helm-reserved key with null-coalescing semantics, so a parent chart or an override file that leaves `global` null makes the render fail: `helm template --set global=null` aborts with `nil pointer evaluating interface {}.imagePullSecrets` on this line. The `features` and `pprof` paths in this same PR were guarded with `default dict`, so this one is the odd exception. Wrapping in parens handles it: `| default (.Values.global).imagePullSecrets`.

deploy/helm/rbgs/templates/crd-upgrade/crd-upgrade.yaml
```diff
@@ -15,7 +15,7 @@ spec:
 {{- end }}
   template:
     spec:
-      {{- with .Values.imagePullSecrets }}
+      {{- with .Values.crdUpgrade.imagePullSecrets | default .Values.global.imagePullSecrets }}
```
cheyang: Same unguarded `.Values.global.imagePullSecrets` as in manager.yaml — a null `global` (parent chart or override file) will nil-pointer the render here too. `| default (.Values.global).imagePullSecrets` fixes it.

cheyang:
@gemini-code-assist /review

cheyang:
Thanks for the quick turnaround on the round-1 comments — I re-checked the two follow-ups on this push and both look right:

- manager.yaml: hoisting `$pprof` to a single declaration at the top of the container block reads cleanly, and `helm template` renders identically whether pprof is enabled or not (the args flag and the ports block both resolve from the one variable), so the scoping is correct.
- update-release.sh: swapping the blanket `sed` for the section-aware `awk` does exactly what the earlier comment asked. Running it against the current values.yaml rewrites only `controller.image.tag` and `crdUpgrade.image.tag` and leaves every other key untouched.

One item from the previous round is still open, and it's non-blocking: `.Values.global.imagePullSecrets` is dereferenced directly in manager.yaml (line 22) and crd-upgrade.yaml (line 18). A null `global` (e.g. `helm template --set global=null`, or a parent chart that leaves `global` unset) still fails the render with `nil pointer evaluating interface {}.imagePullSecrets`. Wrapping it as `| default (.Values.global).imagePullSecrets` fixes it. Happy to see this merge as-is if you'd prefer to handle that guard separately.

For context, the failing e2e-test / e2e-test-manifest checks look environmental rather than caused by this change — the manifest job exercises the kustomize path (independent of these Helm values), and both jobs just show the test workloads never reaching Ready, which matches the earlier round's read on that flake.

cheyang (commented):
Re-reviewed at head `8e45659a` (thanks for the update). The `global.imagePullSecrets` nil-pointer is resolved now — with the safe-navigation change in `manager.yaml`/`crd-upgrade.yaml`, `helm template ./deploy/helm/rbgs --set global=null` renders cleanly. 

One blocking item still remains:

**(Blocking) `test/stress/scripts/deploy-controller.sh` wasn't updated for the regroup.** Its `--set` lines still use the old flat paths (`image.tag`, `resources.limits/requests.*`, `controllerTuning.*`, `pprof.*`). After this PR those top-level keys no longer exist, and `helm --set` silently ignores unknown keys — so the stress deploy falls back to chart defaults (wrong image tag, 2 replicas, default reconcile/QPS tuning) with no error, which quietly invalidates stress runs. It needs the same `controller.*` rename you already applied to the Makefile, the CI workflows, and the release script.

Minor: that script also sets `--set pprof.port=...`, which isn't a chart key (the chart uses `pprof.bindAddress` / `pprof.containerPort`) — it's been a silent no-op. And it'd be worth a short NOTES.txt / upgrade note, since anyone with an existing values file on the old top-level keys (`image`, `imagePullSecrets`, `resources`, `controllerTuning`, `pprof`, ...) will have those overrides silently ignored after upgrading — the same failure mode as the script above.

cheyang (commented):
Re-verified at head `f05428e4`: the stress-script `--set` paths and the `global` nil-pointer are both resolved, and nicely the old no-op `--set pprof.port` is now `controller.pprof.bindAddress` + `controller.pprof.containerPort`. LGTM on the blockers.

One small follow-up (non-blocking): this PR renames breaking top-level keys (`image`, `imagePullSecrets`, `resources`, `schedulerName`, `controllerTuning`, `pprof`, `portAllocator`) into `controller.*` / `global.*`. Anyone with an existing custom values file on the old keys will have those overrides **silently ignored** after upgrading — the same failure mode the stress script just hit. A short `templates/NOTES.txt` (or an upgrade note in the chart README / release notes) calling out the renamed keys would save users a confusing silent misconfiguration. Worth adding either here or as a quick follow-up.

cheyang (commented):
For F4 — here's a ready-to-drop `deploy/helm/rbgs/templates/NOTES.txt`. The migration warning is gated to the version window where the regroup is breaking — **introduced in 0.7.0, no further breaking after 0.8.0** — so it fires for upgrades landing in `>=0.7.0 <0.9.0` and auto-retires at 0.9.0. Validated against a live release via `helm upgrade --dry-run` (chart `0.8.0-alpha.2` → warning shown) and the version window checked across a matrix:

```gotmpl
{{- if .Release.IsInstall }}
{{ .Chart.Name }} ({{ .Chart.AppVersion }}) installed in namespace "{{ .Release.Namespace }}".

Check the controller:
  kubectl -n {{ .Release.Namespace }} get deploy rbgs-controller-manager
  kubectl -n {{ .Release.Namespace }} get pods -l control-plane=controller-manager
{{- end }}
{{- if and .Release.IsUpgrade (semverCompare ">=0.7.0-0 <0.9.0-0" .Chart.Version) }}
{{ .Chart.Name }} upgraded to {{ .Chart.Version }} in namespace "{{ .Release.Namespace }}".

BREAKING VALUES CHANGE (chart values were regrouped as of 0.7.0): keys are now nested under
`global`, `controller`, and `crdUpgrade`. If your values file or --set flags still use the OLD
top-level keys, Helm SILENTLY IGNORES them and the controller runs with chart defaults. Migrate:

  image.*                     ->  controller.image.*
  replicaCount                ->  controller.replicaCount
  resources.*                 ->  controller.resources.*
  nodeSelector / tolerations  ->  controller.nodeSelector / controller.tolerations
  podAnnotations / *SecurityContext -> controller.*
  imagePullSecrets            ->  global.imagePullSecrets
                                  (or controller.imagePullSecrets / crdUpgrade.imagePullSecrets)
  controllerTuning.*          ->  controller.tuning.*
  pprof.enabled               ->  controller.pprof.enabled
  pprof.port                  ->  controller.pprof.bindAddress (e.g. ":6060")
                                  + controller.pprof.containerPort
  schedulerName               ->  controller.features.gangScheduling.schedulerName
  portAllocator.*             ->  controller.features.portAllocator.*

Verify your effective values after upgrading:
  helm -n {{ .Release.Namespace }} get values {{ .Release.Name }}
{{- end }}
```

The `-0` on **both** bounds (`>=0.7.0-0 <0.9.0-0`) is deliberate — Masterminds/`semverCompare` excludes prerelease versions from a range unless the comparators carry a prerelease tag. Verified matrix:

| chart version | in window? |
|---|---|
| 0.6.9 | false |
| 0.7.0-alpha.1 | true |
| 0.7.0 / 0.7.3 | true |
| 0.8.0-alpha.2 | true |
| 0.8.0 / 0.8.5 | true |
| 0.9.0-alpha.1 | false |
| 0.9.0 / 1.0.0 | false |

Without the `-0` on the upper bound, `0.8.0-alpha.2` (the current chart version) would be skipped. Note Helm doesn't expose the *previous* chart version to templates, so gating on the new `.Chart.Version` window is the practical approach; this window closes itself at 0.9.0.

With this in place F4 is fully covered and #406 looks good to merge.

cheyang (approved):
/lgtm
/approve

PR #407, opened 2026-07-21
test(e2e): stabilize flaky v1alpha2 specs on slow CI clusters

diw-zw:
### Ⅰ. Motivation

The v1alpha2 e2e test cases contained several sources of flakiness that could cause spurious failures, especially on slower clusters:

1. **Race on RBG updates** — Several scaling-adapter tests did a `Get` immediately followed by an `Update`. When the controller wrote to the RBG in between, the `Update` failed with a `409 Conflict`, causing the test to fail non-deterministically.
2. **Reliance on reconcile timing** — The coordinated-policy scaling test used `MaxSkew=50%`, which allows the scaler to jump `1->3` in a single batch.
3. **Immediate readiness assertions** — `RoleInstanceSet` tests asserted ready-replica counts with a single synchronous `Expect`, which does not tolerate reconcile latency.

### Ⅱ. Modifications

- **`scaling_adapter.go`**: Introduced a `updateRbgSpecV2Retry` helper that re-fetches the RBG and re-applies the mutation inside a `gomega.Eventually`, retrying on `409 Conflict`. Refactored the three role-mutation test steps (delete role, add role, toggle scaling adapter) to use it. The caller's original `rbg` is left untouched so subsequent assertions still reference it.
- **`coordinated_policy.go`**: Changed `MaxSkew` from `50%` to `33%` so that, with `desiredReplicas=3`, the scaler advances one replica per batch (`1->2->3`). Combined with `OrderReady` gating, this bounds the ready-replicas skew to `1` by design rather than by timing. Updated the comments to explain the reasoning.
- **`ris.go`**: Replaced synchronous `gomega.Expect(countReadyRoleInstances(...))` readiness checks with `gomega.Eventually(...)` polling (using `utils.Timeout`/`utils.Interval`) to tolerate reconcile latency.

### DIFF
--- a/test/e2e/testcase/v1alpha2/scaling_adapter.go
+++ b/test/e2e/testcase/v1alpha2/scaling_adapter.go
@@ -24,6 +24,7 @@ import (
 	workloadsv1alpha2 "sigs.k8s.io/rbgs/api/workloads/v1alpha2"
 	"sigs.k8s.io/rbgs/pkg/scale"
 	"sigs.k8s.io/rbgs/test/e2e/framework"
+	"sigs.k8s.io/rbgs/test/utils"
 	wrappersv2 "sigs.k8s.io/rbgs/test/wrappers/v1alpha2"
 )
 
@@ -76,18 +77,9 @@ func RunRbgScalingAdapterControllerTestCases(f *framework.Framework) {
 					f.ExpectRbgV2Equal(rbg)
 					f.ExpectRbgV2ScalingAdapterEqual(rbg)
 
-					newRbg := &workloadsv1alpha2.RoleBasedGroup{}
-					gomega.Expect(
-						f.Client.Get(
-							f.Ctx, client.ObjectKey{
-								Name:      rbg.Name,
-								Namespace: rbg.Namespace,
-							}, newRbg,
-						),
-					).Should(gomega.Succeed())
-
-					newRbg.Spec.Roles = rbg.Spec.Roles[1:]
-					gomega.Expect(f.Client.Update(f.Ctx, newRbg)).Should(gomega.Succeed())
+					updateRbgSpecV2Retry(f, rbg, func(newRbg *workloadsv1alpha2.RoleBasedGroup) {
+						newRbg.Spec.Roles = rbg.Spec.Roles[1:]
+					})
 					f.ExpectScalingAdapterV2NotExist(rbg, rbg.Spec.Roles[0])
 					f.ExpectRoleScalingAdapterV2Equal(rbg, rbg.Spec.Roles[1], nil)
 
@@ -113,21 +105,12 @@ func RunRbgScalingAdapterControllerTestCases(f *framework.Framework) {
 					f.ExpectRbgV2Equal(rbg)
 					f.ExpectRbgV2ScalingAdapterEqual(rbg)
 
-					newRbg := &workloadsv1alpha2.RoleBasedGroup{}
-					gomega.Expect(
-						f.Client.Get(
-							f.Ctx, client.ObjectKey{
-								Name:      rbg.Name,
-								Namespace: rbg.Namespace,
-							}, newRbg,
-						),
-					).Should(gomega.Succeed())
-
-					newRbg.Spec.Roles = append(
-						newRbg.Spec.Roles, wrappersv2.BuildStandaloneRole("role-2").
-							WithScalingAdapter(true).Obj(),
-					)
-					gomega.Expect(f.Client.Update(f.Ctx, newRbg)).Should(gomega.Succeed())
+					newRbg := updateRbgSpecV2Retry(f, rbg, func(newRbg *workloadsv1alpha2.RoleBasedGroup) {
+						newRbg.Spec.Roles = append(
+							newRbg.Spec.Roles, wrappersv2.BuildStandaloneRole("role-2").
+								WithScalingAdapter(true).Obj(),
+						)
+					})
 					f.ExpectScalingAdapterV2NotExist(rbg, newRbg.Spec.Roles[0])
 					f.ExpectRoleScalingAdapterV2Equal(rbg, newRbg.Spec.Roles[1], nil)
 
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e-test.yml
```diff
@@ -178,50 +178,12 @@ jobs:
...
-          fi
-          echo "RBAC and webhook-cert bootstrap verified"
-
       - name: Pre-load e2e test images
         run: |
           # Pull and load e2e test images to avoid image pull timeout during tests
           E2E_IMAGES=(
-            "registry.cn-hangzhou.aliyuncs.com/acs-sample/nginx:latest"
-            "registry-cn-hangzhou.ack.aliyuncs.com/dev/patio-runtime:v0.2.0"
+            "nginx:latest"
```
cheyang: Why do we make this change?

cheyang:
@gemini-code-assist /review

cheyang (approved):
LGTM — nice, focused stabilization of the flaky v1alpha2 specs. I traced the two non-obvious changes and both hold up.

- The `MaxSkew` 50%→33% change is correct. Against the real `CoordinationScaler`, with two roles scaling 1→3 under `OrderReady`: `target = ceil((minProgress + maxSkew)·desired)`. At 50% that's `ceil((1/3 + 0.5)·3) = ceil(2.5) = 3` — it jumps 1→3 in a single batch, which is exactly what let the ready-replicas skew reach 2 and made the `skew<=1` assertion flaky. At 33% it's `ceil(1.99) = 2` then `ceil(2.99) = 3`, so it steps 1→2→3 and, combined with the `OrderReady` gate, the skew stays ≤1 by design rather than by timing. Good call binding the test to batch behavior.
- `updateRbgSpecV2Retry` is retry-safe: re-`Get`'ing into the reused object resets `Spec.Roles`, so re-applying the append after a 409 doesn't double-append; and leaving the caller's `rbg` untouched keeps the later `rbg.Spec.Roles[0]` assertions valid.

Two optional nits, non-blocking:
1. `updateRbgSpecV2Retry` overlaps heavily with the existing `updateRbgV2` (`test/e2e/testcase/v1alpha2/rbg.go:296`) — both do Get+mutate+Update inside `Eventually`. The real difference is that this one preserves the caller's object and returns the fresh copy; the `Retry` suffix reads a little oddly since `updateRbgV2` retries too. A one-line doc note (or a rename) would make the distinction clearer.
2. The `33%` ↔ `desiredReplicas=3` coupling is a magic constant. It's documented in the comment, but it'll silently stop stepping 1→2→3 if the replica count ever changes.

Approving.

cheyang:
Nice, focused pass at de-flaking the v1alpha2 e2e suite. Converting the ready-count checks in ris.go to Eventually with the shared utils.Timeout/Interval is the right call, and the coordinated_policy.go change (maxSkew 50% -> 33% with the added rationale) makes the skew bound deterministic instead of relying on timing, with a clear comment explaining why. The updateRbgSpecV2Retry helper also sensibly centralises the get-mutate-update-on-conflict pattern. All CI checks are green.

A few non-blocking things worth a look before merge, mostly around the new retry helper: it currently retries on any Update error rather than only on conflicts (its own comment says otherwise), and the add-role test's post-delete cleanup still only inspects role-1. Details inline.

test/e2e/testcase/v1alpha2/scaling_adapter.go
```diff
@@ -238,3 +212,25 @@ func RunRbgScalingAdapterControllerTestCases(f *framework.Framework) {
...
+) *workloadsv1alpha2.RoleBasedGroup {
+	newRbg := &workloadsv1alpha2.RoleBasedGroup{}
+	gomega.Eventually(func() error {
+		if err := f.Client.Get(
+			f.Ctx, client.ObjectKey{Name: rbg.Name, Namespace: rbg.Namespace}, newRbg,
+		); err != nil {
+			return err
+		}
+		mutate(newRbg)
+		return f.Client.Update(f.Ctx, newRbg)
```
cheyang, line 233: The doc comment says this retries "on a 409 Conflict", but the Eventually block treats every Update error the same way and keeps retrying until the timeout. If the Update fails for a non-retryable reason (validation, RBAC, a malformed spec), the test won't surface that error, it'll just spin for the full utils.Timeout (150s) and then fail with a generic timeout. Consider returning early / failing fast when apierrors.IsConflict(err) is false so real bugs show up immediately and only genuine conflicts get retried.

test/e2e/testcase/v1alpha2/scaling_adapter.go
```diff
@@ -113,21 +105,12 @@ func RunRbgScalingAdapterControllerTestCases(f *framework.Framework) {
...
-							}, newRbg,
-						),
-					).Should(gomega.Succeed())
-
-					newRbg.Spec.Roles = append(
-						newRbg.Spec.Roles, wrappersv2.BuildStandaloneRole("role-2").
-							WithScalingAdapter(true).Obj(),
-					)
-					gomega.Expect(f.Client.Update(f.Ctx, newRbg)).Should(gomega.Succeed())
+					newRbg := updateRbgSpecV2Retry(f, rbg, func(newRbg *workloadsv1alpha2.RoleBasedGroup) {
```
cheyang, line 108: Now that the helper returns the updated object as newRbg (and the assertions just above use newRbg.Spec.Roles), the post-Delete cleanup loop a few lines down still ranges over rbg.Spec.Roles, which only contains role-1. role-2 is the role that actually had the scaling adapter enabled, so a leak of its adapter after the RBG is deleted would go unnoticed. Iterating newRbg.Spec.Roles here would close that gap. It's pre-existing rather than introduced by this PR, but since you're already touching this block it's a cheap improvement.

test/e2e/testcase/v1alpha2/scaling_adapter.go
```diff
@@ -76,18 +77,9 @@ func RunRbgScalingAdapterControllerTestCases(f *framework.Framework) {
...
-								Name:      rbg.Name,
-								Namespace: rbg.Namespace,
-							}, newRbg,
-						),
-					).Should(gomega.Succeed())
-
-					newRbg.Spec.Roles = rbg.Spec.Roles[1:]
-					gomega.Expect(f.Client.Update(f.Ctx, newRbg)).Should(gomega.Succeed())
+					updateRbgSpecV2Retry(f, rbg, func(newRbg *workloadsv1alpha2.RoleBasedGroup) {
+						newRbg.Spec.Roles = rbg.Spec.Roles[1:]
```
cheyang, line 81: The mutate closure sets newRbg.Spec.Roles = rbg.Spec.Roles[1:], i.e. it assigns from the originally-captured rbg rather than the freshly re-fetched newRbg, and the result aliases rbg's backing array. It happens to be correct here because the target state is deterministic (always drop role-1), so a conflict retry still re-fetches for a fresh resourceVersion and reapplies the same value. It just reads a little at odds with the "re-fetch before update" intent of the helper. Deriving the new slice from newRbg.Spec.Roles[1:] (and copying it) would make the intent clearer.

PR #409, opened 2026-07-23
feat: add exponential backoff for restart policy with API and docs improvements

diw-zw:
## Summary

This PR addresses review feedback from #394 by improving the restart policy backoff feature across tests, bug fixes, documentation, and API validation:

- **Unit tests**: Add comprehensive unit test coverage for 4 core backoff functions (`syncRestartTrackingFromAPI`, `updateRestartTracking`,
`shouldRecreateInstance` bypass logic, `restartTrackingChanged`) and `checkRestartBackoff` branch paths that previously had zero test coverage
- **Bug fix**: Clear `restartingCache` on instance deletion (IsNotFound and DeletionTimestamp paths) to prevent stale cache entries from blocking newly
created instances with the same name
- **Examples**: Update all 7 v1alpha2 example YAMLs to use the new `restartPolicy` object format (`type:` field) instead of the old string format
- **API validation**: Add `+kubebuilder:default=30`/`+kubebuilder:default=600` annotations for `kubectl explain` and server-side apply support, plus CEL
 XValidation rule enforcing `maxDelaySeconds >= baseDelaySeconds`
- **Documentation**: Update `failure-handling.md` with Exponential Backoff section (formula, stability window, default behavior change, disable/custom
config), update `api.md` and best-practice docs to reflect the new object format
- **TODOs**: Mark two known issues with TODO comments — permanent `LastRestartTime` bypass, slow-path backoff bypasses rolling
update protections

## Changes

| Area | Files | Description |
|------|-------|-------------|
| Tests | `instance_scale_test.go`, `instance_status_test.go` | +811 lines of unit tests for backoff core functions |
| Bug fix | `instance_reconciler.go`, `sync/api.go` | Clear `restartingCache` on instance deletion |
| Examples | 7 YAML files | Migrate `restartPolicy` from string to object format |
| API | `rolebasedgroup_types.go`, 3 CRD YAMLs | Add kubebuilder defaults and CEL validation |
| Docs | `failure-handling.md`, `api.md`, 2 best-practice docs | Document backoff behavior and update field references |
| TODOs | `instance_scale.go` | Mark two issue for future fixes |

### DIFF
--- a/api/workloads/v1alpha2/rolebasedgroup_types.go
+++ b/api/workloads/v1alpha2/rolebasedgroup_types.go
@@ -160,6 +160,7 @@ const (
 )
 
 // RestartPolicyConfig groups restart policy type and backoff configuration.
+// +kubebuilder:validation:XValidation:rule="!has(self.baseDelaySeconds) || !has(self.maxDelaySeconds) || self.maxDelaySeconds >= self.baseDelaySeconds",message="maxDelaySeconds must be greater than or equal to baseDelaySeconds"
 type RestartPolicyConfig struct {
 	// Type defines the restart policy when pod failures happen.
 	// Default is RecreateRoleInstanceOnPodRestart.
@@ -174,12 +175,14 @@ type RestartPolicyConfig struct {
 	// Default is 30.
 	// +optional
 	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default=30
 	BaseDelaySeconds *int32 `json:"baseDelaySeconds,omitempty"`
 
 	// MaxDelaySeconds caps the exponential backoff delay (seconds).
 	// Default is 600.
 	// +optional
 	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default=600
 	MaxDelaySeconds *int32 `json:"maxDelaySeconds,omitempty"`
 }
 
--- a/pkg/reconciler/roleinstance/sync/instance_scale.go
+++ b/pkg/reconciler/roleinstance/sync/instance_scale.go
@@ -475,7 +475,7 @@ func (c *realControl) checkRestartBackoff(instance *workloadsv1alpha2.RoleInstan
 // The first backoff (restartCount=1) equals baseDelay, then doubles each round.
 // When maxDelaySeconds is 0, no cap is applied (unbounded backoff).
 func calculateRestartDelay(baseDelaySeconds, maxDelaySeconds, restartCount int32) int32 {
-	if baseDelaySeconds == 0 || restartCount <= 0 {
+	if baseDelaySeconds <= 0 || restartCount <= 0 {
 		return 0
 	}
 	// Use (restartCount-1) so the first backoff equals baseDelay.
--- a/pkg/reconciler/roleinstance/sync/instance_scale.go
+++ b/pkg/reconciler/roleinstance/sync/instance_scale.go
@@ -404,6 +404,12 @@ func (c *realControl) checkRestartBackoff(instance *workloadsv1alpha2.RoleInstan
 	// the instance status was just updated (Ready=False) from the previous
 	// reconcile, making wasInstanceReady() return false on the next reconcile,
 	// which would cause shouldRecreateInstance to skip the restart entirely.
+	// TODO: the slow path also applies backoff when shouldRecreateInstance
+	// returned false due to Generation != ObservedGeneration or
+	// CurrentRevision != UpdateRevision (rolling update / spec change in
+	// progress). This freezes updates until the backoff window expires.
+	// Fix: only apply slow-path backoff when shouldRecreateInstance returned
+	// false due to wasInstanceReady, not due to Generation/Revision mismatch.
 	if !shouldRecreateInstance(instance, allPods, instance.Status.InPlaceUpdateContainerBaselines) {
 		hasFailedPods := false
 		for _, p := range allPods {
@@ -546,8 +552,9 @@ func shouldRecreateInstance(instance *workloadsv1alpha2.RoleInstance, pods []*v1
 	// The instance may not yet be Ready (pods still Pending after recreation), so
 	// wasInstanceReady() would return false. We bypass to prevent deadlock where
 	// a pod crash before Ready would block recreation indefinitely.
-	// This bypass is bounded: RestartCount resets after a stable period
-	// (max(maxDelay*2, 10min)), and the normal Ready check resumes.
+	// TODO: this bypass is currently permanent — LastRestartTime is never cleared
+	// once set, so wasInstanceReady() never resumes after the first restart.
+	// Fix: clear LastRestartTime after a sustained Ready period (max(maxDelay*2, 10min)).
 	if (!wasInstanceReady(instance) && instance.Status.LastRestartTime == nil) ||
 		instance.Generation != instance.Status.ObservedGeneration {
 		return false
--- a/pkg/reconciler/roleinstance/sync/instance_scale.go
+++ b/pkg/reconciler/roleinstance/sync/instance_scale.go
@@ -476,7 +476,9 @@ func (c *realControl) checkRestartBackoff(instance *workloadsv1alpha2.RoleInstan
 // calculateRestartDelay computes the exponential backoff delay for a restart attempt.
 // Formula: min(baseDelay * 2^(restartCount-1), maxDelay)
 // The first backoff (restartCount=1) equals baseDelay, then doubles each round.
-// When maxDelaySeconds is 0, no cap is applied (unbounded backoff).
+// A maxDelaySeconds of 0 is treated as no cap, but CEL validation requires
+// maxDelaySeconds >= baseDelaySeconds, so this only occurs when both are 0
+// (backoff disabled) — the no-cap branch is effectively unreachable for valid input.
 func calculateRestartDelay(baseDelaySeconds, maxDelaySeconds, restartCount int32) int32 {
 	if baseDelaySeconds <= 0 || restartCount <= 0 {
 		return 0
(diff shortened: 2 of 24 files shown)


pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -404,6 +404,12 @@ func (c *realControl) checkRestartBackoff(instance *workloadsv1alpha2.RoleInstan
 	// the instance status was just updated (Ready=False) from the previous
 	// reconcile, making wasInstanceReady() return false on the next reconcile,
 	// which would cause shouldRecreateInstance to skip the restart entirely.
+	// TODO: the slow path also applies backoff when shouldRecreateInstance
```
cheyang: This slow path also runs when `shouldRecreateInstance` returned false because a spec change / rolling update is in progress (`Generation != ObservedGeneration`, or `CurrentRevision != UpdateRevision`) rather than because of a real crash. When that happens `checkRestartBackoff` still returns a non-zero delay, `Scale()` requeues with `requeueAfter`, and the update is frozen until the backoff window expires — up to `maxDelaySeconds`, growing exponentially.

I reproduced it in a unit test: Recreate policy + a recent `LastRestartTime` + a failed pod + `Generation != ObservedGeneration` → `checkRestartBackoff` returns >0, even though no recreation is going to happen this reconcile. Gating the slow path to the `wasInstanceReady` case only (as the TODO suggests) fixes it. Given it stalls updates, I'd track this as a real issue rather than leaving it as an inline TODO.

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -546,8 +552,9 @@ func shouldRecreateInstance(instance *workloadsv1alpha2.RoleInstance, pods []*v1
 	// The instance may not yet be Ready (pods still Pending after recreation), so
 	// wasInstanceReady() would return false. We bypass to prevent deadlock where
 	// a pod crash before Ready would block recreation indefinitely.
-	// This bypass is bounded: RestartCount resets after a stable period
-	// (max(maxDelay*2, 10min)), and the normal Ready check resumes.
+	// TODO: this bypass is currently permanent — LastRestartTime is never cleared
```
cheyang: Confirmed the bypass is permanent. `updateRestartTracking` resets `RestartCount` after the stability window but never clears `LastRestartTime`, so once it's set this guard bypasses `wasInstanceReady()` forever. In a unit test, a not-Ready instance whose only restart was an hour ago still returns true here, whereas the same instance with `LastRestartTime == nil` is correctly gated. So the "prevent deadlock before first Ready" bypass never re-arms after the first restart. Clearing `LastRestartTime` after a sustained Ready period (as the TODO notes) would restore the intended behavior — worth an issue too.

api/workloads/v1alpha2/rolebasedgroup_types.go
```diff
@@ -174,12 +175,14 @@ type RestartPolicyConfig struct {
...
 	// +optional
 	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default=30
 	BaseDelaySeconds *int32 `json:"baseDelaySeconds,omitempty"`
 
 	// MaxDelaySeconds caps the exponential backoff delay (seconds).
 	// Default is 600.
 	// +optional
 	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default=600
```
cheyang, line 185: Worth calling out a behavior change from this default combined with the new CEL rule above: with `maxDelaySeconds` defaulting to 600, a config that sets `baseDelaySeconds > 600` and omits `maxDelaySeconds` is now rejected at admission — defaulting fills 600, then `600 >= baseDelaySeconds` fails. Before this PR the same config was accepted and just silently capped at 600 at runtime.

Confirmed against the generated CRDs in envtest: `baseDelaySeconds: 700` with `maxDelaySeconds` omitted → `maxDelaySeconds must be greater than or equal to baseDelaySeconds`; omitting both defaults cleanly to 30/600. The tightening is reasonable (base > max was already nonsensical), but a one-line note in the PR/docs would help anyone who was relying on the old cap — they now need to set `maxDelaySeconds` explicitly.

cheyang (commented):
Took a close look at the restart-backoff changes and reproduced the findings locally (unit + envtest against the generated CRDs).

The `ClearRestarting`-on-delete fix checks out: after deletion the in-memory cache entry is gone, so a newly created same-name instance is no longer blocked by `isAlreadyRestarting`.

Two behavioral things I'd flag, both currently only inline TODOs — I think they deserve tracked issues since they affect update behavior, not just code hygiene:
- the `checkRestartBackoff` slow path can freeze rolling updates / spec changes;
- the `LastRestartTime` bypass in `shouldRecreateInstance` is permanent.

And one heads-up on a behavior change from the new `maxDelaySeconds` default. Details inline.

cheyang:
@gemini-code-assist /review

pkg/reconciler/roleinstance/sync/instance_scale.go
```diff
@@ -404,7 +404,16 @@ func (c *realControl) checkRestartBackoff(instance *workloadsv1alpha2.RoleInstan
...
 	// reconcile, making wasInstanceReady() return false on the next reconcile,
 	// which would cause shouldRecreateInstance to skip the restart entirely.
+	//
+	// However, when shouldRecreateInstance returned false due to a rolling
+	// update or spec change (Generation/Revision mismatch), we must NOT apply
+	// backoff — doing so would freeze updates until the backoff window expires.
 	if !shouldRecreateInstance(instance, allPods, instance.Status.InPlaceUpdateContainerBaselines) {
+		// Skip backoff when a spec change or rolling update is in progress
+		if instance.Generation != instance.Status.ObservedGeneration ||
+			instance.Status.CurrentRevision != instance.Status.UpdateRevision {
```
cheyang, line 414: This guard is the most integration-sensitive change in the PR: it makes checkRestartBackoff return 0 (no backoff) while a spec change or rolling update is in progress, so restart backoff no longer freezes an update. That's exactly the kind of controller behaviour that unit tests can't fully protect against a regression on, because it depends on how backoff, recreation, and the update path interleave across reconciles.

The unit coverage added here is genuinely strong (I read through TestCheckRestartBackoff_Branches, TestShouldRecreateInstanceBypass, TestUpdateRestartTracking, TestSyncRestartTrackingFromAPI, etc.), and the core restart-policy feature already has an e2e in test/e2e/testcase/v1alpha2/restart_policy_stability.go. But none of the new behaviours in this PR — backoff not freezing a rolling update/spec change, the hasRecentRestart-bounded Ready bypass, and ClearRestarting on delete — are exercised through the API surface in test/e2e/ or test/envtest/restart_policy. For a feat PR that changes controller reconcile behaviour, please add an integration test (envtest is acceptable here given the existing restart_policy envtest suite) that at least covers: instance enters restart backoff, then a spec/template update is applied, and the update is observed to proceed rather than being blocked until the backoff window elapses. If you believe an integration test genuinely isn't feasible for this change, say so and I'll route it to the maintainer rather than waiving it here.

api/workloads/v1alpha2/rolebasedgroup_types.go
```diff
@@ -160,6 +160,7 @@ const (
 )
 
 // RestartPolicyConfig groups restart policy type and backoff configuration.
+// +kubebuilder:validation:XValidation:rule="!has(self.baseDelaySeconds) || !has(self.maxDelaySeconds) || self.maxDelaySeconds >= self.baseDelaySeconds",message="maxDelaySeconds must be greater than or equal to baseDelaySeconds"
```
cheyang, line 163: With baseDelaySeconds/maxDelaySeconds now carrying kubebuilder defaults of 30/600, CRD defaulting runs before CEL validation, so both fields are effectively always present. That means a user who sets only baseDelaySeconds to a value above 600 (e.g. 900) and omits maxDelaySeconds will be rejected by this rule, because maxDelaySeconds defaults to 600 and 600 >= 900 is false. That's arguably correct (max must be >= base), but the failure is non-obvious from the user's perspective since they never set maxDelaySeconds. Is that the intended behaviour? If so, no change needed — but it may be worth a one-line note in doc/features/failure-handling.md so users know they must raise maxDelaySeconds when raising baseDelaySeconds past the default cap.

diw-zw: This explanation is already included in the failure-handling.md document：

> **Note**: `maxDelaySeconds` must be greater than or equal to `baseDelaySeconds` (enforced by CEL validation at admission). Because `maxDelaySeconds` defaults to 600, setting `baseDelaySeconds` above 600 without explicitly setting `maxDelaySeconds` will be rejected. If you need a base delay above 600s, set `maxDelaySeconds` to the same or higher value.

cheyang:
Re-reviewed after the latest commits — thanks for the quick turnaround. I verified the fixes locally (unit + envtest) and they hold up:

- **F1 (slow-path backoff freezing rolling updates)** — fixed. The early `return 0` when `Generation != ObservedGeneration` or `CurrentRevision != UpdateRevision` correctly lets spec changes/rolling updates proceed instead of being frozen by backoff. Confirmed with a test that a Recreate instance mid-spec-change + failed pod + recent restart now returns `0`, while a genuine crash on a converged instance still backs off.
- **F3 (permanent `wasInstanceReady` bypass)** — fixed. `hasRecentRestart()` bounds the bypass to the stability window; an hour-old `LastRestartTime` no longer bypasses the not-Ready guard, but a recent restart still does (deadlock prevention preserved). Keeping `LastRestartTime` as a historical record and sharing `stableThreshold()` reads cleanly.
- **F4** — the docs note in `failure-handling.md` is exactly the right call-out. Verified in envtest against the generated CRDs: `baseDelaySeconds: 700` with `maxDelaySeconds` omitted is rejected; omitting both defaults to 30/600.

One thing to flag before this is treated as a merge gate — the **new envtest suite is flaky under a full-suite run** (`15 Passed | 2 Failed` of 17), though both failing specs pass in isolation, so the production logic is fine:

- `RestartPolicy=None … should only replace the failed pod` fails with `409 Conflict: the object has been modified` on `Status().Update(pod)`. The controller mutates the pod concurrently — the status update needs `retry.RetryOnConflict` (or a re-`Get` before update). Same pattern applies to the other `Status().Update(pod)` sites in the suite.
- `Recreate … should recreate all component pods when worker fails in LeaderWorker pattern` hits the 60s `Eventually` timeout under full-suite CPU contention (both specs finish in ~30s when run alone). Bumping that spec's timeout should settle it.

Not a correctness issue, but worth fixing so CI doesn't flake. Everything else looks good to me.

cheyang (approved):
/lgtm
/approve

PR #410, opened 2026-07-24
chore: add rbg-inference-deploy agent skill and CLAUDE.md

Syspretor:
## Summary

Add an interactive deployment wizard skill for deploying LLM inference services on Kubernetes using RoleBasedGroup (RBG), along with a CLAUDE.md for project guidance.

### CLAUDE.md
Project-level guidance file for Claude Code, describing build commands, architecture, and development conventions for this repository.

### .claude/skills/rbg-inference-deploy/
An interactive 5-phase deployment wizard that guides users through end-to-end LLM inference service deployment via conversation.

**Phases:**
1. **Prerequisites check** - k8s cluster connectivity, RBG controller status, llmctl CLI availability
2. **Requirement collection** - namespace, engine, model, GPU node labels/taints, resource budget, SLO
3. **Deployment plan analysis** - memory estimation, SLO feasibility, Auto-Benchmark or Cookbook-based config
4. **YAML generation** - RBG YAML with roleTemplates/direct template selection, optional features (InPlaceIfPossible, ScalingAdapter, CoordinatedPolicy, Gang scheduling)
5. **Deploy & verify** - apply, smart Pod Pending diagnosis, smoke test, optional benchmark

**Key design decisions:**
- Default language English; auto-detects Chinese input and switches language for all responses
- Automatically detects GPU node taints before YAML generation; uses direct `template` (not `templateRef`) when taints exist, since `tolerations` are silently dropped by the RBG controller during `templateRef` patch conversion (verified in source code)
- Smart Pod diagnosis: maps scheduler error keywords (`untolerated taint`, `Insufficient nvidia.com/gpu`, etc.) to actionable fixes

## Files
```
CLAUDE.md
.claude/skills/rbg-inference-deploy/
├── SKILL.md               # Main wizard workflow (5 phases)
├── prerequisites.md       # Env check details and installation guides
├── deployment-analysis.md # GPU memory estimation, SLO reference, Auto-Benchmark flow
├── yaml-rules.md          # YAML generation rules with roleTemplates constraints
└── benchmark.md           # Benchmark guide using llmctl
```

### DIFF
--- a/.claude/skills/rbg-inference-deploy/SKILL.md
+++ b/.claude/skills/rbg-inference-deploy/SKILL.md
@@ -1,0 +1,369 @@
+---
+name: rbg-inference-deploy
+description: Interactive wizard for deploying LLM inference services on Kubernetes using RoleBasedGroup (RBG). Guides users through prerequisites checks (k8s cluster, RBG installation, llmctl CLI), deployment configuration (engine, model, GPU, SLO), deployment pattern analysis, RBG YAML generation, deployment, testing, and benchmarking. Use when users want to deploy LLM/AI inference services with RBG, need help creating RBG deployment YAML, or want to configure SGLang/vLLM inference on Kubernetes.
+---
+
+# RBG Inference Deployment Wizard / RBG 推理服务部署向导
+
+## 🌐 Language / 语言
+
+**Default language: English.**
+
+Language switching rules:
+- Detect user's language from each message
+- If the user writes in **Chinese** (contains Chinese characters 一-鿿), switch to **Chinese** for ALL subsequent responses
+- If the user writes in **English**, switch to **English** for ALL subsequent responses
+- Apply consistently to: questions, tables, status messages, YAML comments, error messages, and all user-facing output
+- You may explicitly say: `"Switching to Chinese / 切换为中文"` when language changes
+
+---
+
+## Overview / 概述
+
+**EN**: An interactive wizard that guides you through end-to-end deployment of LLM inference services on Kubernetes via conversation. Executes in 5 sequential phases.
+
+**ZH**: 一个交互式向导，帮助用户通过对话完成 LLM 推理服务在 Kubernetes 上的端到端部署。分为 5 个阶段依次执行。
+
+**Before starting**: Read [prerequisites.md](prerequisites.md), [deployment-analysis.md](deployment-analysis.md), [yaml-rules.md](yaml-rules.md), [benchmark.md](benchmark.md) for full context.
+
+---
+
+## Phase 1: Prerequisites Check / 前置环境检查
+
+> Details: see [prerequisites.md](prerequisites.md)
+
+### 1.1 Check Kubernetes Cluster / 检查 Kubernetes 集群连通性
+
+```bash
+kubectl cluster-info 2>/dev/null
+kubectl get nodes 2>/dev/null
+```
+
+**Result handling / 结果处理**:
+- **Success / 成功**: proceed to 1.2
+- **Failed / 失败**:
+  - EN: "No reachable k8s cluster detected. Switching to Dry Run mode — YAML will be generated but not deployed."
+  - ZH: "未检测到可用 k8s 集群，将以 Dry Run 模式生成 YAML，无法真实部署。"
+  - Record `DRY_RUN=true`, skip 1.2, proceed to Phase 2
+
+### 1.2 Check RBG Controller / 检查 RBG Controller 是否已安装
+
+```bash
+kubectl get deployment rbgs-controller-manager -n rbgs-system 2>/dev/null
+kubectl get crd rolebasedgroups.workloads.x-k8s.io 2>/dev/null
+```
+
+**Result handling / 结果处理**:
+- **Installed & Ready / 已安装且 Ready**: proceed to 1.3
+- **Not installed / 未安装**: prompt user, ask whether to install now
+  - **Yes / 是**: run install flow (see [prerequisites.md](prerequisites.md) § RBG Installation)
+  - **No / 否**: record `DRY_RUN=true`, proceed to 1.3
+
+### 1.3 Check llmctl / 检查 llmctl 是否已安装
+
+```bash
+which llmctl 2>/dev/null || llmctl --help 2>/dev/null
+```
+
+**Result handling / 结果处理**:
+- **Installed / 已安装**: record `LLMCTL_AVAILABLE=true`, proceed to Phase 2
+- **Not installed / 未安装**: show the prompt below (in detected language), ask whether to install:
+
+  **EN prompt**:
+  > `llmctl` is the RBG inference extension CLI offering:
+  > - **Model management**: Auto-download from HuggingFace/ModelScope to PVC
+  > - **Auto-Benchmark**: Automated parameter search + SLO evaluation to find optimal config (tp_size, max_num_seqs, etc.)
+  > - **Benchmark**: One-click load testing with visual Dashboard
+  > - **Service management**: One-click deploy/list/delete
+  >
+  > **Without it**: Model path must be provided manually; deployment params will be estimated via Cookbook instead of empirically tuned.
+
+  **ZH prompt**:
+  > `llmctl` 是 RBG 的推理服务扩展 CLI，提供以下能力：
+  > - **模型管理**：从 HuggingFace/ModelScope 自动下载模型到 PVC
+  > - **Auto-Benchmark**：自动化参数搜索 + SLO 评估，找到最优部署方案
+  > - **Benchmark**：一键对已部署服务进行压测，可视化结果 Dashboard
+  > - **服务管理**：一键 deploy/list/delete 推理服务
+  >
+  > **不安装的影响**：模型需用户自行提供 PVC 路径；部署参数将通过引擎 Cookbook 估算而非实测调优。
+
+  - **Install / 选择安装**: run install (see [prerequisites.md](prerequisites.md) § llmctl installation), record `LLMCTL_AVAILABLE=true`
+  - **Skip / 暂不安装**: record `LLMCTL_AVAILABLE=false`, proceed to Phase 2
+
+---
+
+## Phase 2: Deployment Requirements / 部署需求采集
+
+> Ask one item at a time, confirm before proceeding / 逐步询问用户，每项确认后再进入下一项
+
+### 2.1 Namespace
+
+Ask for the target namespace (default: `default`). If cluster is reachable, check whether it exists; if not, ask whether to create it.
+
+> ZH: 询问目标 namespace（默认 `default`）。若集群可用，检查是否存在，不存在则询问是否自动创建。
+
+```bash
+kubectl get namespace <ns> 2>/dev/null || kubectl create namespace <ns>
+```
+
+### 2.2 Inference Engine & Image / 推理引擎 & 镜像
+
+Ask which inference engine to use / 询问使用哪种推理引擎：
+- **SGLang**: default image `lmsysorg/sglang:latest`, see recommended versions in [deployment-analysis.md](deployment-analysis.md)
+- **vLLM**: default image `vllm/vllm-openai:latest`, see recommended versions in [deployment-analysis.md](deployment-analysis.md)
+
+If the provided image tag is outdated (below minimum recommended version), suggest upgrading but do not block. / 若用户提供的镜像 tag 明显过旧，输出升级建议但不强制。
+
+### 2.3 Model / 模型
+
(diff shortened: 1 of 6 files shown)


cheyang:
@gemini-code-assist /review

cheyang:
I reviewed this by testing the claims rather than reading them, since the skill is prose an agent executes verbatim and CI can't check any of it. Five of the factual claims about this repo don't hold; details inline.

The one I'd fix first is the `tolerations` / `templateRef` premise in yaml-rules.md — it's stated as verified in source, it's the sole justification for forbidding `roleTemplates` on tainted nodes, and both a unit test on `GetResolvedTemplate` and a live cluster run show tolerations do reach the Pod.

One thing that isn't a defect but is worth thinking about: nothing keeps this file honest as the repo moves. The Makefile targets, the CLI commands and the restartPolicy enum were all correct at some point in this project's history. If CLAUDE.md is going to describe build commands and API values, a small CI check that greps the documented `make` targets and enum values against the tree would stop the next drift.

Harness, tests and raw run logs: https://github.com/cheyang/rbg/blob/verify/pr410-deploy-skill/docs/verification/pr410-deploy-skill/README.md

.claude/skills/rbg-inference-deploy/yaml-rules.md
```diff
@@ -0,0 +1,577 @@
+# RBG YAML 生成规则
+
+## 核心原则
+
+1. **roleTemplates 仅适用于无 Taint 节点**：多角色共用相同镜像且目标节点无 Taint 时，才提取公共部分到 `spec.roleTemplates`，各角色通过 `templateRef` 引用并用 `patch` 差异化。
+2. **有 Taint 时必须用直接 `template`**：`tolerations` 字段**无法**通过 `templateRef.patch` 传递到 Pod（RBG controller 已知行为限制，已通过源码验证）。只要节点有 Taint，该角色必须使用直接 `template`，在 `spec.tolerations` 中显式写入容忍规则。
```
cheyang, line 6: I tried to reproduce this and couldn't. I think the premise is backwards, and it's load-bearing for the whole roleTemplates rule.

`GetResolvedTemplate` (api/workloads/v1alpha2/helper.go:315) runs a plain `strategicpatch.StrategicMergePatch` over the entire PodTemplateSpec, and `ConstructPodTemplateSpecApplyConfiguration` (pkg/reconciler/pod_reconciler.go:128) converts the result with the generic unstructured converter. There is no field allow-list on that path, and the sts / deploy / lws / roleinstanceset reconcilers all funnel through it, so nothing strips `tolerations`.

On a live cluster (k8s v1.36.1, controller v0.8.0-cea2a47), one RBG with two roles sharing one roleTemplate:

```
prefill-0   [{"key":"verify.rbg.io/pr410","operator":"Equal","value":"from-patch","effect":"NoSchedule"}]
decode-0    []      # same base template, patch: {}
```

The only place `prefill` could have gotten that toleration from is `templateRef.patch`. Both pods reached Running.

There is a real caveat worth documenting in its place: `corev1.PodSpec.Tolerations` carries no `patchMergeKey`, so a patched `tolerations` list *replaces* the base list instead of merging into it. Replace, not drop.

I'd treat this as blocking rather than a nit because of the blast radius: cloud GPU node pools are almost always tainted, so as written the wizard will never use `roleTemplates` in the environment it targets, and every user ends up with the same inline template duplicated per role. The same reasoning applies to the `affinity` warning at line 577 and the taint branch at lines 173-176.

Unit tests and the full run log: https://github.com/cheyang/rbg/blob/verify/pr410-deploy-skill/docs/verification/pr410-deploy-skill/README.md

.claude/skills/rbg-inference-deploy/SKILL.md
```diff
@@ -0,0 +1,369 @@
...
+> YAML generation rules: see [yaml-rules.md](yaml-rules.md)
+
+### 4.1 Generate Base YAML / 生成基础 YAML
+
+Generate RBG YAML based on confirmed deployment plan, following these rules / 根据确认的部署方案生成 RBG YAML，遵循以下规则：
+
+**roleTemplates conditions** (all must be met) / **roleTemplates 适用条件**（全部满足才使用）：
+- Multiple roles share the same image **and** / 多角色共用相同镜像 **且**
+- **No Taints on target nodes** (`NODE_TAINTS` is empty) / **目标节点无 Taint**（`NODE_TAINTS` 为空）
+- If Taints exist, **must use direct `template`** (not `templateRef`) — tolerations are silently dropped by RBG controller when going through templateRef (verified in source code)
```
cheyang, line 230: Same as my comment on yaml-rules.md:6 — `tolerations` do survive `templateRef.patch`, checked at the helper level and against a real cluster. This is the line the agent actually acts on, so it needs to go too, along with the `NODE_TAINTS` is empty precondition above it.

.claude/skills/rbg-inference-deploy/SKILL.md
```diff
@@ -0,0 +1,369 @@
...
+- Look up recommended config by engine/model/GPU type
+
+### 3.4 Plan Confirmation / 方案确认
+
+Display deployment plan summary in a table; user can adjust via conversation until confirmed / 以表格形式展示最终部署方案，用户可调整直到确认：
+
+```
+┌─────────────────────────────────────────────────────────────┐
+│                      部署方案摘要                            │
+```
```
cheyang, line 200: This fence closes the summary box early, and then the fence at line 215 re-opens instead of closing. The file ends up with 27 fences — an odd number — so fence parity is inverted from here to EOF.

Running the file through `marked`, three headings vanish into a code block: `Phase 4: RBG YAML Generation`, `4.1 Generate Base YAML`, `4.2 Configure Extra Features`. That's the YAML-generation part of the wizard, i.e. the section that matters most.

Deleting this one line fixes the whole file. `markdown-check` in CI doesn't catch it.

CLAUDE.md
```diff
@@ -0,0 +1,162 @@
...
+- `RecreateRBGOnPodRestart`: Entire RBG recreation on pod failure
+- `RecreateRoleInstanceOnPodRestart`: Role instance recreation on pod failure
+
+### Gang Scheduling
+Integrated with Volcano and scheduler-plugins for coordinated pod scheduling across roles.
+
+## CLI (kubectl-rbg)
+
+The CLI is in `cmd/cli/` and provides LLM-focused commands:
+- `kubectl rbg llm config init`: Initialize configuration
```
cheyang, line 131: None of these five are reachable today. `llm.NewLLMCmd` exists in cmd/cli/cmd/llm/llm.go but is never attached to the root command — cmd/cli/cmd/root.go:73-74 registers only `status` and `rollout`. Building the CLI and asking it:

```
Available Commands:
  completion  Generate the autocompletion script for the specified shell
  help        Help about any command
  rollout     Manage the rollout of a rbg object
  status      Display rbg status information
```

`config`, `model` and `svc` don't exist in the tree at all; `llm benchmark` does exist but can't be invoked. The unreachable `llm` tree looks like a pre-existing CLI bug rather than something this PR caused — probably worth its own issue — but the doc shouldn't ship pointing at it.

CLAUDE.md
```diff
@@ -0,0 +1,162 @@
...
+Roles can declare `dependencies` on other roles, ensuring ordered startup. The controller waits for dependency roles to be ready before starting dependent roles.
+
+### Rollout Strategy
+- `RollingUpdate`: Incremental pod replacement
+- Supports `partition`, `maxUnavailable`, `maxSurge`
+- In-place update strategies: `RecreatePod`, `InPlaceIfPossible`, `InPlaceOnly`
+
+### Restart Policy
+- `None`: Default pod restart behavior
+- `RecreateRBGOnPodRestart`: Entire RBG recreation on pod failure
```
cheyang, line 122: The API server rejects this value:

```
spec.roles[0].leaderWorkerPattern.restartPolicy: Unsupported value: "RecreateRBGOnPodRestart": supported values: "None", "RecreateRoleInstanceOnPodRestart"
```

The Go constant is still there and pod_controller.go:283 still compares against it, which is probably why it looks valid, but the kubebuilder enum only allows the other two. keps/inactive-pod-handling/README.md:453 records it as deprecated by #340.

Separately, this section reads as if `roles[].restartPolicy` were a field. In v1alpha2 it isn't — it lives under `leaderWorkerPattern` / `customComponentsPattern`. Worth saying so, since an agent following this will put it one level too high.

CLAUDE.md
```diff
@@ -0,0 +1,162 @@
...
+## Project Overview
+
+RoleBasedGroup (RBG) is a Kubernetes API for orchestrating distributed, stateful AI inference workloads with multi-role collaboration and built-in service discovery. It treats an inference service as a role-based group — a topologized, stateful, coordinated multi-role organism managed as a single unit.
+
+## Build and Development Commands
+
+### Building
+```bash
+make build          # Build controller binary (outputs to bin/manager)
+make build-cli      # Build CLI binary (outputs to bin/kubectl-rbg)
```
cheyang, line 14: Agreeing with Copilot here after checking it myself: neither target exists. Worth adding that no Makefile target builds `cmd/cli` at all right now, so there isn't an alternative command to point the reader at unless the targets get added.

.claude/skills/rbg-inference-deploy/yaml-rules.md
```diff
@@ -0,0 +1,577 @@
...
+3. **templateRef.patch 必须设置**：即使无覆盖也需设置为 `patch: {}`。
```
cheyang, line 7: Flagging that the Copilot comment on this line is wrong and rule 3 as written is correct — I wouldn't relax it. `Patch` is optional in the Go type, but the controller rejects `templateRef` without it (api/workloads/v1alpha2/roletemplate_validation.go:139). Applying such an RBG on a live cluster gets you an `InvalidTemplateRef` event: `spec.roles[0].templatePatch: required when templateRef is set`.

.claude/skills/rbg-inference-deploy/yaml-rules.md
cheyang, line 6: I dug into this against the controller and I don't think the claim holds. When a role uses templateRef, the patch is resolved through GetResolvedTemplate -> applyStrategicMergePatch (api/workloads/v1alpha2/helper.go:356-410), which runs strategicpatch.StrategicMergePatch over a full PodTemplateSpec. So spec.tolerations included in the patch is merged onto the pod template, not silently dropped. Because this "verified in source code" claim also drives the forbidden-fields list below (line 574) and the 'must use direct template when taints exist' rule, it would push users away from a configuration path that actually works. Could you re-check against the current controller and soften or correct the wording?

.claude/skills/rbg-inference-deploy/SKILL.md
cheyang, line 230: This mirrors the tolerations note in yaml-rules.md, and the wizard treats it as a hard rule ('tolerations are silently dropped by RBG controller when going through templateRef (verified in source code)'). Since the patch path applies a strategic merge patch to the whole PodTemplateSpec, spec.tolerations is honored. Worth updating here too so the wizard doesn't force a direct template unnecessarily.

.claude/skills/rbg-inference-deploy/prerequisites.md
```diff
@@ -0,0 +1,166 @@
...
+```
+
+---
+
+## llmctl 安装
+
+### 方式一：从源码构建（推荐）
+
+```bash
+git clone https://github.com/rolebasedgroup/inference-ext-cli.git
```
cheyang, line 48: Can you confirm this clone URL is correct? The install flow depends on `github.com/rolebasedgroup/inference-ext-cli`, but the rest of the repo lives under the `sgl-project` org, so I want to make sure contributors won't hit a dead link when following the from-source install path.

.claude/skills/rbg-inference-deploy/SKILL.md
```diff
@@ -0,0 +1,369 @@
...
+## 🌐 Language / 语言
+
+**Default language: English.**
+
+Language switching rules:
+- Detect user's language from each message
+- If the user writes in **Chinese** (contains Chinese characters 一-鿿), switch to **Chinese** for ALL subsequent responses
+- If the user writes in **English**, switch to **English** for ALL subsequent responses
+- Apply consistently to: questions, tables, status messages, YAML comments, error messages, and all user-facing output
+- You may explicitly say: `"Switching to Chinese / 切换为中文"` when language changes
```
cheyang, line 17: Small inconsistency: this suggests emitting a bilingual notice ("Switching to Chinese / 切换为中文"), while execution note 5 (line 369) says never to mix languages within a single response. Making the switch notice single-language would keep the two rules aligned.

cheyang (approved):
/lgtm
/approve

PR #411, opened 2026-07-24
test: strengthen regression coverage for restart-policy exponential backoff

cheyang:
Strengthens regression coverage for the exponential-backoff restart policy added in #394. **Test-only — no production changes.**

This revision folds the coverage into the existing #394 tests rather than adding parallel files, per review feedback:

## `TestCalculateRestartDelay` (`pkg/reconciler/roleinstance/sync/instance_scale_test.go`)
- Adds boundary rows beyond the int64-overflow guard: capped `63/64/100` and uncapped `63/64/100` (saturating at int32-max), extending the existing table which stopped at `62`/uncapped-`60`.

## `test/envtest/testcase/restart_policy/restart_policy_test.go`
- **`should reset RestartCount after stable period`**: now seeds `RestartCount=5` before the stable-period reset, so it exercises the reset when the counter has climbed above 1. A count of 1 alone can't distinguish a real reset from a monotonic "keep the max".
- **New `should reject a negative baseDelaySeconds on both RoleBasedGroup and RoleInstance`**: asserts an `Invalid` (422) with a `FieldValueInvalid` cause on a `baseDelaySeconds` field — so an unrelated admission/schema failure can't make the test pass while the `Minimum=0` validation is silently gone.

Verified locally against merged `main`: unit table pass; full `restart_policy` envtest suite pass; golangci-lint v1.63.4 clean.

### DIFF
--- a/test/envtest/testcase/restart_policy/restart_policy_test.go
+++ b/test/envtest/testcase/restart_policy/restart_policy_test.go
@@ -17,13 +17,18 @@ limitations under the License.
 package restart_policy
 
 import (
+	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/rbgs/api/workloads/constants"
 	workloadsv1alpha2 "sigs.k8s.io/rbgs/api/workloads/v1alpha2"
@@ -1067,9 +1072,16 @@ var _ = Describe("RestartPolicy Controller Integration", func() {
 			// Simulate a long stable period by backdating LastRestartTime to 11 minutes ago.
 			// The reset threshold is max(maxDelaySeconds*2, 10min) = max(4, 600) = 600s = 10min.
 			// Setting LastRestartTime to 11 minutes ago exceeds this threshold.
+			//
+			// Also seed RestartCount to a value >1: the reset must apply regardless of how
+			// high the counter has climbed. A count of 1 alone would not distinguish a real
+			// reset from a monotonic "keep the max" of the old and new value.
 			ri := getRoleInstance(rbgName, roleName)
 			Expect(ri).NotTo(BeNil())
-			ri.Status.LastRestartTime.Time = time.Now().Add(-11 * time.Minute)
+			ri.Status.RestartCount = 5
+			// Overwrite the pointer rather than mutating LastRestartTime.Time, which
+			// would panic if the field were ever nil.
+			ri.Status.LastRestartTime = &metav1.Time{Time: time.Now().Add(-11 * time.Minute)}
 			Expect(testutil.K8sClient.Status().Update(testutil.Ctx, ri)).Should(Succeed())
 
 			// Trigger another recreation
@@ -1092,8 +1104,9 @@ var _ = Describe("RestartPolicy Controller Integration", func() {
 				return len(active) >= 1
 			}, timeout, interval).Should(BeTrue())
 
-			// RestartCount should be reset to 1 (not 2), because the stable period
-			// caused the counter to reset before incrementing.
+			// RestartCount should be reset to 1 (not 6), because the stable period
+			// causes the counter to reset before incrementing — even though it was
+			// seeded to 5.
 			Eventually(func() int32 {
 				ri := getRoleInstance(rbgName, roleName)
 				if ri == nil {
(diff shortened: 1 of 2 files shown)


test/envtest/testcase/restart_policy/restart_policy_test.go
```diff
@@ -1067,8 +1072,13 @@ var _ = Describe("RestartPolicy Controller Integration", func() {
...
 			// Setting LastRestartTime to 11 minutes ago exceeds this threshold.
+			//
+			// Also seed RestartCount to a value >1: the reset must apply regardless of how
+			// high the counter has climbed. A count of 1 alone would not distinguish a real
+			// reset from a monotonic "keep the max" of the old and new value.
 			ri := getRoleInstance(rbgName, roleName)
 			Expect(ri).NotTo(BeNil())
+			ri.Status.RestartCount = 5
 			ri.Status.LastRestartTime.Time = time.Now().Add(-11 * time.Minute)
 			Expect(testutil.K8sClient.Status().Update(testutil.Ctx, ri)).Should(Succeed())
```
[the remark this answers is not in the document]
cheyang: Addressed in `1685cd12`: the field is now overwritten with a fresh pointer (`ri.Status.LastRestartTime = &metav1.Time{Time: ...}`) instead of mutating `.Time`, so it is panic-safe even if the pointer were nil.

test/envtest/testcase/restart_policy/restart_policy_test.go
```diff
@@ -1092,16 +1102,84 @@ var _ = Describe("RestartPolicy Controller Integration", func() {
...
+				matched := false
+				for _, c := range se.ErrStatus.Details.Causes {
+					if c.Type == metav1.CauseTypeFieldValueInvalid && strings.HasSuffix(c.Field, "baseDelaySeconds") {
+						matched = true
+						break
+					}
+				}
+				ExpectWithOffset(1, matched).To(BeTrue(),
+					"%s: expected a FieldValueInvalid cause on a baseDelaySeconds field, got causes %+v",
+					subject, se.ErrStatus.Details.Causes)
```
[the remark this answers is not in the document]
cheyang: Addressed in `1685cd12`: the helper now binds `details := se.ErrStatus.Details` and asserts `Expect(details).NotTo(BeNil())` before ranging over `details.Causes`; the failure message references the same checked variable, so a nil `Details` yields a clear assertion failure instead of a panic.

cheyang:
@gemini-code-assist /review

test/envtest/testcase/restart_policy/restart_policy_test.go
```diff
@@ -1092,16 +1104,87 @@ var _ = Describe("RestartPolicy Controller Integration", func() {
...
+						MaxDelaySeconds:  ptr.To(int32(600)),
+					},
+					Components: []workloadsv1alpha2.RoleInstanceComponent{
+						{
+							Name: "main",
+							Size: ptr.To(int32(1)),
+							Template: corev1.PodTemplateSpec{
+								Spec: corev1.PodSpec{
+									Containers: []corev1.Container{
+										{Name: "c", Image: "registry.cn-hangzhou.aliyuncs.com/acs-sample/nginx:latest"},
```
cheyang: The RoleBasedGroup arm of this case goes through the wrappers (BuildLeaderWorkerRole -> BuildBasicPodTemplateSpec), but the RoleInstance arm hand-rolls a PodTemplateSpec with the image string spelled out inline. If the default test image ever changes, this literal silently drifts out of sync with everything else. Since the assertion is purely about the Minimum=0 rejection on baseDelaySeconds, the pod template here is incidental — pulling it from the same basic-template helper (or the v1alpha2 RoleInstance wrapper) would keep it consistent and drop the duplicated string. Non-blocking.

test/envtest/testcase/restart_policy/restart_policy_test.go
```diff
@@ -1092,16 +1104,87 @@ var _ = Describe("RestartPolicy Controller Integration", func() {
...
+									},
+								},
+							},
+						},
```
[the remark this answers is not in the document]
cheyang: Addressed in `2ee2cc7c`: the RoleInstance arm now uses `wrappersv2.BuildBasicPodTemplateSpec()` (`test/wrappers/v1alpha2/role_wrapper.go:233`), which sources the image from `utils.DefaultImage` — the same path the RoleBasedGroup arm already took. The duplicated literal is gone.

test/envtest/testcase/restart_policy/restart_policy_test.go
cheyang: Addressed in `2ee2cc7c`: replaced with `Template: wrappersv2.BuildBasicPodTemplateSpec()`, so both arms of the case now go through the same helper and the inline image string is gone.

Verified locally against the branch: focused spec passes, and the full `restart_policy` envtest suite (16 specs) passes in 158s. `go vet` and `gofmt` clean.

cheyang:
@gemini-code-assist /review

cheyang:
Re-review of the new commit (2ee2cc7). The only change since the last round swaps the hand-rolled RoleInstance pod template for the shared wrappersv2.BuildBasicPodTemplateSpec() helper, which resolves the duplicated-image nit from Round 1. The refactor is behaviour-preserving for envtest and does not weaken any of the backoff regression assertions. CI is fully green on the head SHA. Looks ready to merge.

test/envtest/testcase/restart_policy/restart_policy_test.go
```diff
@@ -1092,16 +1104,81 @@ var _ = Describe("RestartPolicy Controller Integration", func() {
...
+					RestartPolicy: workloadsv1alpha2.RestartPolicyConfig{
+						Type:             workloadsv1alpha2.RecreateRoleInstanceOnPodRestart,
+						BaseDelaySeconds: ptr.To(int32(-30)),
+						MaxDelaySeconds:  ptr.To(int32(600)),
+					},
+					Components: []workloadsv1alpha2.RoleInstanceComponent{
+						{
+							Name:     "main",
+							Size:     ptr.To(int32(1)),
+							Template: wrappersv2.BuildBasicPodTemplateSpec(),
```
cheyang, line 1176: Thanks for folding this into wrappersv2.BuildBasicPodTemplateSpec() - that clears up the duplicated inline pod template/image I flagged last round. The substitution is safe for the envtest here since pods are never actually scheduled and none of the assertions in this case depend on the container name or image; the survivor-not-deleted logic (backdated LastRestartTime, pods forced to Pending, hasRecentRestart=false) is untouched, so coverage is unchanged.

test/envtest/testcase/restart_policy/restart_policy_test.go
```diff
@@ -1092,16 +1104,81 @@ var _ = Describe("RestartPolicy Controller Integration", func() {
...
+						WithBaseDelaySeconds(-30).
+						Obj(),
+				}).Obj()
+			expectBaseDelayRejected(testutil.K8sClient.Create(testutil.Ctx, badRBG), "RoleBasedGroup")
+
+			// (b) RoleInstance spec field (same shared RestartPolicyConfig).
+			badRI := &workloadsv1alpha2.RoleInstance{
+				ObjectMeta: metav1.ObjectMeta{Name: "neg-base-ri", Namespace: testNs},
+				Spec: workloadsv1alpha2.RoleInstanceSpec{
+					RestartPolicy: workloadsv1alpha2.RestartPolicyConfig{
```
diw-zw, line 1167: Please rebase and address the structural changes to the restartPolicy field.

PR #413, opened 2026-07-27
feat: add --enable-legacy-workloads flag to conditionally disable leg…

diw-zw:
The RBG controller previously managed pods indirectly through Deployment,
StatefulSet, and LeaderWorkerSet resources. This required broad RBAC
permissions for those resource types. Now that RoleInstance/RoleInstanceSet
is the primary workload management mechanism, these legacy resources are
only needed for backward compatibility.

Add a `--enable-legacy-workloads` controller flag (default: false) that
controls whether the controller watches and reconciles Deployment,
StatefulSet, and LeaderWorkerSet resources. When disabled, the controller
only manages RoleInstanceSet-based workloads, reducing the required RBAC
footprint.

In the Helm chart, add `controller.features.legacyWorkloads.enabled`
(default: true) which:
- Conditionally includes RBAC rules for deployments, statefulsets, and
  leaderworkersets in the ClusterRole
- Passes the corresponding flag to the controller manager

The controllerrevisions RBAC permission is always retained since it is
used for revision management regardless of legacy workload support.

The Helm chart's clusterrole.yaml is now manually maintained (instead of
being auto-generated from config/rbac/role.yaml via the Makefile) to
support the conditional inclusion of legacy workload permissions.

### DIFF
--- a/cmd/rbgs/main.go
+++ b/cmd/rbgs/main.go
@@ -149,6 +149,8 @@ func main() {
 		// Pprof profiling
 		enablePprof bool
 		pprofAddr   string
+		// Legacy workloads (Deployment, StatefulSet, LeaderWorkerSet) support
+		enableLegacyWorkloads bool
 	)
 	flag.StringVar(
 		&metricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. "+
@@ -207,6 +209,13 @@ func main() {
 	)
 	flag.BoolVar(&enablePprof, "enable-pprof", false, "Enable pprof profiling server for performance debugging.")
 	flag.StringVar(&pprofAddr, "pprof-bind-address", ":6060", "The address the pprof endpoint binds to.")
+	flag.BoolVar(
+		&enableLegacyWorkloads, "enable-legacy-workloads", false,
+		"Enable legacy workload types (Deployment, StatefulSet, LeaderWorkerSet). "+
+			"Set to false to disable watching and reconciling these workload types, "+
+			"reducing required RBAC permissions. Only RoleInstanceSet-based workloads "+
+			"will be managed when disabled.",
+	)
 
 	// Register logger flags before Parse so that zap flags (--zap-log-level etc.) are recognized.
 	opts := zap.Options{
@@ -318,7 +327,7 @@ func main() {
 	restConfig.Burst = kubeAPIBurst
 
 	mgr, err := ctrl.NewManager(
-		restConfig, newManagerOptions(webhookMode, webhookServer, metricsServerOptions, probeAddr, enableLeaderElection),
+		restConfig, newManagerOptions(webhookMode, webhookServer, metricsServerOptions, probeAddr, enableLeaderElection, enableLegacyWorkloads),
 	)
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
@@ -352,7 +361,7 @@ func main() {
 	// controller (for recording and injection during reconcile).
 	nodeBindings := instancesync.NewNodeBindingStore()
 
-	rbgReconciler, err := workloadscontroller.NewRoleBasedGroupReconciler(mgr, scheduler.SchedulerPluginType(schedulerName), nodeBindings)
+	rbgReconciler, err := workloadscontroller.NewRoleBasedGroupReconciler(mgr, scheduler.SchedulerPluginType(schedulerName), nodeBindings, enableLegacyWorkloads)
 	if err != nil {
 		setupLog.Error(err, "unable to create rbg controller", "controller", "RoleBasedGroup")
 		os.Exit(1)
@@ -497,7 +506,7 @@ func newWebhookServer(webhookMode string, tlsOpts []func(*tls.Config)) webhook.S
 // newManagerOptions builds the controller-runtime manager options.
 // When webhooks are disabled (enable-webhooks=none), webhook server and leader
 // election are disabled, and metrics are served insecurely.
-func newManagerOptions(webhookMode string, webhookServer webhook.Server, metricsOpts metricsserver.Options, probeAddr string, enableLeaderElection bool) ctrl.Options {
+func newManagerOptions(webhookMode string, webhookServer webhook.Server, metricsOpts metricsserver.Options, probeAddr string, enableLeaderElection bool, enableLegacyWorkloads bool) ctrl.Options {
 	opts := ctrl.Options{
 		Scheme:                 scheme,
 		Metrics:                metricsOpts,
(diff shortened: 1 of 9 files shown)


internal/controller/workloads/rolebasedgroup_controller.go
```diff
@@ -607,6 +613,23 @@ func (r *RoleBasedGroupReconciler) getOrCreateWorkloadReconciler(
...
+		switch workloadType {
+		case constants.DeploymentWorkloadType,
+			constants.StatefulSetWorkloadType,
+			constants.LeaderWorkerSetWorkloadType:
+			return nil, fmt.Errorf(
+				"legacy workload type %s is not supported when legacy workloads are disabled",
+				workloadType,
+			)
+		}
+	}
```
cheyang, line 631: The reject itself looks right, but where it lands makes it whole-group rather than per-role. `getOrCreateWorkloadReconciler` is called from `preCheck` in a loop over roles, and the errors are folded into a `kerrors.NewAggregate`. `Reconcile` calls `preCheck` first and returns on error, so one legacy role short-circuits everything after it: `constructAndUpdateRoleStatuses`, `reconcileRoles`, and `deleteOrphanRoles` never run. An RBG that mixes a legacy role with RoleInstanceSet roles loses the healthy roles too, its status is never written, and it just requeues with backoff.

I ran the PR binary out-of-cluster against a 3-node cluster with `--enable-legacy-workloads=false`, authenticated as `rbgs-controller-sa` so the reduced ClusterRole was actually enforced. With an RBG holding one `apps/v1/Deployment` role and one RoleInstanceSet role, over 60s: 28 `not supported when legacy workloads are disabled` lines, 14 `Failed to create workload reconciler`, 14 `Reconciler error`, and neither a Deployment nor a RoleInstanceSet was created. No `Forbidden`, so the reject is doing its job of avoiding the doomed writes — the healthy role is just collateral damage.

Worth weighing who this lands on: `RoleWorkloadTypeAnnotationKey` is documented in `api/workloads/constants/annotation.go` as primarily set by the v1alpha1 to v1alpha2 conversion webhook, so the RBGs carrying legacy workload types are largely the backward-compat users this flag exists to serve.

To be clear about what I am and am not asking for: rejecting the whole group is defensible — roles are reconciled in dependency order and a RoleBasedGroup is a co-scheduled unit, so partially reconciling one would be worse. What I'd change is how the failure surfaces. Nothing reaches `.status` (the whole status update is skipped, not just the offending role), the reason lives only in an event that expires, and the RBG then retries indefinitely for what is really a configuration error. Rejecting at admission would also let users find out at apply time instead of afterwards. I've put the details in a reply below.

cmd/rbgs/main.go
```diff
@@ -207,6 +209,13 @@ func main() {
 	)
 	flag.BoolVar(&enablePprof, "enable-pprof", false, "Enable pprof profiling server for performance debugging.")
 	flag.StringVar(&pprofAddr, "pprof-bind-address", ":6060", "The address the pprof endpoint binds to.")
+	flag.BoolVar(
+		&enableLegacyWorkloads, "enable-legacy-workloads", false,
```
cheyang, line 213: Agreeing with the earlier comment on this line, with a couple of specifics.

The default registered here is `false`, while `deploy/helm/rbgs/values.yaml` (`legacyWorkloads.enabled: true`), the chart README values table, and the PR description all state `true`. Both shipped install paths do pass the flag explicitly — the Helm template, and `config/manager/manager.yaml` now that `deploy/kubectl/manifests.yaml` has been regenerated — so nothing visibly breaks today. The exposure is everything else: a downstream chart, a hand-written manifest, or running the binary directly all silently lose legacy workload reconciliation.

Go's `flag` package also omits `(default false)` for a false bool, so `--help` gives no hint either way. Registering `true` makes the safe behavior independent of every caller remembering to pass the flag.

cheyang (changes requested):

internal/controller/workloads/rolebasedgroup_controller.go
cheyang: Following up with a concrete shape, since the places that would need to change are outside this diff and I can't anchor a comment on them.

Two corrections to what I wrote above first.

**On the suggestion itself.** I proposed degrading per role — skip the legacy role, keep reconciling the rest. I'm withdrawing that; failing the whole group is the right call here, and I had it backwards. `reconcileRoles` processes roles in dependency order and gates on readiness:

```
:467  // Process roles in dependency order
:469  sortedRoles, err := dependencyManager.SortRoles(ctx, rbg)
:484  ready, err := dependencyManager.CheckDependencyReady(roleCtx, rbg, role)
:490  err := fmt.Errorf("dependencies not met for role '%s'", role.Name)
```

So if the skipped legacy role is something another role declares in `dependencies`, per-role degradation produces a group stuck forever on `dependencies not met` — and that state is indistinguishable from a normal startup wait, which is worse to debug than a clean failure. Beyond the topology, a RoleBasedGroup is a co-scheduled unit; half of one usually means endpoints exist while the actual serving path is broken, and GPUs are held by something that cannot serve.

**On the status field.** I referred to "a role status condition", but `RoleStatus` (`rolebasedgroup_types.go:462`) only carries `Name`, `ReadyReplicas`, `Replicas` and `UpdatedReplicas` — no conditions field. `RoleBasedGroupStatus.Conditions` (`:450`) is the right place for a group-level signal.

So the direction here is fine. What I'd still change is how the failure surfaces, because right now it fails quietly:

1. **Nothing reaches `.status`.** `constructAndUpdateRoleStatuses` (~`:662`) returns on the first legacy role, so the whole `.status` goes unwritten — not just the offending role's. The reason exists only in an event, which expires, so `kubectl get rbg` shows nothing explaining why the group is stuck. A terminal condition on `RoleBasedGroupStatus.Conditions` (e.g. reason `LegacyWorkloadsDisabled`, message listing the roles) would make this inspectable after the fact.

2. **It retries forever.** A disabled workload type is a configuration error, not a transient one, but returning an error from `Reconcile` puts the RBG into indefinite exponential backoff — I measured 14 reconcile attempts in 60s on an idle cluster, all identical. Worth short-circuiting until `spec`/`generation` changes.

3. **`deleteOrphanRoles` (~`:573`) is unreachable** behind the same early return, which also blocks the legacy cleanup path.

**And this makes admission validation the more important half**, not an optional extra as I first framed it. If the semantics are fail-fast, users should learn at `kubectl apply` time that the RBG is unusable under this controller's configuration, rather than having it accepted and then silently wedge. There's already a hook — `RoleBasedGroupValidator` (`rolebasedgroup_admission.go:35`, registered at `rolebasedgroup_webhook.go:28`), and the webhook shares the process with the controller so the flag can be injected directly:

```go
WithValidator(&RoleBasedGroupValidator{
    Client:                mgr.GetClient(),
    EnableLegacyWorkloads: enableLegacyWorkloads,
})
```

Keep an escape hatch in `ValidateUpdate` though: an existing legacy RBG must stay editable, or users can't change the annotation to migrate off it. Strict on CREATE, lenient on an UPDATE that removes legacy, would cover it. Admission alone won't help the RBGs already in the cluster — the converted-from-v1alpha1 ones — which is why the status/backoff points above still matter for them.

**One test worth adding either way:** a mixed RBG with the flag off, asserting the failure is visible in `.status` and doesn't spin. `test/envtest/testutil/setup.go:205` hardcodes `true`, so the disabled path has no coverage at all right now.

PR #414, opened 2026-07-28
feat: add --enable-v1alpha1-compat flag to toggle v1alpha1 API compatibility

diw-zw:
Add a --enable-v1alpha1-compat flag (default true) to toggle support for v1alpha1 API compatibility mode. When disabled, the controller only manages RoleInstanceSet-based workloads and no longer reconciles, watches, or caches v1alpha1 indirect workload types (Deployment,
StatefulSet, LeaderWorkerSet), reducing the required RBAC footprint.

- CLI flag + Helm value: --enable-v1alpha1-compat / controller.features.v1alpha1Compat.enabled (default true). Threaded through manager cache options, reconciler construction, and webhook setup.
- Gated RBAC: Helm ClusterRole rules for Deployment/StatefulSet/LeaderWorkerSet are conditionally included based on the flag. Controller-gen RBAC output is no longer auto-overwritten into the Helm chart.
- Gated watches & cleanup: SetupWithManager only registers Owns for StatefulSet/Deployment/LeaderWorkerSet when compat is enabled. deleteOrphanRoles skips cleanup of those types when disabled.
- Admission validation: ValidateCreate rejects RBGs with v1alpha1 indirect workload types when compat is disabled (fail-fast at kubectl apply time). ValidateUpdate only rejects newly-introduced types — existing RBGs can still be edited to migrate to RoleInstanceSet.
- Terminal status condition: handleLegacyWorkloads runs before preCheck in Reconcile. When compat is disabled and v1alpha1 indirect types are present, it sets Ready=False with reason LegacyWorkloadsDisabled, patches .status, emits a warning event, and returns without requeue —
preventing indefinite exponential backoff for a configuration error.
- Unit tests: Added rolebasedgroup_validation_test.go covering ValidateWorkloadTypes (CREATE) and ValidateWorkloadTypesUpdate (UPDATE), including migration and rename scenarios.


api/workloads/v1alpha2/rolebasedgroup_admission.go
```diff
@@ -33,7 +33,8 @@ import (
 // +kubebuilder:webhook:path=/validate-workloads-x-k8s-io-v1alpha2-rolebasedgroup,mutating=false,failurePolicy=fail,sideEffects=None,groups=workloads.x-k8s.io,resources=rolebasedgroups,verbs=create;update,versions=v1alpha2,name=vrolebasedgroup.kb.io,admissionReviewVersions=v1
 // +kubebuilder:object:generate=false
 type RoleBasedGroupValidator struct {
-	Client client.Client
+	Client               client.Client
+	EnableV1Alpha1Compat bool
 }
```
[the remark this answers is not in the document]
diw-zw: EnableV1Alpha1Compat is set exclusively through SetupWebhookWithManager(mgr, enableV1Alpha1Compat), which receives the value from the --enable-v1alpha1-compat flag (default true). There is no code path that constructs a RoleBasedGroupValidator without going through this function.

internal/controller/workloads/rolebasedgroup_controller.go
```diff
@@ -173,7 +179,15 @@ func (r *RoleBasedGroupReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, err
 	}
 
-	// Step 0: Pre-check validations
+	// Step 0: Reject legacy workload types when v1alpha1 compat is disabled.
+	// This is a configuration error, not a transient one: set a terminal
+	// status condition so the failure is visible via kubectl, then stop
+	// reconciling without requeue until the spec changes.
+	if stop, err := r.handleLegacyWorkloads(ctx, rbg); err != nil || stop {
```
cheyang, line 186: One legacy role stops the whole group, and the stop is terminal.

`handleLegacyWorkloads` runs before `preCheck` and before any role is reconciled, and returns `stop=true, err=nil` if *any* role in `.spec.roles` is a v1alpha1 indirect type. `Reconcile` then returns `ctrl.Result{}, nil` — no error, no `RequeueAfter` — so controller-runtime does not retry. An RBG with one Deployment role and three RoleInstanceSet roles gets none of the four reconciled, and nothing revives it but a fresh watch event.

That makes the enabled→disabled transition lossy for mixed RBGs: flip the flag and they silently stop being managed in full, healthy roles included. Note this is a regression against the previous revision of this branch, where the same whole-group failure at least came back on the error backoff.

Suggest degrading per role — mark only the legacy roles `Ready=False/LegacyWorkloadsDisabled` and keep reconciling the rest — or relying on admission rejection alone (`ValidateCreate` already does this) and letting the controller manage what it can.

api/workloads/v1alpha2/rolebasedgroup_webhook.go
```diff
@@ -21,11 +21,16 @@ import (
...
-// RoleBasedGroup with the Manager.
-func (r *RoleBasedGroup) SetupWebhookWithManager(mgr ctrl.Manager) error {
+// RoleBasedGroup with the Manager. The enableV1Alpha1Compat flag controls
+// whether the admission validator rejects v1alpha1 indirect workload types
+// (Deployment, StatefulSet, LeaderWorkerSet).
+func (r *RoleBasedGroup) SetupWebhookWithManager(mgr ctrl.Manager, enableV1Alpha1Compat bool) error {
 	return ctrl.NewWebhookManagedBy(mgr).
 		For(r).
-		WithValidator(&RoleBasedGroupValidator{Client: mgr.GetClient()}).
+		WithValidator(&RoleBasedGroupValidator{
```
cheyang, line 30: `RoleBasedGroup` gets the flag here, but `RoleBasedGroupSet` just below has no validator at all — only a conversion webhook. With compat disabled that reproduces the exact indefinite-backoff-for-a-configuration-error this PR fixes one level down:

an RBGSet whose `groupTemplate` uses a legacy workload type is admitted, its controller `Create`s each child RBG, the RBG validator rejects every one, `scaleUp` aggregates that into a returned error, and `Reconcile` returns it — so controller-runtime backs off and retries forever. No terminal condition on the RBGSet, no event, nothing in its status to tell the operator why nothing is happening.

Reproduced: 5 reconciles → 10 admission rejections, 0 child RBGs, 0 conditions. RBGSet needs the same treatment as RBG: a validating webhook that knows the flag, plus a terminal condition instead of a returned error.

internal/controller/workloads/rolebasedgroup_controller.go
```diff
@@ -274,6 +288,71 @@ func (r *RoleBasedGroupReconciler) handleRevisions(ctx context.Context, rbg *wor
...
+		Status:             metav1.ConditionFalse,
+		Reason:             LegacyWorkloadsDisabled,
+		Message:            fmt.Sprintf("v1alpha1 compat is disabled (--enable-v1alpha1-compat=false) but roles %v use legacy workload types", legacyRoles),
+		ObservedGeneration: rbg.Generation,
+		LastTransitionTime: metav1.Now(),
+	})
+	rbg.Status.ObservedGeneration = rbg.Generation
+
+	if !reflect.DeepEqual(oldStatus, rbg.Status) {
+		rbgApplyConfig := toRBGApplyConfigurationForStatus(rbg)
```
cheyang, line 341: `toRBGApplyConfigurationForStatus` always re-declares `.status.roleStatuses` from the object it is handed, so for an RBG that was healthy before the flag flip this re-writes the last-known replica counts verbatim — and because the stop is terminal, nothing ever refreshes them. The operator sees `Ready=False(LegacyWorkloadsDisabled)` sitting next to `roleStatuses` claiming `3/3 ready`, indefinitely, after the pods are gone. Worth clearing or marking the stale entries.

(SSA does *not* wipe the rest of `.status` here — I checked that specifically; conditions and roleStatuses are re-declared together.)

internal/controller/workloads/rolebasedgroup_controller.go
```diff
@@ -673,19 +769,25 @@ func (r *RoleBasedGroupReconciler) constructAndUpdateRoleStatuses(
...
-	}
 
-	stsRecon := reconciler.NewStatefulSetReconciler(r.scheme, r.client)
-	if err := stsRecon.CleanupOrphanedWorkloads(ctx, rbg); err != nil {
-		errs = append(errs, err)
-	}
+	// Only clean up v1alpha1 indirect workload types when v1alpha1 compat is enabled.
+	// When v1alpha1 compat is disabled, the controller does not manage
+	// Deployment/StatefulSet/LeaderWorkerSet resources, so cleanup is skipped.
+	if r.enableV1Alpha1Compat {
```
cheyang, line 776: Cleaning up already-created legacy workloads is a different concern from managing new ones. With this gate, an RBG-owned Deployment left behind by a role that has since migrated to RoleInstanceSet is never removed by the controller — it is owner-referenced, so GC will not collect it while the RBG lives.

Whichever way you decide, keep `delete` RBAC for the legacy types, or re-enabling compat later will 403 on cleanup.

cmd/rbgs/main.go
```diff
@@ -619,25 +629,27 @@ func startPprofServer(ctx context.Context, pprofAddr string) error {
...
-				Label: keyExistsSelector,
-			},
+	byObject := map[client.Object]cache.ByObject{
+		&corev1.Service{}: {
+			Label: keyExistsSelector,
 		},
 	}
+
+	// Only cache StatefulSet and Deployment when v1alpha1 compat is enabled.
+	if enableV1Alpha1Compat {
```
cheyang, line 646: `ByObject` is per-type *configuration*, not an allowlist. Dropping the entry doesn't prevent a cached informer for Deployment/StatefulSet — it removes the group-name label selector that *bounded* it. Any cached read with compat off would start an unfiltered cluster-wide informer, which then can't establish its watch because this PR also removes `list`/`watch` on those types from the ClusterRole.

Latent today (no reachable cached read with compat off), but the RBAC half and the cache half disagree about what "disabled" means. Keeping the selector registered in both modes costs nothing.

api/workloads/v1alpha2/rolebasedgroup_validation.go
```diff
@@ -140,6 +142,65 @@ func ValidateScalingAdapterReplicas(ctx context.Context, reader client.Reader, o
...
+// ValidateWorkloadTypesUpdate checks that no NEW v1alpha1 indirect workload
+// types are introduced on update when v1alpha1 compat is disabled. Existing
+// roles that already had legacy types in the old spec are allowed to remain
+// so users can edit the RBG to migrate away from them.
+func ValidateWorkloadTypesUpdate(oldRBG, newRBG *RoleBasedGroup) error {
+	oldLegacy := make(map[string]bool)
+	for i := range oldRBG.Spec.Roles {
+		role := &oldRBG.Spec.Roles[i]
+		if isLegacyWorkloadType(role.GetWorkloadType()) {
+			oldLegacy[role.Name] = true
```
cheyang, line 186: Grandfathering is keyed on role *name* only, so an existing legacy role can be switched to a *different* legacy type while compat is disabled — StatefulSet→Deployment, Deployment→LWS all pass. The stated intent is to let users migrate *away* from legacy types; swapping one unsupported kind for another isn't that. Comparing per-role workload types would keep the rule monotonic.

test/e2e/testcase/v1alpha2/warmup.go
```diff
@@ -211,17 +213,29 @@ func RunWarmupTestCases(f *framework.Framework) {
...
-			// Verify 2 containers (1 image-preload from prefill + 1 customized action from decode)
-			gomega.Expect(warmupPod.Spec.Containers).To(gomega.HaveLen(2))
+			// When both roles land on the same node, the warmup Pod merges actions from both roles (2 containers).
+			// When they are on different nodes, each warmup Pod has only 1 container (single role action).
+			// Either way, the total number of containers across all warmup Pods equals the number of roles (2).
+			totalContainers := 0
+			for _, wp := range warmupPodList.Items {
+				totalContainers += len(wp.Spec.Containers)
+			}
+			gomega.Expect(totalContainers).To(gomega.Equal(2), "total containers across warmup pods should equal number of roles")
```
cheyang, line 234: nit: this change looks unrelated to the flag, and unnecessary for CI. Both e2e jobs create their cluster with `kind create cluster` and no `--config` (`.github/workflows/e2e-test.yml:112`, `:297`), i.e. single-node — so both RBG pods always co-locate, `uniqueNodeCount` is always 1, and the new assertions reduce to the old ones. `62e64713` passed `e2e-test` with the old spec.

If the goal is multi-node dev clusters, note that there the merge path — this spec's stated purpose — becomes nondeterministic and can go unexercised while still passing; pinning both roles to one node would keep it covered. Either way this seems worth its own PR.

cheyang (changes requested):
The flag's design is sound and the RBAC gating is correct — I verified the enabled/disabled ClusterRoles differ exactly as documented, and the flag/chart default mismatch from the earlier revision is genuinely fixed. The chart render fix in `0151936a` also checks out: valid under all 5 value shapes I tried, and a live API server accepts both the ClusterRole and the whole release. Two things still block, both about blast radius rather than the feature itself: one legacy role terminally stops an entire RBG, and `RoleBasedGroupSet` re-creates the indefinite-backoff problem this PR set out to remove.

Two earlier concerns I want to explicitly clear so you don't have to re-argue them: the chart RBAC has **not** drifted from `config/rbac/role.yaml` (identical, 209/209 permission triples), and the Helm nil-safety concern is **not** real — I rendered `controller.features=null` and `v1alpha1Compat=null` and Helm's value coalescing restores the chart defaults, byte-identical to the default render. Worth knowing that the comment block added for that concern is what broke the chart: `62e64713` passed e2e, `66a2500a` failed at *Deploy controller*.

Unrelated suggestion: `lint`, `unit-test`, `envtest` and `build` were all green while the chart was uninstallable — only the 30-minute e2e job caught it. A `helm template … | kubectl apply --dry-run=client -f -` step would catch that class in seconds.

cheyang:
@gemini-code-assist /review

cmd/rbgs/main.go
```diff
@@ -207,6 +209,14 @@ func main() {
 	)
 	flag.BoolVar(&enablePprof, "enable-pprof", false, "Enable pprof profiling server for performance debugging.")
 	flag.StringVar(&pprofAddr, "pprof-bind-address", ":6060", "The address the pprof endpoint binds to.")
+	flag.BoolVar(
+		&enableV1Alpha1Compat, "enable-v1alpha1-compat", true,
```
cheyang, line 213: This flag changes behaviour across the whole stack (cache scoping, controller watches, webhook validation, RBAC), yet the only automated coverage is the unit test in `rolebasedgroup_validation_test.go`. The e2e change in `test/e2e/testcase/v1alpha2/warmup.go` is about warmup pod placement and is unrelated to this feature.

Could you add an e2e case under `test/e2e/testcase/v1alpha2/` that runs with `--enable-v1alpha1-compat=false` and exercises the flag through the API surface? At minimum I'd want to see: (1) the webhook rejecting an RBG that declares a Deployment/StatefulSet/LeaderWorkerSet role, and (2) the controller setting the `LegacyWorkloadsDisabled` condition for an RBG that slips past admission. The suite already has `webhook_validation.go` as a good starting point. Unit tests alone don't give us confidence the flag is wired correctly end to end. If you believe an e2e run is genuinely infeasible in this environment, let's discuss with a maintainer rather than skipping it.

api/workloads/v1alpha2/rolebasedgroup_admission.go
```diff
@@ -33,7 +33,8 @@ import (
 // +kubebuilder:webhook:path=/validate-workloads-x-k8s-io-v1alpha2-rolebasedgroup,mutating=false,failurePolicy=fail,sideEffects=None,groups=workloads.x-k8s.io,resources=rolebasedgroups,verbs=create;update,versions=v1alpha2,name=vrolebasedgroup.kb.io,admissionReviewVersions=v1
 // +kubebuilder:object:generate=false
 type RoleBasedGroupValidator struct {
-	Client client.Client
+	Client               client.Client
+	EnableV1Alpha1Compat bool
```
cheyang, line 37: `EnableV1Alpha1Compat` defaults to the bool zero value (false), while the feature is documented as defaulting to true. The production path sets it explicitly via `SetupWebhookWithManager`, so this is fine today, but any other construction of `RoleBasedGroupValidator{}` (tests, future wiring) silently gets the disabled behaviour and starts rejecting legacy workload types. It might be worth inverting the field (e.g. `DisableV1Alpha1Compat`) so the safe default falls out of the zero value, or guarding against accidental construction. Copilot flagged the same thing.

internal/controller/workloads/rolebasedgroup_controller.go
cheyang, line 776: When compat is disabled, `deleteOrphanRoles` no longer runs the Deployment/StatefulSet/LWS cleanup reconcilers. If an operator turns the flag off on a cluster that already has RBGs managing those workloads, the child objects are neither watched nor garbage-collected here, so they can linger as orphans. Is that the intended migration story? A note in the flag help text (or docs) about the expectation to clean up legacy workloads before disabling compat would help operators avoid surprise leaks.

Makefile
```diff
@@ -60,8 +60,10 @@ help: ## Display this help.
 .PHONY: manifests
 manifests: controller-gen kustomize ## Generate WebhookConfiguration, CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) crd:allowDangerousTypes=true,crdVersions=v1,generateEmbeddedObjectMeta=true,ignoreUnexportedFields=true,maxDescLen=200 rbac:roleName=controller-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases output:rbac:artifacts:config=config/rbac
-	cp config/rbac/role.yaml deploy/helm/rbgs/templates/rbac/clusterrole.yaml
-	sed -i.bak 's/name: controller-role/name: rbgs-controller-role/' deploy/helm/rbgs/templates/rbac/clusterrole.yaml && rm -f deploy/helm/rbgs/templates/rbac/clusterrole.yaml.bak
+	# NOTE: The Helm chart's clusterrole.yaml is manually maintained with
```
cheyang, line 63: Dropping the `cp config/rbac/role.yaml -> helm clusterrole.yaml` step means the Helm ClusterRole is now hand-maintained while `config/rbac/role.yaml` stays generated. The comment acknowledges this, but the two will drift the moment someone adds a kubebuilder RBAC marker and runs `make manifests`. Is there any check (a CI diff or a lint) that can catch divergence, or is the plan to rely on reviewers noticing? Worth calling out how you intend to keep them in sync going forward.

api/workloads/v1alpha2/rolebasedgroup_validation.go
```diff
@@ -140,6 +142,65 @@ func ValidateScalingAdapterReplicas(ctx context.Context, reader client.Reader, o
 	return utilerrors.NewAggregate(allErrs)
 }
 
+// isLegacyWorkloadType returns true for v1alpha1 indirect workload types
+// (Deployment, StatefulSet, LeaderWorkerSet).
+func isLegacyWorkloadType(wt string) bool {
```
cheyang, line 147: `isLegacyWorkloadType` is defined identically here and in `internal/controller/workloads/rolebasedgroup_controller.go`. Since both keys off the same `constants` set, consider exporting one shared helper to avoid the definitions drifting apart if the list of legacy types ever changes.

PR #415, opened 2026-07-29
doc: add en doc best-practice/configuring-inplace-update-and-scheduling-policies

NoobDream2568:
### Ⅰ. Motivation

This is the English docs for inplace update and scheduling policies. Earlier, the Chinese version get merged with WIP label. So this PR tries to fix it and add the English version.

NONE

## Checklist

### DIFF
--- a/doc/best-practice/en/04-configuring-inplace-update-and-scheduling-policies-guide.md
+++ b/doc/best-practice/en/04-configuring-inplace-update-and-scheduling-policies-guide.md
@@ -1,27 +1,27 @@
-# 操作文档：配置原地升级与原地调度策略
+# Operations Guide: Configuring In-Place Update and In-Place Scheduling Strategies
 
-> 对应概念文档：[4. 配置原地升级与原地调度策略](04-configuring-inplace-update-and-scheduling-policies.md)
+> Corresponding concept document: [4. Configuring In-Place Update and In-Place Scheduling Strategies](04-configuring-inplace-update-and-scheduling-policies.md)
 
-## 目标
+## Objectives
 
-验证 RBG 的原地升级（In-Place Update）和原地调度（In-Place Scheduling）策略，包括：
-1. 原地升级与 Grace Period 流量排空：仅变更镜像时 Pod 留在原节点，支持排空等待
-2. 原地调度（Preferred）：Pod 重建时优先调度回历史节点
-3. 原地调度（Required）：Pod 重建时必须调度回历史节点
+Validate RBG's In-Place Update and In-Place Scheduling strategies, including:
+1. In-Place Update with Grace Period traffic draining: Pod stays on the original node when only the image changes, with drain wait support
+2. In-Place Scheduling (Preferred): Pod is preferentially scheduled back to its historical node during recreation
+3. In-Place Scheduling (Required): Pod must be scheduled back to its historical node during recreation
 
-## 前提条件
+## Prerequisites
 
-- Kubernetes 集群版本 >= 1.24
-- 已安装 RBG Controller
-- 镜像可访问：`alpine:3.23.2`、`alpine:3.23.5`
+- Kubernetes cluster version >= 1.24
+- RBG Controller installed
+- Images accessible: `alpine:3.23.2`, `alpine:3.23.5`
 
-> **说明**：本文档使用 `sleep 3600` 作为占位命令，专注于验证 RBG 升级控制面行为，无需 GPU。如需测试真实推理功能，请替换为完整的推理引擎启动命令。
+> **Note**: This document uses `sleep 3600` as a placeholder command, focusing on validating RBG upgrade control plane behavior without requiring GPU. To test real inference functionality, replace with the full inference engine startup command.
 
 ---
 
-## 操作一：原地升级（In-Place Update）与 Grace Period 流量排空
+## Operation 1: In-Place Update with Grace Period Traffic Draining
 
-### 步骤 1：创建 2 副本 RBG，配置原地升级策略
+### Step 1: Create a 2-Replica RBG with In-Place Update Strategy
 
 ```bash
 cat <<'EOF' | kubectl apply -f -
@@ -51,15 +51,15 @@ spec:
 EOF
 ```
 
-### 预期行为
+### Expected Behavior
 
-- 创建 2 个 Pod（`inplace-update-demo-backend-0` 到 `inplace-update-demo-backend-1`）
-- 全部就绪后，RBG 状态为 Ready
+- 2 Pods created (`inplace-update-demo-backend-0` through `inplace-update-demo-backend-1`)
+- After all are ready, RBG status is Ready
 
-### 步骤 2：记录更新前的 Pod 状态
+### Step 2: Record Pod State Before Update
 
 ```bash
-# 记录 Pod 的 RESTARTS，AGE 和 IP
+# Record Pod RESTARTS, AGE, and IP
 kubectl get pods -l rbg.workloads.x-k8s.io/group-name=inplace-update-demo -o wide
 
 > NAME                            READY   STATUS    RESTARTS   AGE   IP           NODE                 NOMINATED NODE   READINESS GATES
@@ -67,32 +67,32 @@ kubectl get pods -l rbg.workloads.x-k8s.io/group-name=inplace-update-demo -o wid
 > inplace-update-demo-backend-1   1/1     Running   0          10s   10.xx.xx.12  e01-xxxxxxxxxxxxxx   <none>           2/2
 ```
 
-### 步骤 3：触发镜像更新
+### Step 3: Trigger Image Update
 
 ```bash
 kubectl patch rbg inplace-update-demo --type='json' \
   -p='[{"op": "replace", "path": "/spec/roles/0/standalonePattern/template/spec/containers/0/image", "value": "alpine:3.23.5"}]'
 ```
 
-### 预期行为（镜像更新）
+### Expected Behavior (Image Update)
 
-对于每个待更新的 Pod：
-1. Controller 将 Pod 的 `InPlaceUpdateReady` 条件设为 `False`，Pod 变为 NotReady
-2. Pod 从 Service 端点中摘除
-3. 等待 30 秒（`gracePeriodSeconds`），让已有连接处理完成
-4. Patch 容器镜像，kubelet 重启容器
-5. 容器就绪后，Pod 恢复 Ready 状态
+For each Pod being updated:
+1. Controller sets the Pod's `InPlaceUpdateReady` condition to `False`, Pod becomes NotReady
+2. Pod is removed from Service endpoints
+3. Wait 30 seconds (`gracePeriodSeconds`) for existing connections to finish processing
+4. Patch the container image, kubelet restarts the container
+5. After the container becomes ready, Pod restores Ready state
 
-同时满足以下原地升级行为：
-- 按序号从高到低逐个更新实例（1 → 0），当前实例完成 drain、容器重启并恢复 Ready 后才更新下一个
-- `type: InPlaceIfPossible`：仅变更镜像，触发原地升级
-- Pod 不离开当前节点，AGE 不会重置
-- 容器的 RESTARTS 计数增加
+The following in-place update behaviors are also satisfied:
+- Instances are updated one by one from high to low ordinal (1 → 0); the next instance starts only after the current one finishes draining, restarts, and becomes Ready again
+- `type: InPlaceIfPossible`: only image changed, triggering in-place update
+- Pod does not leave its current node, AGE does not reset
+- Container RESTARTS count increases
 
-### 验证
+### Verification
 
 ```bash
-# 观察更新过程
+# Observe the update process
 kubectl get pods -l rbg.workloads.x-k8s.io/group-name=inplace-update-demo -o wide -w
 
 > NAME                            READY   STATUS    RESTARTS      AGE   IP            NODE                 NOMINATED NODE   READINESS GATES
@@ -116,7 +116,7 @@ kubectl get pods -l rbg.workloads.x-k8s.io/group-name=inplace-update-demo -o wid
 ```
 
 ```bash
-# 确认所有 Pod 已更新到新镜像，且 Pod 未被重建（AGE 不重置）
+# Confirm all Pods have been updated to the new image and Pods were not recreated (AGE not reset)
 kubectl get pods -l rbg.workloads.x-k8s.io/group-name=inplace-update-demo -o wide
 
 > NAME                            READY   STATUS    RESTARTS   AGE   IP           NODE                 NOMINATED NODE   READINESS GATES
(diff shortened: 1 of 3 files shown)


cheyang:
@gemini-code-assist /review

doc/best-practice/en/04-configuring-inplace-update-and-scheduling-policies.md
```diff
@@ -0,0 +1,375 @@
...
+kubectl get pods -l rbg.workloads.x-k8s.io/group-name=<rbg-name> -o wide
+```
+
+If the Pod AGE is significantly greater than the time corresponding to the container RESTARTS, in-place update has been successfully executed.
+
+## Related Documents
+
+<!-- TODO: The following documents have not been created yet; links will be added once they are complete -->
+
++ Deploying Inference Services with RBG
```
cheyang: The TODO here says these docs haven't been created yet, but `01-deploy-inference-service.md` ("Deploying Inference Services with RBG") and `03-configuring-rolling-updates.md` ("Configuring Rolling Update Strategies") already live in this same `en/` directory. Consider turning those two into working relative links now and leaving the TODO only for "RBG Warmup", which genuinely doesn't exist yet. Non-blocking.

doc/best-practice/en/04-configuring-inplace-update-and-scheduling-policies.md
```diff
@@ -0,0 +1,375 @@
...
+
++ **In-Place Update**: When only the container image changes, the image is updated and the container restarted directly on the original Pod. The Pod never leaves its current node, so all node-level caches are naturally preserved.
++ **In-Place Scheduling**: When in-place update is not feasible (e.g., Pod Spec changes beyond image scope) and the Pod must be recreated, node affinity (nodeAffinity) is injected to schedule the new Pod back to its original node, reusing cached resources on that node.
+
+Used together, they minimize the impact of upgrades on inference service availability and performance metrics.
+
+## Prerequisites
+
++ Kubernetes cluster version >= 1.24
++ RBG Controller installed (see [Installation Guide](https://github.com/sgl-project/rbg))
```
cheyang, line 17: The "Installation Guide" link resolves to the repository landing page rather than an actual installation doc. If a dedicated install/getting-started page exists, pointing there would be more useful; otherwise this is fine to leave as-is. Non-blocking, and it matches the zh counterpart.

cheyang:
@gemini-code-assist /review

doc/best-practice/en/04-configuring-inplace-update-and-scheduling-policies.md
```diff
@@ -0,0 +1,375 @@
...
+```bash
+# Compare container restart count with Pod AGE
+kubectl get pods -l rbg.workloads.x-k8s.io/group-name=<rbg-name> -o wide
+```
+
+If the Pod AGE is significantly greater than the time corresponding to the container RESTARTS, in-place update has been successfully executed.
+
+## Related Documents
+
+<!-- TODO: The following documents have not been created yet; links will be added once they are complete -->
```
cheyang: The two entries below are now real links, but this comment still says the documents "have not been created yet; links will be added once they are complete." That reads as contradictory now. The zh version dropped this comment in the same commit, so the two files are also out of sync here. Suggest removing it (or narrowing it to the still-unlinked RBG Warmup entry) to match zh.

NoobDream2568: Removed.

doc/best-practice/en/04-configuring-inplace-update-and-scheduling-policies.md
```diff
@@ -0,0 +1,373 @@
...
+### Update Strategy Types
+
+RBG provides two update strategies, configured via `rollingUpdate.type`:
+
+| Strategy | Behavior | Applicable Scenario |
+| --- | --- | --- |
+| `InPlaceIfPossible` | Prefer in-place update; fall back to Pod recreation if changes exceed image scope | **Recommended**, default choice for most scenarios |
+| `RecreatePod` | Delete the old Pod and create a new one | Scenarios requiring full recreation |
+
+> **Note**: The `InPlaceOnly` strategy is not recommended. When changes exceed image scope, this strategy does not fall back to Pod recreation and the update gets stuck. Use `InPlaceIfPossible` instead.
```
cheyang: This adds a causal claim the Chinese original does not make. zh:76 only says the strategy is deprecated and not recommended, while here it says the update *gets stuck* because it does not fall back to recreation.

`InPlaceOnly` has no distinct code path. Both update sites branch on `set.Spec.UpdateStrategy.Type != RecreatePodUpdateStrategyType` (`stateful_instance_set_control.go:749`, `statelessmode/sync/update.go:182`) and follow it with `if !inplacing { deleteInstance(...) }`, so a change outside image scope falls back to recreation for `InPlaceOnly` exactly as it does for `InPlaceIfPossible`. `InPlaceOnlyUpdateStrategyType` is referenced nowhere under `internal/` or `pkg/`, only at its own constant definition.

Translating zh:76 as-is would avoid inventing the mechanism:

> **Note**: The `InPlaceOnly` strategy is deprecated and not recommended. Use `InPlaceIfPossible` instead.

There is a related question for the zh side, so this may be worth settling once for both. Nothing in v1alpha2 marks `InPlaceOnly` as deprecated: no `// Deprecated:` comment, no enum exclusion. v1alpha1 did drop the value (`api/workloads/v1alpha1/constant.go:229-241` lists only `Recreate` and `InPlaceIfPossible`), which is probably where "已废弃" comes from. Either add the marker to the API, or describe it as retained but not separately implemented.

doc/best-practice/en/04-configuring-inplace-update-and-scheduling-policies-guide.md
```diff
@@ -0,0 +1,372 @@
...
+
+For each Pod being updated:
+1. Controller sets the Pod's `InPlaceUpdateReady` condition to `False`, Pod becomes NotReady
+2. Pod is removed from Service endpoints
+3. Wait 30 seconds (`gracePeriodSeconds`) for existing connections to finish processing
+4. Patch the container image, kubelet restarts the container
+5. After the container becomes ready, Pod restores Ready state
+
+The following in-place update behaviors are also satisfied:
+- Instances are updated one by one from high to low ordinal (1 → 0), with approximately 30 seconds between the two instance updates
```
cheyang: The ordering half of this line is exactly right: instances really do go high ordinal first, strictly one at a time. The interval is the part that is off.

Running the guide's own settings (`maxUnavailable: 1`, `inPlaceUpdateStrategy.gracePeriodSeconds: 30`) on a live cluster, with 4 replicas so there are three gaps to measure rather than one:

```
t=+61s   o-backend-3  restartCount -> 1
t=+121s  o-backend-2  restartCount -> 1
t=+183s  o-backend-1  restartCount -> 1
t=+243s  o-backend-0  restartCount -> 1

gaps: 60s, 62s, 60s
```

The mechanism accounts for it. Each instance costs `gracePeriodSeconds` of draining plus the container restart and the wait to become Ready again, and only then does `maxUnavailable: 1` release the next one. So 30s is the per-pod drain, not the spacing.

Note this is a different statement from lines 136 and 370, which say each pod waits about 30s between going NotReady and the image update. That describes the drain itself and is consistent with `gracePeriodSeconds: 30`. Only this line conflates the two. The zh guide has the same sentence at line 87, so both sides need it.

doc/best-practice/en/04-configuring-inplace-update-and-scheduling-policies.md
```diff
@@ -0,0 +1,373 @@
...
+            containers:
+              - name: engine
+                image: alpine:3.23.5
+                command: ["sleep", "3600"]
+```
+
+#### Parameter Description
+| Parameter | Type | Required | Default | Description |
+| --- | --- | --- | --- | --- |
+| `rollingUpdate.type` | string | No | `InPlaceIfPossible` | Update strategy type |
```
cheyang, line 121: `InPlaceIfPossible` is a real default, applied by `pkg/reconciler/roleinstanceset_reconciler.go:232-234` when the field is empty. It is not in the schema, though: the live CRD has neither `enum` nor `default`, only the promise in its description. Checked against the real apiserver with `--dry-run=server`:

```
ACCEPTED  type: TotallyBogusValue
ACCEPTED  type: InPlaceIfPosible   (one 's' short)
with the field omitted, the returned object has: {"type": "<ABSENT>"}
```

A misspelling therefore lands in the `!= RecreatePod` branch and is treated as an in-place update, with nothing rejecting it. A footnote would warn readers; adding `+kubebuilder:validation:Enum` to the API would be the stronger fix. This is inherited from zh:121 rather than introduced here.

doc/best-practice/en/04-configuring-inplace-update-and-scheduling-policies-guide.md
```diff
@@ -0,0 +1,372 @@
...
+
+> NAME                                  READY   STATUS    RESTARTS   AGE   IP           NODE                 NOMINATED NODE   READINESS GATES
+> inplace-scheduling-required-backend-0   1/1     Running   0          30s   10.xx.xx.xx  node-A               <none>           2/2
+> inplace-scheduling-required-backend-1   1/1     Running   0          30s   10.xx.xx.xx  node-B               <none>           2/2
+> inplace-scheduling-required-backend-2   1/1     Running   0          30s   10.xx.xx.xx  node-C               <none>           2/2
+> inplace-scheduling-required-backend-3   1/1     Running   0          30s   10.xx.xx.xx  node-D               <none>           2/2
+```
+
+```bash
+# Confirm all Pods contain the new environment variable
```
cheyang, line 344: Nothing to change here. The English is correct and the note is for the other direction: zh guide:344 comments `# 确认所有 Pod 使用新镜像`, but the command inspects `env[?(@.name=="new_env")]` and step 3 never touches the image. This line renders it as the environment variable, which is right. The zh side looks like leftover copy-paste from step 1 and could be updated separately to `# 确认所有 Pod 已包含新增环境变量`, matching the wording already at zh:236.

cheyang (changes requested):
The translation quality here is high. I compared it mechanically rather than by eye: every fenced code block is byte-identical to the Chinese apart from translated comments, heading levels align 26/26 on both files, table rows 40/40 and 4/4, bullets 20/20 and 35/35, and there are no dead links. No field name, replica count, image tag, `gracePeriodSeconds` or `maxUnavailable` value drifted, and no section was dropped or invented. I also checked the negations and conditionals specifically, since those are where translations usually go wrong, and found none reversed.

I see you have already removed the stale TODO at en:371 in `e8040253`, so that check now passes and it is not in this review.

Since the sandbox cluster is finally able to run RBGs end to end, I took the whole document's behavioural content live (3-node, v1.36.1-aliyun.1, controller v0.8.0-47cfe17). Almost all of it is correct, which is worth saying explicitly:

* an image-only change really does update in place. `startTime` and the pod UID are unchanged while `restartCount` goes 0 to 1, so "AGE not reset, RESTARTS increased" is exactly right.
* the controller injects `InPlaceUpdateReady` (and `InstancePodReady`) and sets the condition, matching your step-by-step mechanism.
* instances update strictly high ordinal first: measured 3, 2, 1, 0.
* `role-inplace-scheduling: Preferred` injects `preferredDuringScheduling` with weight 100 on `kubernetes.io/hostname` toward the historical node, and `Required` injects `requiredDuringScheduling`. Both as documented.

Scripts and captured output are on `verify/pr415-en-inplace-doc` at `cheyang/rbg@27bcb6aa`, under `docs/verification/pr415-en-inplace-doc/`.

## What blocks merge

One thing. The `InPlaceOnly` note at en:76 adds a causal claim the Chinese original does not make, that the update "gets stuck" because it does not fall back to recreation. `InPlaceOnly` has no distinct code path at all (both update sites branch only on `!= RecreatePod`, and the constant is referenced nowhere under `internal/` or `pkg/`), so it behaves identically to `InPlaceIfPossible` and does fall back. Translating zh:76 as-is avoids the problem.

While there: nothing in v1alpha2 actually marks `InPlaceOnly` deprecated either, with no `// Deprecated:` comment and no enum exclusion. v1alpha1 did drop the value, which is likely where the Chinese "已废弃" came from. It might be worth settling for both languages at once, either by marking the API or by describing it as retained but not separately implemented.

## Smaller things

The guide's "approximately 30 seconds between the two instance updates" is the one measurement that came out wrong: with the guide's own settings the gaps are 60s, 62s, 60s. The 30s figure is the per-pod drain, which lines 136 and 370 describe correctly, not the spacing, which additionally includes the container restart and the wait to become Ready. Both languages have that sentence.

Also inherited from the Chinese: the documented default for `rollingUpdate.type` is applied by the controller, not the schema, so the apiserver accepts `TotallyBogusValue` and near-miss typos like `InPlaceIfPosible` and treats them as in-place updates.

And one in the other direction. zh guide:344's comment says the command confirms the new image when it actually inspects an env var. Your English rendering is the correct one and the Chinese could be back-ported.

## Not verified

The guide's literal sample `kubectl get pods` blocks (node-A through node-D) were not reproduced, since the test cluster has three nodes rather than four. The behaviour they illustrate is confirmed; the transcript is not. `gracePeriodSeconds: 0` is also untested.

cheyang:
@gemini-code-assist /review

cheyang (approved):
Re-reviewed at `a01bccfb`. The blocker is fixed, and the wording you landed on is better than what I suggested.

en:76 now reads:

> **Note**: The `InPlaceOnly` strategy is retained in the API but not separately implemented; it behaves identically to `InPlaceIfPossible` (falling back to Pod recreation when in-place update is not feasible). Use `InPlaceIfPossible` instead.

That drops the "gets stuck" causal claim the Chinese never made, and states the actual situation — retained in the API, no distinct code path — without inventing a deprecation the API does not declare. It is now more accurate than the zh original, so this is a candidate to back-port rather than something to bring into line.

The measurement is fixed too. I could not find "approximately 30 seconds" attached to the spacing between instance updates in either language any more; `30` now appears only where it is correct — the `gracePeriodSeconds` drain at guide:82 and the NotReady-to-image-update wait at guide:136, plus the zh summary row at 370. That was the one number that came out wrong against the cluster (measured 60s, 62s, 60s), so it is good to see it gone rather than adjusted to a different wrong value.

Two open threads remain, both of which I marked non-blocking at the time and neither of which needs anything here:

- **en:17** — the "Installation Guide" link resolving to the repo landing page. Matches the zh counterpart, so it is a repo-wide docs question, not this PR's.
- **en:121** — `InPlaceIfPossible` being a controller-applied default with no `enum` or `default` in the schema, so `InPlaceIfPosible` is silently accepted and treated as an in-place update. That is an API gap, not a translation defect.

Approving.

For en:121 I'd suggest a separate issue against the API rather than a footnote here: `+kubebuilder:validation:Enum` on `rollingUpdate.type` would make the apiserver reject the typo outright, which is worth more than a warning in one of the two language docs. Happy to file it if you'd rather not.

PR #417, opened 2026-08-06
e2e: warmup merge multi-role actions - roles may be scheduled onto di…

diw-zw:
### Ⅰ. Motivation

The e2e case `warmup / should complete warmup job with targetRoleBasedGroup mode and merge multi-role actions` fails on any cluster with more than one schedulable node.

The case is meant to verify **action merging**: when several roles of an RBG are scheduled onto the *same* node and each role declares a different warmup action, the controller must create exactly **one** warmup Pod on that node containing every role's action. To assert that, it hard-codes `warmupPodList.Items == 1`, `Desired == 1` and `Succeeded == 1`.

However nothing in the case makes the two roles co-locate — the pod template carries no `nodeSelector`, `nodeName` or affinity, and the expected node was simply read from `podList.Items[0].Spec.NodeName`. On a multi-node cluster the scheduler is free to spread `prefill` and `decode`, which yields one warmup Pod per node and fails the assertions:

```
[FAILED] in [It] - test/e2e/testcase/v1alpha2/warmup.go:215
Warmup debug info {"name": "warmup-targetrbg-test", "phase": "Completed", "desired": 2,
  "succeeded": 2, "conditions": "[... Reason:WarmupCompleted Message:2/2 nodes warmed up successfully]"}
Warmup pod {"name": "warmup-targetrbg-test-cn6xz", "node": "cn-hongkong.10.90.173.36"}
Warmup pod {"name": "warmup-targetrbg-test-q97hb", "node": "cn-hongkong.10.131.210.163"}
```

Note the warmup feature itself behaved correctly here (`2/2 nodes warmed up successfully`) — only the case's implicit single-node assumption was violated. The remaining warmup cases are unaffected because they all target an explicit node through `TargetNodes{NodeNames: []string{nodeName}}`.

### Ⅱ. Modifications

`test/e2e/testcase/v1alpha2/warmup.go` — make the co-location premise explicit instead of relying on the cluster having a single schedulable node:

- Pin both roles to one node by setting `podTemplate.Spec.NodeSelector = {kubernetes.io/hostname: <node>}`, reusing the existing `getFirstAvailableNodeName(f)` helper that the sibling cases already use.
- Replace "read the node from the first pod" with an explicit premise guard asserting that both RBG pods really landed on the pinned node, so a scheduling surprise reports itself clearly instead of surfacing as a confusing count mismatch further down.
- The merge assertions (1 warmup Pod on that node, 2 containers, `Desired`/`Succeeded == 1`) are kept **unchanged** — the intent of the case is preserved, only its premise is now enforced.

Mutually required inter-pod affinity was considered and rejected: two pods that each require the other to already be present on the node can never be scheduled (deadlock). A one-directional affinity (`decode` → `prefill`) would work but depends on creation order and costs an extra scheduling round-trip, so a `nodeSelector` is used for a deterministic, single-shot placement.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,6 +24,7 @@ require (
 	k8s.io/cli-runtime v0.34.1
 	k8s.io/client-go v0.34.1
 	k8s.io/code-generator v0.34.1
+	k8s.io/component-helpers v0.33.3
 	k8s.io/klog/v2 v2.130.1
 	k8s.io/kubernetes v1.34.1
 	k8s.io/utils v0.0.0-20250604170112-4c0f3b243397
@@ -118,7 +119,6 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	k8s.io/apiserver v0.34.1 // indirect
 	k8s.io/component-base v0.34.1 // indirect
-	k8s.io/component-helpers v0.33.3 // indirect
 	k8s.io/controller-manager v0.32.7 // indirect
 	k8s.io/gengo/v2 v2.0.0-20250604051438-85fd79dbfd9f // indirect
 	k8s.io/kube-openapi v0.0.0-20250710124328-f3f2b991d03b // indirect
--- a/test/e2e/testcase/v1alpha2/warmup.go
+++ b/test/e2e/testcase/v1alpha2/warmup.go
@@ -135,33 +135,33 @@ func RunWarmupTestCases(f *framework.Framework) {
 				}).Obj()
 			gomega.Expect(f.Client.Create(f.Ctx, rbg)).Should(gomega.Succeed())
 
-			// Wait for RBG pods to be scheduled
-			ginkgo.By("Waiting for RBG pods to be scheduled")
-			gomega.Eventually(func() bool {
+			// Wait until exactly the two role pods the controller will act on are scheduled.
+			// Terminating pods are filtered out to mirror getDesiredNodesToWarmup, so a pod being
+			// replaced cannot make the co-location check below fail on a count instead of on
+			// placement. The snapshot is reused so both are decided from the same observation.
+			ginkgo.By("Waiting for both role pods to be scheduled")
+			var scheduledPods []corev1.Pod
+			gomega.Eventually(func() []corev1.Pod {
+				scheduledPods = nil
 				podList := &corev1.PodList{}
-				err := f.Client.List(f.Ctx, podList,
+				if err := f.Client.List(f.Ctx, podList,
 					client.InNamespace(f.Namespace),
-					client.MatchingLabels{constants.GroupNameLabelKey: rbg.Name})
-				if err != nil || len(podList.Items) < 2 {
-					return false
+					client.MatchingLabels{constants.GroupNameLabelKey: rbg.Name}); err != nil {
+					return nil
 				}
-				for _, p := range podList.Items {
-					if p.Spec.NodeName == "" {
-						return false
+				for _, pod := range podList.Items {
+					if pod.Spec.NodeName == "" || pod.DeletionTimestamp != nil {
+						continue
 					}
+					scheduledPods = append(scheduledPods, pod)
 				}
-				return true
-			}, utils.Timeout, utils.Interval).Should(gomega.BeTrue(), "RBG pods should be scheduled")
+				return scheduledPods
+			}, utils.Timeout, utils.Interval).Should(gomega.HaveLen(2), "both RBG role pods should be scheduled")
 
 			// Guard the premise of this case: both roles must have landed on the pinned node,
 			// otherwise the merge behaviour below is not what is being exercised.
 			ginkgo.By("Verifying both roles were scheduled on the same node")
-			podList := &corev1.PodList{}
-			gomega.Expect(f.Client.List(f.Ctx, podList,
-				client.InNamespace(f.Namespace),
-				client.MatchingLabels{constants.GroupNameLabelKey: rbg.Name})).Should(gomega.Succeed())
-			gomega.Expect(podList.Items).To(gomega.HaveLen(2))
-			for _, pod := range podList.Items {
+			for _, pod := range scheduledPods {
 				gomega.Expect(pod.Spec.NodeName).To(gomega.Equal(nodeName),
 					"pod %s should be pinned to node %s", pod.Name, nodeName)
 			}
--- a/test/e2e/testcase/v1alpha2/warmup.go
+++ b/test/e2e/testcase/v1alpha2/warmup.go
@@ -101,8 +101,15 @@ func RunWarmupTestCases(f *framework.Framework) {
 		})
 
 		ginkgo.It("should complete warmup job with targetRoleBasedGroup mode and merge multi-role actions", func() {
-			ginkgo.By("Creating RBG with 2 roles")
+			ginkgo.By("Creating RBG with 2 roles pinned to the same node")
+			// This case verifies action merging: when several roles of the RBG sit on one node, the
+			// controller must create a single warmup Pod there holding every role's action. Both
+			// roles are therefore pinned to one node with a nodeSelector. Mutually required pod
+			// affinity cannot be used for this: neither pod could ever be scheduled, since each
+			// would wait for the other to already exist on the node.
+			nodeName := getFirstAvailableNodeName(f)
 			podTemplate := buildWarmupPodTemplate()
+			podTemplate.Spec.NodeSelector = map[string]string{corev1.LabelHostname: nodeName}
 			rbg := wrappersv2.BuildBasicRoleBasedGroup("warmup-rbg-test", f.Namespace).
 				WithRoles([]workloadsv1alpha2.RoleSpec{
 					{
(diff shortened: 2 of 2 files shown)


cheyang:
@gemini-code-assist /review

test/e2e/testcase/v1alpha2/warmup.go
```diff
@@ -101,8 +101,15 @@ func RunWarmupTestCases(f *framework.Framework) {
...
-			ginkgo.By("Creating RBG with 2 roles")
+			ginkgo.By("Creating RBG with 2 roles pinned to the same node")
+			// This case verifies action merging: when several roles of the RBG sit on one node, the
+			// controller must create a single warmup Pod there holding every role's action. Both
+			// roles are therefore pinned to one node with a nodeSelector. Mutually required pod
+			// affinity cannot be used for this: neither pod could ever be scheduled, since each
+			// would wait for the other to already exist on the node.
+			nodeName := getFirstAvailableNodeName(f)
 			podTemplate := buildWarmupPodTemplate()
+			podTemplate.Spec.NodeSelector = map[string]string{corev1.LabelHostname: nodeName}
```
cheyang, line 113: Pinning both roles is the right call here. The merge assertions further down only mean anything when the roles co-locate, and the old `podList.Items[0].Spec.NodeName` read just happened to hold on a single-node cluster, so this makes the premise real instead of assumed.

One side effect worth calling out: on a multi-node cluster this case used to accidentally exercise the per-node fan-out in `getDesiredNodesToWarmup` — that's exactly where the `desired: 2, succeeded: 2` in your failure log came from. After the pin, nothing under `test/e2e/` covers `targetRoleBasedGroup` with roles spread over more than one node, so the one-warmup-Pod-per-unique-node path has no e2e coverage left. CI can't pick it up either: `.github/workflows/e2e-test.yml` creates the kind cluster with no config, so it's a single node.

Not a blocker for this fix. Could you follow up with a separate case that deliberately spreads the roles (pod anti-affinity on `kubernetes.io/hostname`) and asserts `Desired == 2` with one warmup Pod per node, skipped when the cluster has fewer than two schedulable nodes? Then this case stays a deterministic merge test and the fan-out gets its own.

test/e2e/testcase/v1alpha2/warmup.go
```diff
@@ -146,13 +153,18 @@ func RunWarmupTestCases(f *framework.Framework) {
...
-			ginkgo.By("Getting node name from RBG pods")
+			// Guard the premise of this case: both roles must have landed on the pinned node,
+			// otherwise the merge behaviour below is not what is being exercised.
+			ginkgo.By("Verifying both roles were scheduled on the same node")
 			podList := &corev1.PodList{}
 			gomega.Expect(f.Client.List(f.Ctx, podList,
 				client.InNamespace(f.Namespace),
 				client.MatchingLabels{constants.GroupNameLabelKey: rbg.Name})).Should(gomega.Succeed())
-			nodeName := podList.Items[0].Spec.NodeName
+			gomega.Expect(podList.Items).To(gomega.HaveLen(2))
```
cheyang: Minor: the `Eventually` above returns true at `len(podList.Items) >= 2`, but this line asserts exactly 2. If an extra Pod is ever visible between the two List calls — a replacement, or one still terminating (the controller's own `getDesiredNodesToWarmup` explicitly skips Pods carrying a `DeletionTimestamp`) — the case fails on a count rather than on the co-location invariant you're actually guarding, which is a murkier failure than the one you're fixing.

Cheapest fix is to make the `Eventually` gate wait for exactly 2 scheduled Pods; then this line is guaranteed by construction and the loop below carries the real assertion. Fine to leave as is too — with one replica per role it holds today.

test/e2e/testcase/v1alpha2/warmup.go
diw-zw: Agreed on the coverage point, but I'd like to keep it out of this PR.

This case was written for one specific purpose: assert that when two roles land on the *same* node, the controller merges their actions into a single warmup Pod. The bug being fixed here is that the case never enforced that premise — on a multi-node cluster it stopped testing merging at all and started testing fan-out by accident. The `nodeSelector` makes the original intent explicit; it doesn't remove intended coverage.

cheyang:
@gemini-code-assist /review

test/e2e/testcase/v1alpha2/warmup.go
```diff
@@ -128,31 +135,36 @@ func RunWarmupTestCases(f *framework.Framework) {
...
-
-			// Get the node where both pods are running
-			ginkgo.By("Getting node name from RBG pods")
-			podList := &corev1.PodList{}
-			gomega.Expect(f.Client.List(f.Ctx, podList,
-				client.InNamespace(f.Namespace),
-				client.MatchingLabels{constants.GroupNameLabelKey: rbg.Name})).Should(gomega.Succeed())
-			nodeName := podList.Items[0].Spec.NodeName
+				return scheduledPods
+			}, utils.Timeout, utils.Interval).Should(gomega.HaveLen(2), "both RBG role pods should be scheduled")
```
cheyang, line 160: This is what I was after. The gate now waits for exactly two scheduled, non-terminating pods and the co-location check reads that same snapshot, so a pod being replaced mid-check can no longer turn a placement failure into a count failure. The skip condition also matches `getDesiredNodesToWarmup` exactly (`NodeName == "" || DeletionTimestamp != nil`), which is the part worth mirroring — and if a third pod is briefly visible, the gate just keeps polling instead of failing outright.

On the fan-out coverage: your reasoning holds, and keeping it out of this PR is the right call — this case is about merging, not spreading. Could you open an issue for the spread-roles case so it doesn't quietly disappear? Nothing here blocks merge from my side.

cheyang (approved):
/lgtm
/approve

test/e2e/testcase/v1alpha2/warmup.go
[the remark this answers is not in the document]
diw-zw: check node condition and validate taints, now

cheyang (commented):
/lgtm
/approve

cheyang:
@gemini-code-assist /review

test/e2e/testcase/v1alpha2/warmup.go
```diff
@@ -460,23 +473,64 @@ func RunWarmupTestCases(f *framework.Framework) {
...
+}
+
+// isNodeAvailable reports whether the node is schedulable for a pod without
+// custom tolerations: it must be Ready, not cordoned, not being deleted, and
+// carry no taint that the default pod tolerations cannot tolerate.
+func isNodeAvailable(node *corev1.Node) bool {
+	if node.Spec.Unschedulable || !node.DeletionTimestamp.IsZero() {
+		return false
+	}
+	if _, untolerated := corev1helper.FindMatchingUntoleratedTaint(node.Spec.Taints, defaultPodTolerations, nil); untolerated {
```
cheyang, line 500: This closes the gap Copilot flagged, and I agree it was real: once the case relies on the scheduler to honour the hostname selector, handing back a cordoned or tainted node turns a placement failure into a timeout with no useful signal. Ready + `Unschedulable` + `DeletionTimestamp` + taints is the right set to check, and failing with "no Ready and schedulable nodes" beats silently returning `Items[0]`.

One detail on the taint call: passing a `nil` inclusion filter means every taint is considered, including `PreferNoSchedule`, which the scheduler only treats as a soft signal. `TaintToleration` in kube-scheduler filters with `t.Effect == NoSchedule || t.Effect == NoExecute` for exactly that reason. So on a cluster where the workers carry a `PreferNoSchedule` taint, both loops fall through and the case fails on node selection even though the pods would have scheduled fine. Passing that same effect filter would make this match what the scheduler does.

Related: with the Ready gate already in place, `defaultPodTolerations` is close to a no-op — `node.kubernetes.io/not-ready` and `unreachable` only appear when Ready is False/Unknown, which you have already rejected above. The tolerations do no harm, but the filter argument is the part that actually decides anything here.

Not a blocker: kind nodes carry no `PreferNoSchedule` taint, and e2e is green at this SHA.

cheyang, line 500: Passing `nil` as the `inclusionFilter` makes `FindMatchingUntoleratedTaint` consider every taint regardless of effect, since `getFilteredTaints` returns the list unchanged when the filter is nil. `PreferNoSchedule` is a scoring preference and never keeps a pod off a node, so on a cluster whose nodes carry one, both loops in `getFirstAvailableNodeName` fall through and all five warmup cases die in `ginkgo.Fail("no Ready and schedulable nodes found in the cluster")` even though the pods would have scheduled fine. On `main` the same cluster works, because the old helper just returned `Items[0]`, so this is a regression the helper change introduces.

I raised this earlier as an aside. I've since reproduced it with a unit test on the helper, so it's more concrete than I first made it sound: a Ready, uncordoned node carrying only a `PreferNoSchedule` taint comes back unavailable. Adding the effect filter flips that test, and cordoned / NotReady / Ready=Unknown / no-Ready-condition / `NoSchedule` / `NoExecute` all stay rejected, so the filter doesn't loosen anything you meant to catch.

```go
doNotSchedule := func(t *corev1.Taint) bool {
    return t.Effect == corev1.TaintEffectNoSchedule || t.Effect == corev1.TaintEffectNoExecute
}
if _, untolerated := corev1helper.FindMatchingUntoleratedTaint(node.Spec.Taints, defaultPodTolerations, doNotSchedule); untolerated {
    return false
}
```

This is the same predicate as kube-scheduler's `DoNotScheduleTaintsFilterFunc`, whose body is these two lines verbatim. I wouldn't import it, though: it lives in `k8s.io/kubernetes/pkg/scheduler/framework/plugins/helper`, which isn't vendored here, so reusing the symbol would mean pulling scheduler plugin internals into an e2e test. Keep the local closure.

go.mod
```diff
@@ -24,6 +24,7 @@ require (
 	k8s.io/cli-runtime v0.34.1
 	k8s.io/client-go v0.34.1
 	k8s.io/code-generator v0.34.1
+	k8s.io/component-helpers v0.33.3
```
cheyang, line 27: nit: promoting this out of the indirect block is right once the code imports it, but it stays at `v0.33.3` while `k8s.io/api`, `apimachinery`, `client-go` and the rest are `v0.34.1`. It's now the only direct `k8s.io` require off that line. Nothing breaks today: `vendor/modules.txt` already carried `scheduling/corev1`, and `go build -mod=vendor ./...` plus `go vet ./test/e2e/...` are clean at this SHA. Still, a direct dependency is the one worth keeping aligned. `go get k8s.io/component-helpers@v0.34.1 && go mod vendor`.

cheyang (changes requested):
The premise checks out, and I verified it rather than taking it on faith. On a 3-node cluster the base-branch spec fails exactly as the description says: `desired: 2, succeeded: 2`, one warmup Pod on each of two nodes, failing at `warmup.go:215`. The same spec passes at `7b544482`. So the fix works, and pinning the roles is the right way to make the merge assertions mean something. Reusing one List snapshot and filtering terminating Pods also settles the race between the old gate and the check below it.

One thing worth recording: CI cannot see any of this. `e2e-test.yml` runs `kind create cluster` with no config, so it's single-node and this case passes there for the wrong reason. All 10 checks are green at this SHA, which is necessary but not sufficient evidence for this particular change.

Requesting changes on one thing, the taint filter in `isNodeAvailable`. It's narrow, since it only bites on clusters whose nodes carry a `PreferNoSchedule` taint, but it's a real regression against `main`, it's now reproduced rather than suspected, and the fix is three lines. I'd rather it go in before merge than land as a follow-up. Details inline.

Everything else is a nit. The merge assertions and the snapshot logic don't need to change.

cheyang (approved):
/lgtm
/approve

PR #418, opened 2026-08-06
fix: worker pods get no DNS identity under sharedServiceSelection=All

diw-zw:
### Ⅰ. Motivation

`leaderWorkerPattern.sharedServiceSelection: All` is documented as "the shared headless Service targets every Pod of the role", but worker Pods never actually join that Service in any usable way: they are reachable by IP through the Service's A record, yet they get **no per-Pod DNS name**, because `constructRoleInstanceTemplateByLeaderWorkerPattern` only sets `serviceName` on the **leader** component. `serviceName` is what `setComponentPodIdentity` turns into `pod.spec.hostname` / `pod.spec.subdomain`, so workers end up with neither.

Reproduced on a live cluster with a role that explicitly sets `All`:

```text
NAME                          STATUS    HOSTNAME                      SUBDOMAIN
vllm-pd-cmb-hami-decode-0-0   Running   vllm-pd-cmb-hami-decode-0-0   s-vllm-pd-cmb-hami-decode
vllm-pd-cmb-hami-decode-0-1   Running   <none>                        <none>      <-- worker
```

Secondly, the field had **no default**, while the controller's de-facto behaviour for an unset field was "Service selects all Pods, only the leader has a DNS identity" — a third state that matches neither documented value.

### Ⅱ. Modifications

**Controller** (`pkg/reconciler/`)

- `roleinstanceset_reconciler.go`: under `All`, **every** component is bound to the shared Service via `serviceName`. It is only written when non-empty, so `LeaderOnly` keeps the worker field absent rather than `""` (which would change the ControllerRevision hash).
- `svc_reconciler.go`: the selector now reads the effective policy through the helper below.

**Effective-policy helper** (`api/workloads/v1alpha2/helper.go`)

- Added `(*RoleSpec).GetSharedServiceSelection()`. For an unset field: `RoleInstanceSet` role → `LeaderOnly`; any other workload type → `All`.
- Not a CRD `+kubebuilder:default`: CRD defaulting runs *before* CEL validation, so a stored `LeaderOnly` would be rejected by the `RoleSpec` rule on every role using another workload type — including roles that never set the field. Applying it in the controller keeps the stored field `nil` and leaves the CEL rule intact.
- Scoped by workload type rather than defaulting unconditionally, so the controller and the CEL rule agree. `apps/v1/StatefulSet + leaderWorkerPattern` is rejected by no validation today and `StatefulSetReconciler` ignores the pattern, but it does call `reconcileHeadlessService`. Its Pods never receive `component-name`, so an unconditional `LeaderOnly` would narrow that selector to a label no Pod carries → **zero** endpoints.

**API** — field comment updated only. No struct field changed; the `RoleSpec` CEL rule is unchanged.

**Docs** — KEP updated: semantics, the controller-side default and its scope, the restored `### Validation` section, and a new **Upgrade Considerations** section.

**Generated** — CRDs/manifests regenerated (description text only); `make generate` is clean.

### DIFF
--- a/api/workloads/v1alpha2/helper.go
+++ b/api/workloads/v1alpha2/helper.go
@@ -274,9 +274,8 @@ func (r *RoleSpec) GetWorkerTemplatePatch() *runtime.RawExtension {
 	return r.LeaderWorkerPattern.WorkerTemplatePatch
 }
 
-// GetSharedServiceSelection returns the effective shared service selection policy. It mirrors
-// the CRD default (LeaderOnly), so patterns built in process resolve the same way as patterns
-// defaulted by the API server.
+// GetSharedServiceSelection returns the effective shared service selection policy, applying the
+// LeaderOnly default when the field is unset.
 func (lwp *LeaderWorkerPattern) GetSharedServiceSelection() SharedServiceSelectionPolicy {
 	if lwp == nil || lwp.SharedServiceSelection == nil {
 		return SharedServiceSelectionLeaderOnly
--- a/api/workloads/v1alpha2/helper.go
+++ b/api/workloads/v1alpha2/helper.go
@@ -274,13 +274,28 @@ func (r *RoleSpec) GetWorkerTemplatePatch() *runtime.RawExtension {
 	return r.LeaderWorkerPattern.WorkerTemplatePatch
 }
 
-// GetSharedServiceSelection returns the effective shared service selection policy, applying the
-// LeaderOnly default when the field is unset.
-func (lwp *LeaderWorkerPattern) GetSharedServiceSelection() SharedServiceSelectionPolicy {
-	if lwp == nil || lwp.SharedServiceSelection == nil {
-		return SharedServiceSelectionLeaderOnly
+// GetSharedServiceSelection returns the effective shared service selection policy of the role,
+// resolved in this order:
+//
+//   - a role without leaderWorkerPattern, or on any workload type other than RoleInstanceSet,
+//     resolves to All. The policy is only supported for RoleInstanceSet + leaderWorkerPattern, and
+//     the Pods of other workload types never carry the component-name label, so honouring a stored
+//     LeaderOnly would leave the shared Service matching no Pod at all. The CEL rule on RoleSpec
+//     already rejects that value at admission, but it cannot protect a cluster whose CRDs lag the
+//     controller.
+//   - otherwise the stored value, when set.
+//   - otherwise LeaderOnly.
+func (r *RoleSpec) GetSharedServiceSelection() SharedServiceSelectionPolicy {
+	if r == nil || r.LeaderWorkerPattern == nil {
+		return SharedServiceSelectionAll
 	}
-	return *lwp.SharedServiceSelection
+	if r.GetWorkloadType() != constants.RoleInstanceSetWorkloadType {
+		return SharedServiceSelectionAll
+	}
+	if r.LeaderWorkerPattern.SharedServiceSelection != nil {
+		return *r.LeaderWorkerPattern.SharedServiceSelection
+	}
+	return SharedServiceSelectionLeaderOnly
 }
 
 // GetRestartPolicyConfig returns the effective RestartPolicyConfig for this role based on its pattern.
--- a/api/workloads/v1alpha2/helper.go
+++ b/api/workloads/v1alpha2/helper.go
@@ -274,6 +274,16 @@ func (r *RoleSpec) GetWorkerTemplatePatch() *runtime.RawExtension {
 	return r.LeaderWorkerPattern.WorkerTemplatePatch
 }
 
+// GetSharedServiceSelection returns the effective shared service selection policy. It mirrors
+// the CRD default (LeaderOnly), so patterns built in process resolve the same way as patterns
+// defaulted by the API server.
+func (lwp *LeaderWorkerPattern) GetSharedServiceSelection() SharedServiceSelectionPolicy {
+	if lwp == nil || lwp.SharedServiceSelection == nil {
+		return SharedServiceSelectionLeaderOnly
+	}
+	return *lwp.SharedServiceSelection
+}
+
 // GetRestartPolicyConfig returns the effective RestartPolicyConfig for this role based on its pattern.
 func (r *RoleSpec) GetRestartPolicyConfig() RestartPolicyConfig {
 	if r == nil {
--- a/keps/260-leaderonly-service/README.md
+++ b/keps/260-leaderonly-service/README.md
@@ -116,22 +116,22 @@ This problem occurs with runtimes such as `sglang`. In a cross-node engine, work
 The problem here is different from the per-replica headless Service problem. This KEP aims to keep the role shared headless Service while controlling which Pods are targeted by it.
 
 
-- the shared Service name does not change, and the controller only needs to update Service selectors
-- Pod `ServiceName` and `subdomain` do not change and direct Pod DNS identity does not change
+- the shared Service name never changes, and the controller only needs to update the Service selector
+- the leader Pod's `ServiceName`, `subdomain`, and DNS identity never change
 
-This makes `LeaderOnly` a selector policy, not a service identity policy.
-
+The policy also decides which components of the role instance are bound to the shared Service, so
+worker Pods gain or lose their `hostname`/`subdomain` when it changes.
 
 ### Goals
 
-1. Introduce a clear API for controlling which Pods are selected by the shared headless Service of a role.
-2. Keep current behavior as the default and support a `LeaderOnly` mode for `RoleInstanceSet + leaderWorkerPattern`.
-3. Ensure that switching between `All` and `LeaderOnly` does not require Pod recreation or Service renaming.
+1. Introduce a clear API for controlling which Pods of a role participate in its shared headless Service.
+2. Default to `LeaderOnly`, which matches the common inference topology where only leader Pods serve requests, and support `All` for roles whose worker Pods must be reachable by DNS.
+3. Ensure that switching between `All` and `LeaderOnly` never renames or recreates the shared Service, and never changes the leader Pod's identity.
 
 ### Non-Goals
 
 1. Introducing per-replica headless Services or per-replica subdomain policy.
-2. Changing Pod network identity, `ServiceName`, or `subdomain`.
+2. Changing the shared Service name, or the leader Pod's `ServiceName`, `subdomain`, and FQDN.
 3. Supporting patterns that do not define a leader component.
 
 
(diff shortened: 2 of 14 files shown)


keps/260-leaderonly-service/README.md
```diff
@@ -174,8 +174,9 @@ configure the service manually instead of automatically.
 
 ```go
 type LeaderWorkerPattern struct {
-    // SharedServiceSelection indicates the service policy of the role
+    // SharedServiceSelection indicates the service policy of the role. Defaults to LeaderOnly.
     // +optional
+    // +kubebuilder:default=LeaderOnly
```
NoobDream2568: Actually we are not adding CRD defaulting here. As you mentioned in file `rolebasedgroup_types.go`,
> The default is deliberately not a CRD default: defaulting runs before validation, so a stored LeaderOnly would be rejected by the RoleSpec validation rule on every role that uses another workload type.

Therefore, the documentation should remain aligned with the current API behavior.

NoobDream2568:
If users are actually using the `All` mode, **upgrading the RBG Controller will trigger an immediate rolling upgrade of their `RoleInstance`, causing Pod recreation and service disruption.** This might be acceptable in some sense, because if users truly need the `All` mode, the previous incorrect behavior did not meet their expectations, and the inference service may not have been functioning correctly anyway. However, this needs to be explicitly communicated to them.

One possible approach to avoid Pod restart entirely might be to mutate at creation time via webhooks, rather than through the Controller. However, this would introduce additional complexity and long-term maintenance overhead.

Just raising this for discussion.

diw-zw:
> This might be acceptable in some sense, because if users truly need the `All` mode, the previous incorrect behavior did not meet their expectations, and the inference service may not have been functioning correctly anyway.

Explicitly setting it to 'All' indicates that the user genuinely needs access to the worker, so a restart is the appropriate fix. If we prevent the pod from restarting, wouldn't that actually be the wrong approach?

cheyang:
@gemini-code-assist /review

pkg/reconciler/svc_reconciler.go
```diff
@@ -107,8 +107,8 @@ func (r *ServiceReconciler) constructServiceApplyConfiguration(
 		constants.GroupNameLabelKey: rbg.Name,
 		constants.RoleNameLabelKey:  role.Name,
 	}
-	if role.IsLeaderWorkerPattern() && role.LeaderWorkerPattern.SharedServiceSelection != nil &&
-		*role.LeaderWorkerPattern.SharedServiceSelection == workloadsv1alpha2.SharedServiceSelectionLeaderOnly {
+	if role.IsLeaderWorkerPattern() &&
+		role.LeaderWorkerPattern.GetSharedServiceSelection() == workloadsv1alpha2.SharedServiceSelectionLeaderOnly {
```
cheyang: This is where the fix stops being just a fix. Before this PR, `SharedServiceSelection == nil` left the selector at group+role labels only, so the shared Service resolved to leader *and* worker pods. Routing `nil` through `GetSharedServiceSelection()` turns it into `LeaderOnly`, and the reconciler then adds `component-name=leader` to the selector of every leader-worker role that never set the field.

That field shipped in v0.7.0 (#273), so this lands on already-running workloads: as soon as the new controller starts it patches the Service in place and worker IPs drop out of the A record and the EndpointSlice. No rollout, no event, nothing a user can correlate with the upgrade. And every leaderWorkerPattern example in this repo leaves the field unset -- `examples/inference/agg-leader-worker.yaml`, `examples/inference/pd-disagg-leader-worker.yaml`, `examples/basic/rbg/patterns/leader-worker-pattern.yaml` -- so the blast radius is roughly everyone who followed the docs.

The worker-DNS half of this PR is a real bug fix and I would merge that on its own. Changing what `nil` means is a separate, breaking API decision that happens to be riding along under a `fix:` title. Either keep `nil -> All` here so the patch stays strictly additive, or split the default change into its own PR with a breaking-change note and an upgrade section. Which of the two do you prefer?

keps/260-leaderonly-service/README.md
cheyang: @NoobDream2568 was right about this, and the comment still stands at f8f2a59 -- it is not stale code. The `+kubebuilder:default=LeaderOnly` marker was dropped from `rolebasedgroup_types.go` in that same commit ("Resume leaderworkerpattern cel validation") and the regenerated CRDs no longer carry `default: LeaderOnly` either. So this snippet, plus line 197 ("the API server defaults it to `LeaderOnly`"), documents a design that is no longer in the tree.

The same commit also restored the `RoleSpec` CEL rule, yet this PR still deletes the `### Validation` section that used to document it. The rule is now undocumented in its own KEP.

Please re-sync the file with what actually landed: drop the marker from the snippet, state that the default is applied by the controller in `GetSharedServiceSelection()` and that the stored field stays `nil`, and put the Validation section back. The PR description needs the same pass -- it still says the CEL rule had to be removed because the default required it, which is the opposite of f8f2a59, and that text is what ends up in the squash commit message.

api/workloads/v1alpha2/helper.go
```diff
@@ -274,6 +274,15 @@ func (r *RoleSpec) GetWorkerTemplatePatch() *runtime.RawExtension {
 	return r.LeaderWorkerPattern.WorkerTemplatePatch
 }
 
+// GetSharedServiceSelection returns the effective shared service selection policy, applying the
+// LeaderOnly default when the field is unset.
+func (lwp *LeaderWorkerPattern) GetSharedServiceSelection() SharedServiceSelectionPolicy {
```
cheyang: The fallback is unconditional, which puts it at odds with the CEL rule that came back in f8f2a59: admission refuses an explicit `LeaderOnly` on anything that is not `RoleInstanceSet + leaderWorkerPattern`, while this helper hands `LeaderOnly` to exactly those roles whenever the field is `nil`.

For LWS roles that is harmless, since `reconcileHeadlessService` is only reached from the RoleInstanceSet and StatefulSet reconcilers. The case I am less comfortable with is `apps/v1/StatefulSet` + `leaderWorkerPattern`. Nothing rejects that combination today (the only pattern validation is standalone vs leader-worker mutual exclusion), `StatefulSetReconciler` ignores the pattern entirely, and its pods never receive `component-name` -- that label is only written by `InitComponentPodLabels` and the RIS component templates. `constructServiceApplyConfiguration` would still narrow the selector to `component-name=leader`, so the headless Service ends up matching zero pods where today it matches all of them.

Do you consider that combination reachable in practice? If yes, the fallback probably belongs behind a `GetWorkloadType() == RoleInstanceSet` check so the controller and the CEL rule agree on scope.

keps/260-leaderonly-service/README.md
```diff
@@ -253,29 +241,30 @@ Discovery artifacts and environment variables remain stable because the shared S
...
-- CRD enum validation rejects unsupported combinations via CEL rule
+- Component `serviceName` propagation: `All` binds leader and worker, `LeaderOnly` binds the leader only
+- In-place Service selector update: `All` → `LeaderOnly` and `LeaderOnly` → `All`
 
 ##### Integration tests
 
-- `All` creates one shared headless Service per role and includes leader and worker Pods
-- `LeaderOnly` creates one shared headless Service per role and includes only leader Pods
-- `All <-> LeaderOnly` updates the shared Service in place
+- An unset field is defaulted to `LeaderOnly` by the API server, including on `LeaderWorkerSet` roles, which must not be rejected
```
cheyang: This bullet says the opposite of the test it describes. The envtest at head asserts that an unset field stays `nil` (`Expect(...SharedServiceSelection).Should(BeNil())`) and that an explicit `LeaderOnly` on an LWS role is rejected with "only supported for RoleInstanceSet".

Line 220 reads the same way: "the policy has no effect on them" is true for the reconciler but not for admission, which hard-rejects the value. Worth fixing alongside the default wording above so the KEP test plan actually matches `test/envtest/testcase/rbg/shared_service_selection_test.go`.

test/envtest/testcase/rbg/shared_service_selection_test.go
```diff
@@ -0,0 +1,217 @@
...
+			By("verifying only the leader pod has a network identity")
+			pods := podsByComponent(rbg.Name, rbg.Spec.Roles[0].Name, 2)
+			leader := pods[string(constants.LeaderComponentType)]
+			Expect(leader.Spec.Hostname).Should(Equal(leader.Name))
+			Expect(leader.Spec.Subdomain).Should(Equal(rbg.GetServiceName(&rbg.Spec.Roles[0])))
+			Expect(pods[string(constants.WorkerComponentType)].Spec.Subdomain).Should(BeEmpty())
+		})
+	})
+
+	Context("When the policy is All", func() {
```
cheyang, line 173: Copilot flagged the same gap on the e2e file and I agree with it after checking the code: nothing here or in e2e covers `All -> LeaderOnly`. Both new contexts create the RBG at a fixed policy, and the e2e case only walks `LeaderOnly -> All`.

The reverse direction is the one whose mechanism I would want pinned down by a test. Dropping `WithServiceName` from the worker apply config relies on server-side apply removing a field the `rbg` field manager previously owned (`PatchObjectApplyConfiguration` uses `client.Apply` with `Force: true`), and `ServiceName` is a plain `string` with `omitempty`, so there is no explicit-empty escape hatch. Reading the code I believe it works, but that is inference rather than verification.

A third context that creates with `All`, patches to `LeaderOnly`, then asserts the worker component's `serviceName` is cleared and the replacement worker pods come back with an empty `subdomain` would close it.

keps/260-leaderonly-service/README.md
diw-zw: Good catch. I'll go ahead and update the documentation.

pkg/reconciler/svc_reconciler.go
diw-zw: Thanks for pushing on this — it deserves an explicit answer. My answer is neither option, and here's why.

## `nil -> LeaderOnly` is the fix, not a rider

The pre-PR state of an unset field wasn't `All`. It was a third state matching neither documented value: the selector included worker Pods, *and* workers had no `serviceName`, so no `hostname`/`subdomain`. Workers were endpoints of a Service under which they had no DNS identity. Collapsing that into one of the two documented values *is* the fix — splitting it would ship a release where `All` still means "in the A record, but unaddressable".

Given the choice, `LeaderOnly` is the non-destructive side:

- unset -> `LeaderOnly` touches **no Pod**. Workers never had a `hostname`/`subdomain`, so the worker `serviceName` stays absent, the component revision is unchanged, nothing rolls.
- unset -> `All` would hand every worker Pod in every existing leader-worker role a `subdomain` it didn't have. Those fields are immutable — that's a cluster-wide worker rollout on upgrade, for users who never asked for worker DNS.

So "strictly additive" is additive in the API but is the option that recreates Pods.

## Nothing in these examples ever talks to a worker

In the sglang pattern these examples use, a worker is never something a user or a router addresses. The PD example points its router at the leader Pods *by name* — `--prefill http://pd-disagg-lws-prefill-0.s-pd-disagg-lws-prefill:8000`, `--decode http://pd-disagg-lws-decode-0.s-pd-disagg-lws-decode:8000` — and the Dynamo examples route through their own frontend, with engines registering into etcd/NATS. Workers exist purely as TP ranks: they dial the leader at `--dist-init-addr $(RBG_LWP_LEADER_ADDRESS):6379` with `--node-rank $(RBG_LWP_WORKER_INDEX)`, and a rank>0 sglang node never serves a usable API. Nothing resolves a worker by name, nothing enumerates worker IPs, and none of the six examples even declares `servicePorts`.

So `LeaderOnly` is not a new constraint imposed on these examples — it is the policy they should have been declaring from day one. Exposing workers as endpoints was never a feature anyone relied on; it was the unset default papering over the fact that these roles had no way to say what they meant.

So a user who deployed one of these examples isn't broken. The only change is worker IPs leaving the A record and EndpointSlice — and per-Pod worker DNS, the thing that would let them use those IPs individually, never resolved anyway.

RBG doesn't consume worker endpoints either: `pkg/discovery/env_builder.go:54` builds `$(instance)-0.<svc>.<ns>` (the leader, unchanged under both policies), and `config_builder.go:126` is per-replica, never per-worker-Pod. The shared headless Service is also created with no ports at all (`svc_reconciler.go:118-125`) — a DNS identity mechanism, not a routing target.

cheyang (approved):
/lgtm
/approve

PR #420, opened 2026-08-07
ci: surface /release-test results in the PR merge box

diw-zw:
An issue_comment run is attached to the last commit on the default branch, so its checks are only visible on the Actions page. Mirror both e2e job results onto the PR head commit as commit statuses, with a pending status posted up front so the run shows immediately rather than 40 minutes later.

fixes #XXXX

## Checklist

### DIFF
--- a/.github/workflows/release-test.yml
+++ b/.github/workflows/release-test.yml
@@ -11,6 +11,10 @@ env:
   VOLCANO_VERSION: v1.12.0
   LWS_VERSION: v0.7.0
   SCHEDULER_PLUGINS_VERSION: v0.32.7
+  # Commit status contexts mirrored onto the PR head commit. Shared by the pending step
+  # and the reporting job so the two cannot drift apart.
+  STATUS_CONTEXT_HELM: Release Test / e2e-test-helm
+  STATUS_CONTEXT_MANIFEST: Release Test / e2e-test-manifest
 
 # Declare default permissions as read only.
 permissions: read-all
@@ -23,6 +27,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
+      statuses: write
     outputs:
       pr_head_sha: ${{ steps.get-pr.outputs.head_sha }}
       pr_ref: ${{ steps.get-pr.outputs.ref }}
@@ -50,6 +55,24 @@ jobs:
             });
             core.setOutput('head_sha', pr.data.head.sha);
             core.setOutput('ref', pr.data.head.ref);
+
+      - name: Report pending status on PR head
+        uses: actions/github-script@v7
+        env:
+          HEAD_SHA: ${{ steps.get-pr.outputs.head_sha }}
+        with:
+          script: |
+            const runUrl = `${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}`;
+            for (const name of [process.env.STATUS_CONTEXT_HELM, process.env.STATUS_CONTEXT_MANIFEST]) {
+              await github.rest.repos.createCommitStatus({
+                ...context.repo,
+                sha: process.env.HEAD_SHA,
+                state: 'pending',
+                context: name,
+                target_url: runUrl,
+                description: 'Running',
+              });
+            }
 
   # E2E test using Helm installation with default chart images (no build)
   e2e-test-helm:
@@ -326,3 +349,37 @@ jobs:
         if: always()
         run: |
           kind delete cluster --name rbgs-manifest-e2e || true
+
+  # An issue_comment run is attached to the last commit on the default branch, so its own
+  # checks never surface in the PR merge box. Mirror the results onto the PR head commit.
+  report-status:
+    needs: [check-command, e2e-test-helm, e2e-test-manifest]
+    if: always() && needs.check-command.result == 'success'
+    runs-on: ubuntu-latest
+    permissions:
+      statuses: write
+    steps:
+      - name: Report job results on PR head
+        uses: actions/github-script@v7
+        env:
+          HEAD_SHA: ${{ needs.check-command.outputs.pr_head_sha }}
+          HELM_RESULT: ${{ needs.e2e-test-helm.result }}
+          MANIFEST_RESULT: ${{ needs.e2e-test-manifest.result }}
+        with:
+          script: |
+            const runUrl = `${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}`;
+            const states = { success: 'success', cancelled: 'error', skipped: 'error' };
+            const results = [
+              [process.env.STATUS_CONTEXT_HELM, process.env.HELM_RESULT],
+              [process.env.STATUS_CONTEXT_MANIFEST, process.env.MANIFEST_RESULT],
+            ];
+            for (const [name, result] of results) {
+              await github.rest.repos.createCommitStatus({
+                ...context.repo,
+                sha: process.env.HEAD_SHA,
+                state: states[result] ?? 'failure',
+                context: name,
+                target_url: runUrl,
+                description: result,
+              });
+            }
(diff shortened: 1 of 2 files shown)


diw-zw:
> /release-test

The GitHub workflow changes won't be active until after the merge. For the actual effect, please see https://github.com/diw-zw/rbg/pull/1."

cheyang:
@gemini-code-assist /review

.github/workflows/release-test.yml
```diff
@@ -4,25 +4,39 @@ on:
   issue_comment:
     types: [created]
 
+# A later /release-test on the same PR supersedes earlier ones; without this the older run
+# can finish last and overwrite the newer run's commit statuses.
+concurrency:
+  group: release-test-${{ github.event.issue.number }}
```
cheyang: The group key is only the PR number, but this workflow fires on every `issue_comment: created`, not just on `/release-test`. The `if:` on `check-command` filters the jobs, not the run, and concurrency is evaluated before that. So any comment on the PR (a reviewer reply, the author answering a question, a bot posting a summary) opens a new run in `release-test-<number>` and, with `cancel-in-progress: true`, kills the e2e run that is already 20 minutes in.

That also breaks the thing this PR is adding. In the cancelling run `check-command` is skipped, so `report-status` never runs, and the two pending contexts you posted up front stay pending on the head commit until somebody types `/release-test` again. The merge box ends up worse than before: stuck yellow with no run behind it.

Keeping non-command comments in their own group is enough to fix it:

```yaml
concurrency:
  group: release-test-${{ github.event.issue.number }}-${{ contains(github.event.comment.body, '/release-test') }}
  cancel-in-progress: true
```

Command runs still supersede each other, which is what you wanted, but a chatty PR thread can no longer cancel one.

.github/workflows/release-test.yml
```diff
@@ -4,25 +4,39 @@ on:
...
 
 jobs:
   check-command:
+    # statuses: write below lets this job write to the PR merge box, so only trusted
+    # commenters may run the command.
     if: |
       github.event.issue.pull_request &&
-      contains(github.event.comment.body, '/release-test')
+      contains(github.event.comment.body, '/release-test') &&
+      contains(fromJSON('["OWNER", "COLLABORATOR"]'), github.event.comment.author_association)
```
cheyang: This is the right fix for the trust gap on the earlier push, and I like that `statuses: write` only lands on `check-command` and `report-status` — neither checks out PR code, so the elevated token never gets near the fork's tree while the two e2e jobs keep inheriting the top-level `read-all`.

Two notes on the list itself. `OWNER` never shows up on an org-owned repo, and `MEMBER` is missing, so an sgl-project member who isn't a direct repo collaborator gets turned away. Comments on this PR come back as `COLLABORATOR`, so the people using the command today are fine, but adding `MEMBER` costs nothing and saves a confusing dead end later.

The bigger one is that a rejected command is completely silent — no reaction, no comment, no status. Whoever typed it just sees nothing happen and reasonably concludes CI is broken. Worth reacting `-1` (or dropping one line) when the command text matched but the association didn't.

.github/workflows/release-test.yml
```diff
@@ -326,3 +358,37 @@ jobs:
...
+    needs: [check-command, e2e-test-helm, e2e-test-manifest]
+    if: always() && needs.check-command.result == 'success'
+    runs-on: ubuntu-latest
+    permissions:
+      statuses: write
+    steps:
+      - name: Report job results on PR head
+        uses: actions/github-script@v7
+        env:
+          HEAD_SHA: ${{ needs.check-command.outputs.pr_head_sha }}
```
cheyang, line 402: Pinning the status to `pr_head_sha` captured in `check-command` is the right call — it's the commit the merge box reads, and it keeps the result attached to the code that was actually tested even if the author pushes mid-run.

What I can't tell from the diff is the intended lifecycle after that. The next push produces a head commit with no `Release Test / *` status at all, and nothing re-triggers the command. That's fine while these stay informational. But if you add them to branch protection as required checks, the PR will sit on "Expected — Waiting for status" indefinitely until someone re-runs `/release-test` by hand.

Which of the two is the goal? If it's the required-check route, it'd help to say so in CONTRIBUTING — `/release-test` isn't mentioned anywhere in the repo today, so a contributor has no way to know the comment is a merge prerequisite.

cheyang:
@gemini-code-assist /review

.github/workflows/release-test.yml
```diff
@@ -4,25 +4,41 @@ on:
   issue_comment:
     types: [created]
 
+# A later /release-test on the same PR supersedes earlier ones; without this the older run
+# can finish last and overwrite the newer run's commit statuses. Concurrency is evaluated
+# before any job `if:`, so the key repeats the command gate — otherwise an unrelated comment
+# on the PR would open a run in this group and cancel an e2e run already in flight.
+concurrency:
+  group: release-test-${{ github.event.issue.number }}-${{ contains(github.event.comment.body, '/release-test') && contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.comment.author_association) }}
```
cheyang, line 12: Went back over the cancellation path with this key in place: an unrelated comment now evaluates to `...-false` and lands in a different group, so it can no longer kill an e2e run that is already going. That closes the problem from the last round.

One leftover from collapsing both conditions into a single boolean: a rejected `/release-test` also evaluates to `false`, so `reject-command` shares one group with every ordinary comment on the thread. Any comment posted right after an unauthorized attempt cancels the rejection run before the reaction and the notice go out, which puts us back to the silent drop that job exists to prevent. Keeping the two conditions as separate segments avoids it:

```yaml
group: release-test-${{ github.event.issue.number }}-${{ contains(github.event.comment.body, '/release-test') }}-${{ contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.comment.author_association) }}
```

Rejections then sit in `true-false` and unrelated chatter in `false-false`.

.github/workflows/release-test.yml
```diff
@@ -326,3 +386,37 @@ jobs:
         if: always()
         run: |
           kind delete cluster --name rbgs-manifest-e2e || true
+
+  # An issue_comment run is attached to the last commit on the default branch, so its own
+  # checks never surface in the PR merge box. Mirror the results onto the PR head commit.
+  report-status:
+    needs: [check-command, e2e-test-helm, e2e-test-manifest]
+    if: always() && needs.check-command.result == 'success'
```
cheyang, line 394: `always()` is also true when the run itself gets cancelled, so a superseded run still reaches this job: `check-command` already finished as `success`, the two e2e jobs come back `cancelled`, and the mapping below turns that into an `error` commit status on the head SHA.

That matters when `/release-test` is issued twice against the same head SHA, which is easy to do on a 40-minute suite. Run 1 gets cancelled and writes `error`, run 2 writes `pending` at roughly the same moment, and whichever lands second wins — so the merge box can sit on a red X for the whole run even though the suite is healthy. It does heal once run 2 reports, but that is the exact misleading state this PR is trying to remove.

`if: !cancelled() && needs.check-command.result == 'success'` keeps a superseded run quiet and lets the newest run own the contexts. The trade-off is that a manually cancelled run with no successor leaves the two contexts pending, which seems fine given they are informational and a new push drops them anyway.

.github/workflows/release-test.yml
```diff
@@ -51,6 +67,50 @@ jobs:
...
+
+  # A silently ignored command looks like broken CI, so tell the commenter it was dropped.
+  reject-command:
+    if: |
+      github.event.issue.pull_request &&
+      contains(github.event.comment.body, '/release-test') &&
+      !contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.comment.author_association)
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
```
cheyang, line 96: `check-command` posts its rocket reaction on the same comment with only `pull-requests: write`, and this job only ever runs on PR comments (`github.event.issue.pull_request`), so `issues: write` looks like it can go. Not worth another round on its own, just noting it while the permission blocks are fresh.

cheyang (approved):
/lgtm
/approve

PR #423, opened 2026-08-10
ci: drop the removed benchmark dashboard image from the nightly build

diw-zw:
Removing the CLI in #330 deleted the dashboard Dockerfile along with the docker-build-benchmark-dashboard and docker-push-benchmark-dashboard Makefile targets, but left the nightly workflow calling them. The step has failed with "No rule to make target" every night since, so no run has reported success.

fixes #XXXX

## Checklist

### DIFF
--- a/.github/workflows/docker-build-push.yml
+++ b/.github/workflows/docker-build-push.yml
@@ -61,15 +61,6 @@ jobs:
           echo "Target platforms: $PLATFORMS"
           make docker-buildx-push
 
-      - name: Build and push Benchmark Dashboard image
-        env:
-          TAG: ${{ steps.version.outputs.tag }}
-          IMG_REPO: ${{ secrets.DOCKERHUB_USERNAME }}
-        run: |
-          echo "Building and pushing Benchmark Dashboard image with tag: $TAG"
-          make docker-build-benchmark-dashboard
-          make docker-push-benchmark-dashboard
-
       - name: Summary
         run: |
           echo "## Docker Images Build and Push Summary" >> $GITHUB_STEP_SUMMARY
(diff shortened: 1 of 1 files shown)


cheyang:
@gemini-code-assist /review

.github/workflows/docker-build-push.yml
```diff
@@ -61,15 +61,6 @@ jobs:
...
-      - name: Build and push Benchmark Dashboard image
-        env:
-          TAG: ${{ steps.version.outputs.tag }}
-          IMG_REPO: ${{ secrets.DOCKERHUB_USERNAME }}
-        run: |
-          echo "Building and pushing Benchmark Dashboard image with tag: $TAG"
-          make docker-build-benchmark-dashboard
-          make docker-push-benchmark-dashboard
-
       - name: Summary
```
cheyang, line 64: Checked this against the tree rather than just the diff. After #330 there is no `docker-build-benchmark-dashboard` or `docker-push-benchmark-dashboard` target left in the Makefile and the dashboard Dockerfile is gone, so this step could not have succeeded. The run history lines up: the last green nightly was 2026-05-08, the day #330 merged, and every run since has failed right here. In each of those runs the multi-arch controller/CRD push above completed fine and this was the only failing step, which then skipped `Summary` and turned the whole run red. Nothing else in the tree references those targets or the `rbgs-benchmark-dashboard` image, and the Summary lines that remain match exactly what `docker-buildx-push` publishes, so the deletion is self-contained.

Not for this PR: the nightly stayed red for roughly three months before anyone picked it up. A follow-up issue for alerting on scheduled-run failures seems worth opening so the next stale `make` target does not sit unreported that long.

cheyang (approved):
/lgtm
/approve

PR #424, opened 2026-08-11
fix: restore restartPolicy as a string to keep v0.7.0 objects decodable

diw-zw:
## Summary

Changing `restartPolicy` from a string to a struct broke every cluster upgrading from v0.7.0: stored objects still hold a bare string, so the typed informer fails to decode them. One undecodable object fails the whole List, leaving the controller permanently unable to sync.

- Restore `restartPolicy` to `RestartPolicyType` (string) on `LeaderWorkerPattern`, `CustomComponentsPattern` and `RoleInstanceSpec`, marked deprecated.
- Move backoff configuration to a new optional `restartPolicyConfig` field (pointer, so "unset" stays distinguishable from an explicit empty config).
- Resolution order: `restartPolicyConfig.type` → deprecated `restartPolicy` → pattern default. The Go getters and the v1alpha1 conversion path apply the same precedence.
- Controller read policy in order of: restartPolicyConfig.type → restartPolicy → default value

## Upgrade notes

- **From v0.7.0**: no action required.
- **From v0.8.0-alpha.x**: those pre-releases defined `restartPolicy` as an *object*, which is no longer decodable. Migrate `RoleBasedGroup`, `RoleBasedGroupSet`, `RoleInstanceSet` and `RoleInstance` objects by recreating the workloads. Leaving one object in the old shape blocks listing that whole resource type.


api/workloads/v1alpha2/rolebasedgroup_defaulter.go
```diff
@@ -0,0 +1,62 @@
...
+		return fmt.Errorf("expected *RoleBasedGroup but got %T", obj)
+	}
+	klog.V(4).InfoS("defaulting RoleBasedGroup", "name", rbg.Name, "namespace", rbg.Namespace)
+
+	for i := range rbg.Spec.Roles {
+		role := &rbg.Spec.Roles[i]
+		resolved := role.GetRestartPolicyConfig()
+		switch {
+		case role.LeaderWorkerPattern != nil:
+			role.LeaderWorkerPattern.RestartPolicyConfig = &resolved
```
cheyang: This write-back makes the deprecated field inert after the first admission.

`resolveRestartPolicyConfig` gives `restartPolicyConfig.type` absolute priority, and this runs on every CREATE and UPDATE. So the value resolved *from* `restartPolicy` gets stored in the field that outranks `restartPolicy`, and every later edit to the deprecated field is ignored.

On a cluster running this branch:

```
# after creating a role with only the deprecated field set
restartPolicy       = "RecreateRoleInstanceOnPodRestart"
restartPolicyConfig = {..."type":"RecreateRoleInstanceOnPodRestart"}

# operator then sets restartPolicy: None
restartPolicy       = "None"                                          <- what was asked for
restartPolicyConfig = {..."type":"RecreateRoleInstanceOnPodRestart"}  <- what the controller uses
```

The update is accepted with no warning and the role keeps recreating its instance.

This hits any path that carries the stored object back: `kubectl edit`, a targeted `kubectl patch`, or a typed client doing read-modify-write. A full `kubectl apply` happens to escape it, because the merge patch replaces `spec.roles` wholesale and drops `restartPolicyConfig` along with it, after which the webhook resolves from the new value again. That narrows the blast radius, but it also means the field works or not depending on which command you reach for, and that is harder to debug than a consistent failure.

It lands on exactly the users this PR protects: someone upgrading from v0.7.0 whose manifests only know `restartPolicy`.

Smallest fix that keeps the documented precedence is to materialize only when the deprecated field is empty:

```go
case role.LeaderWorkerPattern != nil:
    if role.LeaderWorkerPattern.RestartPolicy == "" {
        role.LeaderWorkerPattern.RestartPolicyConfig = &resolved
    }
```

Anyone who really sets both still gets config-wins, which your e2e matrix pins, while v0.7.0-shaped objects keep the deprecated field authoritative and therefore editable. I tried this against the repro and it clears it.

I am also not convinced the write-back should stay. It only makes the effective value visible on the stored object, and it costs a new `failurePolicy: Fail` dependency on the RoleBasedGroup write path plus cluster-scoped `mutatingwebhookconfigurations` RBAC. The getters already resolve correctly without it.

doc/features/failure-handling.md
```diff
@@ -13,17 +13,40 @@ RBG supports multiple failure handling policies: `None` and `RecreateRoleInstanc
...
+3. Otherwise the pattern default (`RecreateRoleInstanceOnPodRestart`) applies.
+
+The admission webhook writes the resolved type back into `restartPolicyConfig.type` and leaves
+`restartPolicy` untouched. New workloads should only set `restartPolicyConfig`.
+
+### Upgrading
+
+- **From v0.7.0**: no action required. Objects storing `restartPolicy` as a string keep working.
+- **From v0.8.0-alpha.x**: those pre-releases briefly defined `restartPolicy` as an *object*.
+  That shape is no longer decodable. Before upgrading, migrate every affected object -
```
cheyang: "Before upgrading" is the order that loses data, and the move it describes does not carry the values.

Three things I hit walking this through on a real cluster that had four ordinary RoleInstances from the current release:

**1. The controller breaks before the CRDs are upgraded.** `helm upgrade` rolls the controller image, but the CRDs go through the separate `crd-upgrade` Job. So the new binary meets the old objects first:

```
"Failed to watch","type":"*v1alpha2.RoleInstance",
"error":"failed to list *v1alpha2.RoleInstance: json: cannot unmarshal object into
         Go struct field RoleInstanceSpec.items.spec.restartPolicy of type v1alpha2.RestartPolicyType"
```

The pods stayed `1/1 Running` and Ready the whole time, so nothing surfaces this. The RoleInstance cache never syncs, which matches what the PR description says about the v0.7.0 direction.

**2. Migrating at this point deletes the policy.** `restartPolicyConfig` is not in the schema yet, so it is pruned, and the patch still reports success:

```
$ kubectl patch roleinstance nginx-cluster-backend-0 --type=merge \
    -p '{"spec":{"restartPolicy":null,"restartPolicyConfig":{"type":"None"}}}'
Warning: unknown field "spec.restartPolicyConfig"
roleinstance.workloads.x-k8s.io/nginx-cluster-backend-0 patched

restartPolicy:       null
restartPolicyConfig: null
```

**3. After the CRDs are upgraded, the stored values are no longer readable.** The new schema types `restartPolicy` as a string, so the API server prunes the object's properties on the way out:

```
nginx-cluster-backend-0 restartPolicy= {}
nginx-cluster-backend-1 restartPolicy= {}
```

`{}` still fails the typed decode, so the object is still broken, but the type and backoff values are gone from anything the API will hand back. That makes the obvious one-liner a trap:

```
kubectl patch ... --type=json -p '[{"op":"move","from":"/spec/restartPolicy","path":"/spec/restartPolicyConfig"}]'
```

It unbreaks the LIST, which is convincing, and moves nothing. The object comes back with schema defaults and no `type`, so a role configured for `RecreateRoleInstanceOnPodRestart` quietly stops recreating.

What worked was the opposite order, re-supplying the values explicitly:

```bash
kubectl patch roleinstance <name> -n <ns> --type=merge \
  -p '{"spec":{"restartPolicy":null,"restartPolicyConfig":{"type":"<type>"}}}'
```

After that all four objects decoded, the controller restarted clean, and the workload stayed Ready. Note `<type>` has to come from the RoleInstanceSet template or the parent RBG, since the object being repaired can no longer tell you.

Could this section say CRDs first, then migrate, and carry that command plus a warning about the `move` form? "or delete and recreate the workloads" is also worth softening, since that means an inference outage, and it turns out not to be necessary.

One related note: I expected the API server to reject writes to un-migrated objects, which would have ruled out in-place repair. It does not, because CRD validation ratcheting skips re-validating an unchanged invalid field. So the in-place path is available.

pkg/webhook/certmanager.go
```diff
@@ -225,35 +225,91 @@ func (m *CertManager) retry(ctx context.Context, kind, name string, op func() er
...
 		return nil
 	}
 
-	upToDate := true
-	for i := range vwc.Webhooks {
-		if !reflect.DeepEqual(vwc.Webhooks[i].ClientConfig.CABundle, caCert) {
-			upToDate = false
+	needsPatch := false
+	for i := 0; i < webhooksField.Len(); i++ {
+		bundle := webhooksField.Index(i).FieldByName("ClientConfig").FieldByName("CABundle").Bytes()
```
cheyang: nit: `dupl` forced this dedup, so something had to give, but `reflect` seems like a steep price for it. `FieldByName` moves field checking to runtime, and only `Webhooks` is `IsValid()`-guarded, so a third caller with a different shape panics here instead of failing to compile. That matters a little more than usual because this runs in `bootstrapWebhookCerts` and the new webhook is `failurePolicy: Fail`.

Generics, or a small accessor returning `[]*[]byte` per concrete type, would satisfy `dupl` while keeping the compiler involved.

Not blocking, and I did check it works: I exercised the paths your tests skip, including an absent `caBundle` (what `helm install` ships), a config where only one of three entries is stale, and a re-run against an already-correct bundle. All three pass without panicking, and the already-correct case leaves the object untouched.

cheyang (dismissed):
Restoring `restartPolicy` to a string is the right call, and the v0.7.0 decode path is now covered. Thanks for the lint pass and the conversion tests. Two things still block for me, both about the transition rather than the core change.

The defaulting webhook writes the resolved type into the field that outranks the deprecated one, so after the first admission the deprecated field stops having any effect on read-modify-write updates. And the upgrade note in `doc/features/failure-handling.md` asks operators to migrate before upgrading, which is the one order in which the migration silently deletes the policy.

I reproduced both against a real cluster running the current release plus a build of this branch, so the inline comments quote observed output.

doc/features/failure-handling.md
```diff
@@ -13,17 +13,54 @@ RBG supports multiple failure handling policies: `None` and `RecreateRoleInstanc
...
+
+  1. Back up the affected objects first (`kubectl get rolebasedgroups,rolebasedgroupsets,roleinstancesets,roleinstances -A -o json > backup.json`) — once the CRDs are upgraded, the stored values are unrecoverable. Then upgrade the CRDs to this release (apply the CRD manifests, or run the chart's CRD upgrader Job on its own). If you run a full `helm upgrade` instead, the controller image is rolled before the CRD upgrader finishes, and the new controller fails to list objects still stored in the old shape.
+  2. Migrate every affected object — `RoleInstance`, `RoleInstanceSet`, `RoleBasedGroup` and `RoleBasedGroupSet` — by clearing the object-shaped `restartPolicy` and re-supplying the values under `restartPolicyConfig`:
```
cheyang: This step lists four kinds, but the examples below cover only `RoleInstance` and `RoleInstanceSet`. For the other two the same patch shape is destructive.

A RoleBasedGroup's policy sits at `spec.roles[i].leaderWorkerPattern.restartPolicy`. `spec.roles` is a plain list, and for a custom resource `--type=merge` is an RFC-7386 merge patch, so naming `roles` replaces the whole array. Adapting the RoleInstance example is the obvious thing to try:

```bash
kubectl patch rolebasedgroup <name> -n <ns> --type=merge \
  -p '{"spec":{"roles":[{"name":"role-1","leaderWorkerPattern":{"restartPolicy":null,"restartPolicyConfig":{"type":"None"}}}]}}'
```

A real API server accepts that and leaves the role without its pod template:

```
after the merge patch: template=false  restartPolicyConfig=&{None ...}
```

There is no warning and no validation error to go on. The role is left unbuildable, and this is a migration step people run under time pressure on a cluster whose controller is already not reconciling.

An indexed JSON patch is safe, and I confirmed it keeps the template and `size` intact:

```bash
kubectl patch rolebasedgroup <name> -n <ns> --type=json -p '[
  {"op":"remove","path":"/spec/roles/0/leaderWorkerPattern/restartPolicy"},
  {"op":"add","path":"/spec/roles/0/leaderWorkerPattern/restartPolicyConfig",
   "value":{"type":"<type>"}}]'
```

One index per affected role, and `customComponentsPattern` instead of `leaderWorkerPattern` where that is the pattern in use. Re-applying the original manifest with `restartPolicyConfig` substituted in also works and may be easier to explain, since most people have the manifest.

Could step 2 carry one of those for RoleBasedGroup and RoleBasedGroupSet, and say explicitly that the merge-patch form is only safe for the two kinds where the field is not inside a list?

doc/features/failure-handling.md
```diff
@@ -13,17 +13,54 @@ RBG supports multiple failure handling policies: `None` and `RecreateRoleInstanc
...
+     kubectl patch roleinstance <name> -n <namespace> --type=merge \
+       -p '{"spec":{"restartPolicy":null,"restartPolicyConfig":{"type":"<type>"}}}'
+
+     # RoleInstanceSet: roleInstanceTemplate inlines RoleInstanceSpec, so the
+     # fields sit directly under roleInstanceTemplate (no nested "spec" key).
+     kubectl patch roleinstanceset <name> -n <namespace> --type=merge \
+       -p '{"spec":{"roleInstanceTemplate":{"restartPolicy":null,"restartPolicyConfig":{"type":"<type>"}}}}'
+     ```
+
+     `<type>` must be re-supplied explicitly: once the new CRDs are installed, the API server prunes the old object shape on read for every affected resource — including the parent `RoleBasedGroup` — so no object can tell you its previous value. Take it from your manifests, or back the objects up first (`kubectl get <kind> -o json > backup.json`).
```
cheyang: "every affected resource" and "no object can tell you its previous value" are a bit stronger than what the API server does.

`RoleInstanceSet.spec.roleInstanceTemplate` is declared `x-kubernetes-preserve-unknown-fields` with no properties, so nothing under it gets pruned. After upgrading the CRDs the old values are still readable there:

```
RoleInstanceSet spec.roleInstanceTemplate.restartPolicy after CRD upgrade:
  found=true value={"baseDelaySeconds":30,"type":"RecreateRoleInstanceOnPodRestart"}
```

The typed decode still fails, so the object still has to be migrated. Only the "you cannot recover the value" part does not hold for this kind. RoleInstance and RoleBasedGroup are pruned as described.

Erring on the safe side is reasonable, so this is only worth a tweak: something like "pruned for RoleInstance and RoleBasedGroup; RoleInstanceSet keeps them because its template is schemaless" would save people a trip to their backup. The backup advice in step 1 is good either way.

cheyang (dismissed):
Re-checked at `16969e7f`. Removing the defaulter and the MutatingWebhookConfiguration outright is a better answer than the conditional I suggested, and it clears the finding. On a real API server, `kubectl edit` and a targeted JSON patch of the deprecated `restartPolicy` both take effect now, and nothing materializes `restartPolicyConfig` behind the user's back. The precedence survived too: an explicit `restartPolicyConfig.type` still wins. The `dupl` workaround in `certmanager.go` went away with it, so the `reflect` question is moot too.

The rewritten upgrade procedure covers everything I asked about and a few things I had missed. I ran the two commands it gives against a real API server and both work, including the point about `roleInstanceTemplate` inlining `RoleInstanceSpec`.

One thing left, on the same section: step 2 asks operators to migrate RoleBasedGroups but the only worked example is the RoleInstance one, and adapting that example to a RoleBasedGroup deletes the role's pod template without complaining. Details inline, along with a smaller wording point.

cheyang:
@gemini-code-assist /review

doc/features/failure-handling.md
diw-zw: Disables v0.8.0-alpha.3 and v0.8.0-alpha.4. While the underlying breaking changes have been resolved, upgrading to or from these versions is still problematic. Since all other upgrades are seamless, we simply block these two versions instead of building a migration workaround.

doc/features/failure-handling.md
```diff
@@ -13,17 +13,34 @@ RBG supports multiple failure handling policies: `None` and `RecreateRoleInstanc
...
+1. If `restartPolicyConfig.type` is set, it is used.
+2. Otherwise, if `restartPolicy` is set, its value is used.
+3. Otherwise the pattern default (`RecreateRoleInstanceOnPodRestart`) applies.
+
+The two fields are resolved at runtime — the controller and the conversion webhook always read them through the precedence above, and stored objects are not modified. New workloads should only set `restartPolicyConfig`.
+
+### Upgrading
+
+- **From v0.7.0**: no action required. The `restartPolicy` wire format is unchanged.
+- **From v0.8.0-alpha.3 / v0.8.0-alpha.4**: these pre-releases briefly defined `restartPolicy` as an *object* instead of a string. That shape is no longer decodable by v0.8.0+ controllers. These two versions are unsupported and should not be used. If you deployed them, delete all affected objects (`RoleBasedGroup`, `RoleBasedGroupSet`, `RoleInstance`, `RoleInstanceSet`) and recreate them with v0.8.0+.
```
cheyang: Two things on this bullet.

`deploy/helm/rbgs/Chart.yaml` on this branch still carries `version: 0.8.0-alpha.4` — that string only moves in a separate release chore, so anyone who installed from main after the alpha.4 tag sees `0.8.0-alpha.4` in `helm ls` while running code that stores the string field. Reading this line, they would conclude they have to delete every workload they own. Pinning it to the released artifacts ("the released v0.8.0-alpha.3 and v0.8.0-alpha.4 images") keeps the destructive instruction aimed at the people who actually need it.

The other half is what the rewrite dropped. The previous revision of this bullet said that leaving even one object in the old shape stops the new controller from listing that resource type at all. That is the symptom an operator actually observes — a controller that comes up and never syncs — and it is also why cleaning up only some of the objects does not help. It survived in the PR body but not here, and this page is where someone debugging a wedged controller will land.

cheyang (commented):
The destructive migration example is gone, so my earlier blocker is resolved. Declaring the two alpha tags unsupported and telling people to recreate the workloads is a fine call for pre-releases — no need to carry a migration path for a shape that only existed for two builds.

I checked the version claim against the tags: v0.8.0-alpha.1 and alpha.2 still stored `restartPolicy` as a string, alpha.3 and alpha.4 stored it as an object, and this branch restores the string plus `restartPolicyConfig`. Naming exactly those two is correct. Code is unchanged since my last pass and CI is green including both e2e jobs.

One thing left before this merges: the PR description no longer matches the branch, and it becomes the squash commit message. Three claims are stale — the `RoleBasedGroupDefaulter` defaulting webhook, the mutating-webhook caBundle wiring with the new `mutatingwebhookconfigurations` RBAC, and the in-place v0.8.0-alpha.x migration steps. None of them are in the tree anymore, and the two Copilot comments on this SHA are chasing exactly those sentences rather than a real gap in the code. Trimming the summary to what actually shipped would stop that from confusing whoever reads the release notes.

NoobDream2568:
**Upgrade impact for v0.7.0 users: a role-level `restartPolicy: None` silently flips to the default**

In v0.7.0 `restartPolicy` lived on the role (`spec.roles[].restartPolicy`). The new controller no longer reads that field — it only resolves `restartPolicyConfig.type` → pattern-level `restartPolicy` → default. So on upgrade:

- `restartPolicy: None` → **behavior changes**: the role falls back to the default `RecreateRoleInstanceOnPodRestart`, so instances that used to survive a pod restart now get recreated.
- `RecreateRoleInstanceOnPodRestart` → no behavior change (which matches the new default).

Verified on kind: the field survives the CRD upgrade but is pruned as an unknown field on the first `spec` write, so the loss is silent. Since the description says "From v0.7.0: no action required", it's worth one line in `doc/features/failure-handling.md`: the role-level `restartPolicy` is no longer read, and roles that had `None` need `restartPolicyConfig.type: None` set explicitly.

api/workloads/v1alpha2/rolebasedgroup_webhook.go
```diff
@@ -20,8 +20,8 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
-// SetupWebhookWithManager sets up the conversion and validating webhooks for
-// RoleBasedGroup with the Manager.
+// SetupWebhookWithManager sets up the conversion and validating webhooks
+// for RoleBasedGroup with the Manager.
```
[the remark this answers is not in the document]
diw-zw: The defaulting webhook was intentionally removed in commit 04ecd11c. Resolution is done entirely at runtime via GetRestartPolicyConfig() (helper.go:304-316), which calls resolveRestartPolicyConfig with pattern-specific defaults. No write-back, no mutating webhook, no certmanager changes are needed. The PR description has been updated to remove references to the defaulting webhook.

pkg/webhook/certmanager.go
```diff
@@ -226,32 +226,33 @@ func (m *CertManager) retry(ctx context.Context, kind, name string, op func() er
 }
 
 func (m *CertManager) patchOneValidatingWebhook(ctx context.Context, name string, caCert []byte) error {
-	vwc := &admissionregistrationv1.ValidatingWebhookConfiguration{}
-	if err := m.client.Get(ctx, client.ObjectKey{Name: name}, vwc); err != nil {
+	config := &admissionregistrationv1.ValidatingWebhookConfiguration{}
+	if err := m.client.Get(ctx, client.ObjectKey{Name: name}, config); err != nil {
 		return fmt.Errorf("getting ValidatingWebhookConfiguration %s: %w", name, err)
```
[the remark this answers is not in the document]
diw-zw: The defaulting webhook was intentionally removed in commit 04ecd11c. Resolution is done entirely at runtime via GetRestartPolicyConfig() (helper.go:304-316), which calls resolveRestartPolicyConfig with pattern-specific defaults. No write-back, no mutating webhook, no certmanager
    changes are needed. The PR description has been updated to remove references to the defaulting webhook.

api/workloads/v1alpha1/rolebasedgroup_conversion_test.go
```diff
@@ -1040,3 +1040,169 @@ func TestRoleBasedGroup_RoundTrip_RoleTemplates(t *testing.T) {
 	assert.Equal(t, "gpu-base", role.TemplateSource.TemplateRef.Name)
 	assert.JSONEq(t, string(patch.Raw), string(role.TemplatePatch.Raw))
 }
+
+// ── RestartPolicy conversion ─────────────────────────────────────────────────
+
+func TestRoleBasedGroup_ConvertTo_RestartPolicyEmptyMapsToNone(t *testing.T) {
+	// v1alpha1 empty RestartPolicy must convert to v1alpha2 RestartPolicyNone
```
cheyang: This comment codifies a behaviour that contradicts the v1alpha1 field it converts from. Agreeing with Copilot's note here, and I checked it independently rather than taking its word for it.

`api/workloads/v1alpha1/rolebasedgroup_types.go` says the empty value is pattern-dependent:

```go
// RestartPolicy defines the restart policy when pod failures happen.
// The default value is RecreateRoleInstanceOnPodRestart for LWS and None for STS & Deploy.
// Therefore, no default value is set.
```

So an unset policy on a role with `leaderWorkerSet` means Recreate. `convertRestartPolicyV1alpha1ToV2` maps every empty value to `None`, and because conversion writes it into `restartPolicyConfig.Type` explicitly, the v1alpha2 LeaderWorkerPattern default never gets a chance to apply. Converting a v1alpha1 LWS role that never set the field:

```
converted restartPolicyConfig = &{Type:None BaseDelaySeconds:<nil> MaxDelaySeconds:<nil>}
expected "RecreateRoleInstanceOnPodRestart", got "None"
```

The same conversion for a StatefulSet/Deployment-shaped role correctly yields `None`, which the v1alpha1 doc also specifies, so a fix has to stay pattern-aware rather than flipping the mapping wholesale. Leaving `restartPolicyConfig` nil for the LWS-empty case and letting `GetRestartPolicy()` supply the pattern default would do it.

The mapping is the same on the merge base, so this is not a regression from this PR and I would not hold the PR for it. It is worth its own change. The ask here is narrower: please don't land a test that states the inverted behaviour is intended. Either drop this case for now, or keep it and reword it to say it records current (incorrect) behaviour pending the conversion fix. As written, whoever fixes the conversion later will find a green test telling them the old result was correct.

doc/features/failure-handling.md
```diff
@@ -13,17 +13,34 @@ RBG supports multiple failure handling policies: `None` and `RecreateRoleInstanc
...
+
+1. If `restartPolicyConfig.type` is set, it is used.
+2. Otherwise, if `restartPolicy` is set, its value is used.
+3. Otherwise the pattern default (`RecreateRoleInstanceOnPodRestart`) applies.
+
+The two fields are resolved at runtime — the controller and the conversion webhook always read them through the precedence above, and stored objects are not modified. New workloads should only set `restartPolicyConfig`.
+
+### Upgrading
+
+- **From v0.7.0**: no action required. The `restartPolicy` wire format is unchanged.
```
cheyang, line 41: "The `restartPolicy` wire format is unchanged" is true for what a user writes on a RoleBasedGroup, but not for the `roleInstanceTemplate` the controller writes, and that one is what ControllerRevision hashes.

v0.7.0's reconciler stored the policy as a bare string:

```go
RoleInstanceTemplate().WithRestartPolicy(role.GetRestartPolicy())
```

and this release stores an object:

```go
RoleInstanceTemplate().WithRestartPolicyConfig(...)
```

Same effective policy, different revision:

```
v0.7.0 shape  -> ris-76c7b4b4d5  {"restartPolicy":"None"}
this release  -> ris-766d7fb8f   {"restartPolicyConfig":{"baseDelaySeconds":30,"maxDelaySeconds":600,"type":"None"}}
```

I checked what that costs on a running controller before raising it, because "the revision changes" on its own would be alarming. It is fine: a new revision gets created and `updateRevision` points at it, while the RoleInstances keep their UIDs. It goes through the in-place path since the pod template is untouched, so no pods are recreated.

What operators do see is a revision appearing right after the upgrade and one `controller-revision-hash` relabel across their pods, with `updateRevision != currentRevision` until that settles. On a set already near `revisionHistoryLimit` it also costs one slot. Someone watching a rollout dashboard during the upgrade will notice.

So the line just needs to be narrower. Something like "no action required; the `restartPolicy` field you set is still accepted" plus a note that the upgrade produces one revision change per RoleInstanceSet without restarting pods. Saying the wire format is unchanged invites the conclusion that nothing moves at all.

For reference on the general shape of this: upstream added the `StatefulSetSemanticRevisionComparison` feature gate for exactly this class of problem, described as mitigating "spurious statefulset rollouts due to controller revision comparison mismatches which are not semantically significant (e.g. serialization differences or missing defaulted fields)". Deployment sidesteps it differently, comparing templates with `apiequality.Semantic.DeepEqual` rather than by hash. Not something to act on here given pods don't restart, but it is the reason I went looking.

cheyang (commented):
Re-checked at `6fc49a96`. Both things I was blocking on are resolved: removing the defaulter and the MutatingWebhookConfiguration fixed the inert-deprecated-field problem, and dropping the migration procedure removes the RoleBasedGroup gap I raised last round. Nothing in this review blocks merge.

I also chased down what the template rename does to ControllerRevision, since `roleInstanceTemplate` is what gets hashed and this release serializes the policy differently than v0.7.0 did. Running the real RoleInstanceSet controller against a v0.7.0-shaped template and then switching it to the new shape: a new revision is created and `updateRevision` moves to it, but the RoleInstances keep their UIDs and are not recreated. It resolves through the in-place path because the pod template itself is untouched. So the churn is real but harmless, and the only thing I would change is one sentence in the upgrade note.

One pre-existing bug surfaced along the way, in the conversion default. It is not from this PR, so it belongs in its own change, but this PR adds a test that pins the wrong behaviour as expected. Details inline.

cheyang:
Thanks for digging into the upgrade path. I went to reproduce this because a silent policy flip would be serious, and the conclusion does not hold up. Two things are off, though the instinct behind it points at something real.

**The role-level field only exists in v1alpha1.** Checking the v0.7.0 CRD schema directly:

| v0.7.0 CRD version | `spec.roles[].restartPolicy` |
|---|---|
| v1alpha1 | present |
| v1alpha2 | absent (the policy sits on `leaderWorkerPattern` / `customComponentsPattern`) |

v0.7.0's v1alpha2 `RoleSpec` has no `RestartPolicy` field at all; `Pattern` is embedded inline and the policy lives inside the pattern. So a v1alpha2 user could never set it at the role level, and there is nothing there to lose on upgrade.

**For v1alpha1 objects the field is read, and `None` survives.** The conversion webhook maps the role-level value into `restartPolicyConfig.Type` via `convertRestartPolicyV1alpha1ToV2`. Converting a v1alpha1 LWS role on this branch:

```
role-level None     -> restartPolicyConfig={Type:None}                      effective "None"
role-level Recreate -> restartPolicyConfig={Type:RecreateRoleInstanceOnPodRestart}  effective "RecreateRoleInstanceOnPodRestart"
```

`None` does not fall back to the pattern default.

On the kind observation: if the field was set on a **v1alpha2** object, it was already an unknown field in v0.7.0 and would have been pruned there too. The pruning is real, it just is not something this PR introduced.

**The real problem runs the other way, and hits a different case.** `convertRestartPolicyV1alpha1ToV2` maps every *empty* value to `None`, but the v1alpha1 field documents an empty value as pattern-dependent:

```go
// The default value is RecreateRoleInstanceOnPodRestart for LWS and None for STS & Deploy.
// Therefore, no default value is set.
```

So a v1alpha1 LWS role that never set the policy converts to `None` when it should resolve to `Recreate`. That is Recreate turning into None, the opposite direction, and it hits roles that left the field alone rather than roles that set it. I raised it at https://github.com/sgl-project/rbg/pull/424#discussion_r3772798798 with a repro; it predates this PR, so it wants its own change.

Worth noting the upgrade note does overstate things slightly, just not in the way described here: the `restartPolicy` a user writes is still accepted, but the `roleInstanceTemplate` the controller writes changed shape, which produces one ControllerRevision change per RoleInstanceSet on upgrade. I checked what that costs and pods are not recreated, so it is cosmetic.

cheyang (approved):
Approving at `1c67becd`.

The problem this PR fixes is real and I reproduced it before reviewing the fix: a stored object holding `restartPolicy` as a bare string makes the typed LIST fail for the entire resource type, and a controller-runtime cache over it never syncs, so the controller is stranded rather than skipping the one object. Restoring the string is the right call.

On the two things I was blocking on, plus one that came out of them:

`04ecd11c` removed `RoleBasedGroupDefaulter` and the MutatingWebhookConfiguration outright instead of patching around the write-back. That was the better of the two options I offered, and it also drops the `failurePolicy: Fail` dependency on the RoleBasedGroup write path and the cluster-scoped RBAC that came with it. On a real API server, `kubectl edit` and a targeted JSON patch of the deprecated field now both take effect, and an explicit `restartPolicyConfig.type` still wins.

`6fc49a96` replaced the migration procedure with "these two pre-releases are unsupported". I had pushed for a documented procedure, and walking it on a live cluster showed why dropping it is more honest: the historical ControllerRevision data carries the old shape too, clearing it strands the RoleInstance revision labels, and repairing that triggers a full recreate. Declaring alpha.3 and alpha.4 unsupported is a cleaner boundary than a procedure with that many edges.

`7878f352`/`1c67becd` fixed the v1alpha1 conversion default, which predated this PR. The fix is better than what I suggested: an empty value now stays empty so the pattern-aware default applies, and `GetRawRestartPolicyType()` keeps the reverse path faithful so a read-back then update does not change the stored representation. Splitting the round-trip test into `expectedRaw` and `expectedResolved` is the right shape.

One more thing worried me enough to measure it, and it came out fine. The `roleInstanceTemplate` the controller writes did change shape, so upgrading produces a new ControllerRevision per RoleInstanceSet. Running the real RoleInstanceSet controller against a v0.7.0-shaped template and switching it to the new one, `updateRevision` moves but the RoleInstances keep their UIDs. It resolves in place because the pod template is untouched, so no pods restart. 

That leaves the `wire format is unchanged` wording in the upgrade note, which claims a bit more than holds. Not worth holding the PR for; fold it into whatever touches that file next.

PR #427, opened 2026-08-17
test(envtest): pin RoleInstanceSet revision stability across restart policy serialization

cheyang:
A RoleInstanceSet's ControllerRevision is derived from the entire `roleInstanceTemplate`, so changing how any field is serialized moves the revision hash even when the effective configuration is identical. #424 did exactly that: the restart policy moved from the `restartPolicy` string to `restartPolicyConfig`, which produces one new revision per RoleInstanceSet on upgrade.

A revision change on its own is fine. Recreating the RoleInstances is not, since the pod template did not change and the workload has no reason to restart. Nothing currently pins that boundary, so a future change to the template serialization could start rolling every workload on upgrade without a test noticing.

This adds two specs to the existing envtest suite:

- the main one stores the policy through the deprecated string field, waits for the RoleInstance, then re-serializes the same policy under `restartPolicyConfig`. It asserts the RoleInstance UIDs are unchanged, and separately that `updateRevision` did move, so the check cannot pass by the set simply never being re-revisioned.
- a small one covering that both field shapes resolve to the same effective policy, and that an explicit `restartPolicyConfig.type` outranks the deprecated field.

Test-only, no production changes. Verified against `main` at 4f94d944.

### DIFF
--- a/test/envtest/testcase/restart_policy/restart_policy_revision_test.go
+++ b/test/envtest/testcase/restart_policy/restart_policy_revision_test.go
@@ -42,13 +42,23 @@ import (
 // because the pod template did not change and the workload has no reason to restart.
 // These specs pin that boundary.
 var _ = Describe("RoleInstanceSet revision stability", func() {
-	const ns = "default"
+	var testNs string
+
+	BeforeEach(func() {
+		testNs = fmt.Sprintf("test-revision-%d", time.Now().UnixNano())
+		testutil.CreateNamespace(testNs)
+	})
+
+	AfterEach(func() {
+		// Unconditional, so a failed assertion cannot leak the set into later specs.
+		testutil.DeleteNamespace(testNs)
+	})
 
 	// legacyShapedSet stores the policy through the deprecated string field, which
 	// is the shape a pre-restartPolicyConfig release wrote.
 	legacyShapedSet := func(name string) *workloadsv1alpha2.RoleInstanceSet {
 		return &workloadsv1alpha2.RoleInstanceSet{
-			ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: ns},
+			ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: testNs},
 			Spec: workloadsv1alpha2.RoleInstanceSetSpec{
 				Replicas: ptr.To(int32(1)),
 				Selector: &metav1.LabelSelector{MatchLabels: map[string]string{"app": name}},
@@ -73,7 +83,7 @@ var _ = Describe("RoleInstanceSet revision stability", func() {
 
 	instanceUIDs := func(setName string) map[string]types.UID {
 		list := &workloadsv1alpha2.RoleInstanceList{}
-		Expect(testutil.K8sClient.List(testutil.Ctx, list, client.InNamespace(ns))).To(Succeed())
+		Expect(testutil.K8sClient.List(testutil.Ctx, list, client.InNamespace(testNs))).To(Succeed())
 		out := map[string]types.UID{}
 		for i := range list.Items {
 			ri := &list.Items[i]
@@ -88,7 +98,7 @@ var _ = Describe("RoleInstanceSet revision stability", func() {
 
 	revisionsFor := func(setName string) []string {
 		crList := &apps.ControllerRevisionList{}
-		Expect(testutil.K8sClient.List(testutil.Ctx, crList, client.InNamespace(ns))).To(Succeed())
+		Expect(testutil.K8sClient.List(testutil.Ctx, crList, client.InNamespace(testNs))).To(Succeed())
 		var names []string
 		for i := range crList.Items {
 			cr := &crList.Items[i]
@@ -112,10 +122,23 @@ var _ = Describe("RoleInstanceSet revision stability", func() {
 			return len(before)
 		}, 60*time.Second, time.Second).Should(BeNumerically(">", 0))
 
-		stored := &workloadsv1alpha2.RoleInstanceSet{}
-		key := client.ObjectKey{Namespace: ns, Name: name}
-		Expect(testutil.K8sClient.Get(testutil.Ctx, key, stored)).To(Succeed())
-		beforeRevision := stored.Status.CurrentRevision
+		key := client.ObjectKey{Namespace: testNs, Name: name}
+
+		// Wait for the revision to be published before capturing it. If it were
+		// still empty here, the closing assertion would be satisfied by
+		// UpdateRevision merely becoming non-empty and would not show a
+		// re-revision at all.
+		By("waiting for status.currentRevision to be published")
+		var beforeRevision string
+		Eventually(func() string {
+			stored := &workloadsv1alpha2.RoleInstanceSet{}
+			if err := testutil.K8sClient.Get(testutil.Ctx, key, stored); err != nil {
+				return ""
+			}
+			beforeRevision = stored.Status.CurrentRevision
+			return beforeRevision
+		}, 60*time.Second, time.Second).ShouldNot(BeEmpty(),
+			"status.currentRevision must be populated before it can be compared")
 		GinkgoWriter.Printf("before: instances=%v currentRevision=%q revisions=%v\n",
 			before, beforeRevision, revisionsFor(name))
 
@@ -148,10 +171,13 @@ var _ = Describe("RoleInstanceSet revision stability", func() {
 
 		// The revision does move, which is what makes the check above meaningful:
 		// the set really was re-revisioned and still did not roll.
+		Expect(after.Status.UpdateRevision).NotTo(BeEmpty(),
+			"status.updateRevision should be published after the template changed")
 		Expect(after.Status.UpdateRevision).NotTo(Equal(beforeRevision),
 			"the reserialized template should produce a new revision")
 
-		Expect(testutil.K8sClient.Delete(testutil.Ctx, after)).To(Succeed())
+		// No explicit Delete: AfterEach removes the namespace, so cleanup does not
+		// depend on reaching the end of the spec.
 	})
 
 	It("resolves the same effective policy from either field shape", func() {
--- a/test/envtest/testcase/restart_policy/restart_policy_revision_test.go
+++ b/test/envtest/testcase/restart_policy/restart_policy_revision_test.go
@@ -1,0 +1,179 @@
+/*
+Copyright 2026 The RBG Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package restart_policy
+
+import (
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	apps "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


test/envtest/testcase/restart_policy/restart_policy_revision_test.go
```diff
@@ -0,0 +1,205 @@
...
+		Expect(after.Status.UpdateRevision).NotTo(BeEmpty(),
+			"status.updateRevision should be published after the template changed")
+		Expect(after.Status.UpdateRevision).NotTo(Equal(beforeRevision),
+			"the reserialized template should produce a new revision")
+
+		// No explicit Delete: AfterEach removes the namespace, so cleanup does not
+		// depend on reaching the end of the spec.
+	})
+
+	It("resolves the same effective policy from either field shape", func() {
```
cheyang, line 183: This second spec is a pure in-memory check of GetRestartPolicy precedence — it never talks to the API server, so it doesn't really need the envtest harness. The same three cases (legacy string, config type, and config type winning over the deprecated field) are already covered by the table tests in api/workloads/v1alpha2/restartpolicy_test.go (TestRoleInstanceSpec_GetRestartPolicy_NoSyntheticDefault, and the "config type wins over legacy field" case in TestRoleSpec_GetRestartPolicyConfig). I'd drop it here and keep this file focused on the cluster-level revision-stability behavior that the unit tests can't reach. Non-blocking.

cheyang (commented):
Test-only change, and the main spec is the valuable part: it pins that a restartPolicy -> restartPolicyConfig serialization change re-revisions the set without recreating the RoleInstances, and it guards against a trivial pass by separately asserting updateRevision actually moved. The precedence assertions match resolveRestartPolicyConfig, and cleanup is unconditional via AfterEach. One minor note inline about the second spec duplicating existing unit coverage. No blockers.

test/envtest/testcase/restart_policy/restart_policy_revision_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+// A RoleInstanceSet's ControllerRevision is derived from the whole
+// roleInstanceTemplate, so any change to how a field is serialized moves the
+// revision hash even when the effective configuration is identical. That happens
+// on a controller upgrade: an older release stored the restart policy as the
+// `restartPolicy` string, this one stores it under `restartPolicyConfig`.
+//
+// A revision change on its own is acceptable. Recreating the RoleInstances is not,
+// because the pod template did not change and the workload has no reason to restart.
+// These specs pin that boundary.
```
diw-zw, line 43: Fidelity gap worth a sentence here. The real upgrade path is the RBG reconciler re-rendering the RoleInstanceSet template via server-side apply (`pkg/reconciler/roleinstanceset_reconciler.go:163-168`), which always writes `restartPolicyConfig` and never the legacy string. This file hand-builds an unowned, legacy-shaped RIS (line 116) and hand-edits it (lines 145-157), bypassing that path entirely. A future reader will assume the upgrade path is covered; please note that it isn't.

test/envtest/testcase/restart_policy/restart_policy_revision_test.go
```diff
@@ -0,0 +1,205 @@
...
+		for i := range list.Items {
+			ri := &list.Items[i]
+			for _, o := range ri.OwnerReferences {
+				if o.Name == setName {
+					out[ri.Name] = ri.UID
+				}
+			}
+		}
+		return out
+	}
```
diw-zw, line 97: This asserts one layer above the thing the file claims to protect. Line 42 says "the pod template did not change and the workload has no reason to restart" — that is a statement about **pods**, but this helper only collects RoleInstance UIDs.

In-place update does churn the RoleInstance: it rewrites `.Spec`, writes the `InPlaceUpdateState` annotation, and flips `InPlaceUpdateReady=False` when a readiness gate is present (`pkg/inplace/instance/inplaceupdate/inplace_update.go:138-155`). So RoleInstance UID stability is not the same claim as "the workload did not restart". Pod UID stability is the observable users care about, it is cheap to collect in envtest, and it should hold in this scenario — please assert that too.

Minor: the owner match is by `o.Name` only and does not `break` after a hit. The per-spec namespace makes collisions unlikely, but a `Kind` check plus `break` is one line each.

test/envtest/testcase/restart_policy/restart_policy_revision_test.go
```diff
@@ -0,0 +1,205 @@
...
+		for i := range crList.Items {
+			cr := &crList.Items[i]
+			for _, o := range cr.OwnerReferences {
+				if o.Name == setName {
+					names = append(names, fmt.Sprintf("%s(rev=%d)", cr.Name, cr.Revision))
+				}
+			}
+		}
+		return names
+	}
```
diw-zw, line 112: `revisionsFor` is ~14 lines that exist solely to feed the two `GinkgoWriter.Printf` calls on lines 143 and 170. Keeping it is fine, but please label it as debug-only so nobody later mistakes it for part of the assertions.

test/envtest/testcase/restart_policy/restart_policy_revision_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+	It("keeps RoleInstances when only the restart policy serialization changes", func() {
```
diw-zw, line 114: The title claims serialization-specificity that the body cannot deliver.

In-place eligibility for a RoleInstance is decided entirely by `canNotInPlaceUpdate` (`pkg/inplace/instance/inplaceupdate/inplace_update_defaults.go:140`):

```go
func canNotInPlaceUpdate(newTemplate, oldTemplate *workloadsv1alpha2.RoleInstanceTemplate) bool {
	return componentSizeChanges(newTemplate.Components, oldTemplate.Components)
}
```

Only component count/size changes fall back to recreate. Everything else — image, env, `restartPolicy`, whatever — goes through the in-place branch of `applyTargetUpdate` (`pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control.go:786`) and the UID survives. Change line 153 to `RecreateRoleInstanceOnPodRestart`, i.e. a genuine behavioural change rather than a reserialization, and this spec still passes. So it does not pin "serialization-only changes are safe"; it pins "template changes are handled in-place", which is a different and much weaker statement.

If the property you want is "a restartPolicy shape change stays in-place", assert it where the decision is made: build two `ControllerRevision`s differing only in restart policy shape and assert `defaultCalculateInPlaceUpdateSpec` returns non-nil. That is ~20 deterministic lines in `pkg/inplace/instance/inplaceupdate/inplace_update_defaults_test.go`, needs no apiserver, and fails loudly the moment someone widens `canNotInPlaceUpdate` — which is the regression the description is actually worried about.

test/envtest/testcase/restart_policy/restart_policy_revision_test.go
```diff
@@ -0,0 +1,205 @@
...
+		// UpdateRevision merely becoming non-empty and would not show a
+		// re-revision at all.
+		By("waiting for status.currentRevision to be published")
+		var beforeRevision string
+		Eventually(func() string {
+			stored := &workloadsv1alpha2.RoleInstanceSet{}
+			if err := testutil.K8sClient.Get(testutil.Ctx, key, stored); err != nil {
+				return ""
+			}
+			beforeRevision = stored.Status.CurrentRevision
```
diw-zw, line 138: The baseline is captured from `Status.CurrentRevision` but line 176 compares it against `Status.UpdateRevision`. Those are different fields; the comparison only works because the two coincide immediately after creation. Capture `Status.UpdateRevision` here so the comparison is field-consistent and does not quietly break if the controller ever advances `CurrentRevision` on a different schedule.

test/envtest/testcase/restart_policy/restart_policy_revision_test.go
```diff
@@ -0,0 +1,205 @@
...
+		Eventually(func() error {
+			live := &workloadsv1alpha2.RoleInstanceSet{}
+			if err := testutil.K8sClient.Get(testutil.Ctx, key, live); err != nil {
+				return err
+			}
+			live.Spec.RoleInstanceTemplate.RestartPolicy = "" //nolint:staticcheck // clearing the pre-upgrade shape
+			live.Spec.RoleInstanceTemplate.RestartPolicyConfig = &workloadsv1alpha2.RestartPolicyConfig{
+				Type:             workloadsv1alpha2.RestartPolicyNone,
+				BaseDelaySeconds: ptr.To(int32(30)),
+				MaxDelaySeconds:  ptr.To(int32(600)),
```
diw-zw, line 155: These are the CRD defaults (`api/workloads/v1alpha2/rolebasedgroup_types.go:178,186`), so setting them explicitly is redundant on its own. That said, they do mirror what the RBG reconciler writes (`pkg/reconciler/roleinstanceset_reconciler.go:163-166`), which is a good reason to keep them. Worth a one-line comment saying so, otherwise the next reader will delete them as noise.

test/envtest/testcase/restart_policy/restart_policy_revision_test.go
```diff
@@ -0,0 +1,205 @@
...
+			}
+			return testutil.K8sClient.Update(testutil.Ctx, live)
+		}, 30*time.Second, time.Second).Should(Succeed())
+
+		By("checking the RoleInstances are not recreated")
+		Consistently(func() map[string]types.UID {
+			return instanceUIDs(name)
+		}, 45*time.Second, 3*time.Second).Should(Equal(before),
+			"a serialization-only change must not recreate the RoleInstances")
```
diw-zw, line 164: The ordering race from the earlier automated review still stands — the second commit fixed the `currentRevision` capture but not this.

This 45s window opens before the controller has observed the update from line 157, and whether the revision actually moved is not checked until the one-shot `Get` on lines 166-176, *after* the window closes. Two consequences:

1. The entire window can be consumed pre-reconcile, so UID stability is never observed across the actual revision transition — the spec passes having proved nothing.
2. If reconcile lags past the window, the trailing assertion on line 176 flakes.

Reorder so the window covers the transition: `Eventually` until `Status.UpdateRevision` leaves the baseline, *then* `Consistently` on the UIDs. That fixes both failure modes and lets you cut the window to ~15s.

test/envtest/testcase/restart_policy/restart_policy_revision_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+		// An explicit config type wins over the deprecated field.
+		both := workloadsv1alpha2.RoleInstanceSpec{
+			RestartPolicy: workloadsv1alpha2.RecreateRoleInstanceOnPodRestart, //nolint:staticcheck
+			RestartPolicyConfig: &workloadsv1alpha2.RestartPolicyConfig{
+				Type: workloadsv1alpha2.RestartPolicyNone,
+			},
+		}
+		Expect(both.GetRestartPolicy()).To(Equal(workloadsv1alpha2.RestartPolicyNone))
+	})
```
diw-zw, line 204: Please drop this spec. It never touches the cluster, yet it pays the `BeforeEach`/`AfterEach` namespace create-and-delete (lines 47-55) and the whole envtest apiserver bootstrap.

It is also strictly weaker than existing coverage: `api/workloads/v1alpha2/restartpolicy_test.go:154` (`TestRoleInstanceSpec_GetRestartPolicy_NoSyntheticDefault`) already asserts the deprecated field (line 187 here), the config type (line 194), and the precedence (line 203) — plus the unset-stays-empty case this spec omits.

diw-zw (commented):
Thanks for writing this down — the upgrade concern in the description is real and worth pinning. My reservation is about what the specs actually assert versus what the description claims they assert. Details are inline; summarising the two that matter most.

**The main spec passes for a reason unrelated to serialization.** In-place eligibility is decided by `canNotInPlaceUpdate` (`pkg/inplace/instance/inplaceupdate/inplace_update_defaults.go:140`), which rejects in-place *only* when component count or size changes. Every template-only change — image, env, `restartPolicy`, serialization-shaped or semantically different — therefore takes the in-place path and the RoleInstance UID never changes. Flip line 153 from `RestartPolicyNone` to `RecreateRoleInstanceOnPodRestart` and the spec still passes, so it cannot distinguish "same effective config, new hash" from "config actually changed" — which is the entire premise.

**Runtime cost.** Lines 123, 140, 158 and 163 add roughly 60s + 60s + 30s + 45s of async budget to `make test-envtest` (`Makefile:87-91`). Fixing the ordering race and dropping the second spec cuts that to about a third while asserting strictly more.

Net recommendation: replace the envtest spec with a unit test on `defaultCalculateInPlaceUpdateSpec`, add pod-UID stability if you still want an integration-level guard, and delete the second spec. Nothing here is a correctness bug in production code — it's about whether this test earns its runtime and pins the stated boundary.

PR #429, opened 2026-08-19
chore(deps): upgrade dependencies, kubectl v1.36.4, and migrate golangci-lint v2 to fix CVEs

diw-zw:
## What this PR does
Remediate CVEs by updating dependencies and toolchain, rather than
rewriting the CRD upgrader. The kubectl-based crd-upgrade flow is kept.

- Bump Go module dependencies (go.mod/go.sum/vendor) and align OTel to v1.43.0
- Upgrade CRD-upgrader kubectl: v1.31.0 → v1.36.4
- Bump CI Go toolchain to 1.26.6 with GOTOOLCHAIN=local
- Migrate golangci-lint config to v2 (+ resulting gofmt/goimports reformat)
- Fix controller requeue backoff (split error / RequeueAfter branches)

## Notes
- crd-upgrade remains kubectl-based (no Go rewrite).
- kubectl v1.36.4 is latest stable; ±1 minor skew from apiserver applies.


.github/workflows/envtest.yml
```diff
@@ -7,7 +7,8 @@ on:
     branches: [ "main" ]
 
 env:
-  GO_VERSION: 1.24.1
+  GO_VERSION: 1.25.0
```
cheyang: The `Envtest` job fails on this PR because of this pin. The log shows:

```
sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.24.1 requires go >= 1.26.0 (running go 1.25.0; GOTOOLCHAIN=local)
```

The dependency bump pulls in tooling that needs Go 1.26, but the workflow sets `GO_VERSION: 1.25.0` together with `GOTOOLCHAIN: local`, so the runner can't upgrade the toolchain to build `setup-envtest`. Worth noting the build images were already moved to `golang:1.26.6` (Dockerfile and tools/crd-upgrade/Dockerfile), so the CI toolchain is now a step behind the image the release actually ships on. Please align `GO_VERSION` here (and in the other four workflows) with 1.26.x, or drop `GOTOOLCHAIN: local` so `auto` can fetch the required toolchain.

.github/workflows/project-check.yml
```diff
@@ -7,7 +7,8 @@ on:
     branches: [ "main" ]
 
 env:
-  GO_VERSION: 1.24.1
+  GO_VERSION: 1.25.0
```
cheyang: Same root cause takes down the `lint` job here:

```
github.com/golangci/golangci-lint/v2@v2.13.1 requires go >= 1.26.0 (running go 1.25.0; GOTOOLCHAIN=local)
```

The Makefile bump to `GOLANGCI_LINT_VERSION ?= v2.13.1` (Makefile:323) requires Go 1.26, but this workflow still pins `GO_VERSION: 1.25.0` with `GOTOOLCHAIN: local`. Bump `GO_VERSION` to 1.26.x here as well (matching the `golang:1.26.6` Dockerfiles), or relax `GOTOOLCHAIN`. Both `envtest` and `lint` need to go green before this can merge.

cheyang (changes requested):
Thanks for the CVE cleanup — swapping the kubectl-based CRD upgrader for a small Go binary built from the same vendored tree is a solid direction, and the `main.go` logic (create-or-replace each CRD, then re-apply the conversion webhook via `client.MergeFrom`) reads correctly and mirrors the old script's behavior.

One thing blocks merge: the Go toolchain version is inconsistent across the build setup, and it's breaking CI. The dependency bump raises the minimum Go for two build tools to 1.26 (`golangci-lint v2.13.1` and `setup-envtest v0.24.1`), and the Dockerfiles were bumped to `golang:1.26.6`, but the five workflows only moved `GO_VERSION` to 1.25.0 with `GOTOOLCHAIN: local`. As a result the `lint` and `envtest` jobs both fail with `requires go >= 1.26.0 (running go 1.25.0; GOTOOLCHAIN=local)`. Bumping `GO_VERSION` to 1.26.x across all five workflows (to match the Dockerfiles) — or setting `GOTOOLCHAIN=auto` — should clear both. Details inline.

cheyang (approved):
Previous CI blocker is resolved. GO_VERSION is now 1.26.6 across all five workflow files, matching the Dockerfile builder image. GOTOOLCHAIN=local is retained so the build remains pinned.

The golangci-lint v2 migration looks well-structured: exclusion path regexes are properly anchored, goconst/staticcheck settings are sensible (ignore test literals, whitelist Ginkgo/Gomega dot-imports), and the formatter section correctly hosts gofmt/goimports (which are no longer linters in v2).

The two controller fixes (`rolebasedgroup_controller.go`, `roleinstanceset_controller.go`) are a genuine improvement: splitting `err || requeue` ensures controller-runtime properly propagates the error without ignoring it due to a non-zero Result.

setup-envtest is now pinned to v0.24.1 (was `@latest`), improving reproducibility.

All 11 CI checks pass. LGTM.

PR #430, opened 2026-08-20
KEP-430: Gang Scheduling for RBG

diw-zw:
### Ⅰ. Motivation

Currently, RBG's gang scheduling only supports the RBG level (minMember) — all pods must be scheduled simultaneously or all wait. In PD-disaggregated (Prefill-Decode disaggregated) inference scenarios, this "all-or-nothing" constraint is too strict: when resources are insufficient to schedule all pods, the service cannot start at all, even though a partial deployment (e.g., 2 prefill + 1 decode) would be sufficient to run the inference service.

This KEP introduces **two-level gang scheduling** for RBG:
1. **RBG level** (Group level): All pods must be scheduled simultaneously or all wait (existing behavior).
2. **Role level** (SubGroup level): Each role can independently configure a minimum number of replicas to schedule (e.g., prefill ≥ 2, decode ≥ 1), with both constraints satisfied simultaneously.

### Ⅱ. Modifications

This PR adds the KEP document (`keps/430-gang-scheduling/`) with the following design:

- **API**: New `scheduling.gang` sub-strategy in CoordinatedPolicy's `SchedulingCoordinationStrategy` domain, with `minReplicas map[string]int32` for per-role gang constraints.
- **Scheduler Abstraction**: `PodGroupManager` interface split into:
  - `BasicGangScheduling` — base interface (minMember = GetGroupSize, pod.spec.schedulerName injection)
  - `RoleBasedGangScheduling` — optional interface for per-role gang (e.g., Volcano subGroupPolicy)
  - Capability discovery via Go type assertion (no `SupportsSubGroups() bool`)
  - `GangSchedulingStrategy` passed directly to scheduler implementation (no intermediate type)
- **Webhook Validation**: Self-contained validations (key validity, value > 0, scheduler capability). Rejects CoordinatedPolicy creation when `minReplicas` is non-empty but the scheduler does not support subGroup.
- **Runtime Safety Net**: Returns error (no degradation) when scheduler is switched after CoordinatedPolicy creation.
- **pod.spec.schedulerName**: Controller auto-injects via `GetSchedulerName()`; users no longer need to manually set it in pod templates.
- **Volcano Reference Implementation**: Translates `GangSchedulingStrategy` to PodGroup `subGroupPolicy` with field mapping.
- **GetGroupSize Fix**: Corrects calculation for `CustomComponentsPattern` (sum of component sizes).
- **Backward Compatibility**: Annotation `group-gang-scheduling=true` preserved for users not using CoordinatedPolicy.

### DIFF
--- a/keps/430-gang-scheduling/README.md
+++ b/keps/430-gang-scheduling/README.md
@@ -1,0 +1,613 @@
+# KEP-430: Gang Scheduling
+
+## Table of Contents
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+    - [User Stories](#user-stories)
+        - [Story 1: Partial Gang for PD-Disaggregated Inference](#story-1-partial-gang-for-pd-disaggregated-inference)
+        - [Story 2: Basic Gang Scheduling](#story-2-basic-gang-scheduling)
+    - [Design Overview](#design-overview)
+    - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+    - [API: CoordinatedPolicy Scheduling Strategy](#api-coordinatedpolicy-scheduling-strategy)
+    - [Scheduler Abstraction](#scheduler-abstraction)
+    - [Webhook and Capability Validation](#webhook-and-capability-validation)
+    - [Volcano Implementation (Reference)](#volcano-implementation-reference)
+    - [Controller Flow](#controller-flow)
+    - [GetGroupSize Enhancement](#getgroupsize-enhancement)
+    - [Label and Annotation Conventions](#label-and-annotation-conventions)
+    - [Backward Compatibility](#backward-compatibility)
+    - [Test Plan](#test-plan)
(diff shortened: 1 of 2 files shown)


NoobDream2568:
Besides the `minMember` field, users deploying inference services on clusters with hyper nodes or complex topologies may also want the affinity as well — i.e., placing pods within a single hyper node to improve performance.

Volcano provides the `networkTopology` field for this purpose. Should we expose a similar field in the API, even though it is not a generic one?

```yaml
networkTopology:
  mode: hard
  highestTierAllowed: 1
```

diw-zw:
> Besides the `minMember` field, users deploying inference services on clusters with hyper nodes or complex topologies may also want the affinity as well — i.e., placing pods within a single hyper node to improve performance.
> 
> Volcano provides the `networkTopology` field for this purpose. Should we expose a similar field in the API, even though it is not a generic one?
> 
> ```yaml
> networkTopology:
>   mode: hard
>   highestTierAllowed: 1
> ```

This is a great feature, but it falls outside the scope of the current gang scheduling discussion. We might consider adding a separate KEP for it in the future.

cheyang (approved):
/lgtm
/approve

PR #431, opened 2026-08-21
fix: validate rolling strategy only when replicas gt 0

vie-serendipity:
### Ⅰ. Motivation

Bug fix. When the replica count is 0 and `maxUnavailable` is set to percent format, scaling the replicas from 1 to 0 gets stuck.

### Ⅱ. Modifications

Validate maxSurge and maxUnavailable only when replicas gt 0.

fixes #XXXX

## Checklist

### DIFF
--- a/pkg/reconciler/sts_reconciler.go
+++ b/pkg/reconciler/sts_reconciler.go
@@ -640,7 +640,7 @@ func validateRolloutStrategy(
 	if err != nil {
 		return nil, err
 	}
-	if maxUnavailable == 0 && maxSurge == 0 {
+	if replicas > 0 && maxUnavailable == 0 && maxSurge == 0 {
 		return nil, fmt.Errorf(
 			"RollingUpdate is invalid: " +
 				"rolloutStrategy.rollingUpdate.maxUnavailable may not be 0 when maxSurge is 0",
--- a/pkg/reconciler/sts_reconciler_test.go
+++ b/pkg/reconciler/sts_reconciler_test.go
@@ -38,6 +38,48 @@ import (
 	wrappersv2 "sigs.k8s.io/rbgs/test/wrappers/v1alpha2"
 )
 
+func TestValidateRolloutStrategyAllowsZeroReplicas(t *testing.T) {
+	tests := []struct {
+		name           string
+		maxUnavailable intstr.IntOrString
+	}{
+		{name: "percentage", maxUnavailable: intstr.FromString("30%")},
+		{name: "integer", maxUnavailable: intstr.FromInt32(0)},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			strategy := &workloadsv1alpha2.RolloutStrategy{
+				Type: workloadsv1alpha2.RollingUpdateStrategyType,
+				RollingUpdate: &workloadsv1alpha2.RollingUpdate{
+					MaxUnavailable: ptr.To(tt.maxUnavailable),
+					MaxSurge:       ptr.To(intstr.FromInt32(0)),
+					Partition:      ptr.To(intstr.FromInt32(0)),
+				},
+			}
+
+			_, err := validateRolloutStrategy(strategy, 0)
+
+			assert.NoError(t, err)
+		})
+	}
+}
+
+func TestValidateRolloutStrategyRejectsZeroBudgetWhenReplicasGreaterThanZero(t *testing.T) {
+	strategy := &workloadsv1alpha2.RolloutStrategy{
+		Type: workloadsv1alpha2.RollingUpdateStrategyType,
+		RollingUpdate: &workloadsv1alpha2.RollingUpdate{
+			MaxUnavailable: ptr.To(intstr.FromInt32(0)),
+			MaxSurge:       ptr.To(intstr.FromInt32(0)),
+			Partition:      ptr.To(intstr.FromInt32(0)),
+		},
+	}
+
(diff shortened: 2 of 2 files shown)


diw-zw (commented):
The production fix in `pkg/reconciler/sts_reconciler.go` is correct. When `replicas=0`, percentage-based `maxUnavailable` resolves to `0` via `GetScaledValueFromIntOrPercent`, so the `maxUnavailable == 0 && maxSurge == 0` validation must be skipped because there are no pods to roll.

Two suggestions to strengthen the test coverage:

1. **Also test the integer `maxUnavailable=0` path.**
   The new `TestValidateRolloutStrategyAllowsZeroReplicas` only exercises `intstr.FromString("30%")`. Since the guard now applies to any `maxUnavailable=0` when `replicas=0`, consider adding the integer form:

   ```go
   func TestValidateRolloutStrategyAllowsZeroReplicas(t *testing.T) {
       cases := []struct {
           name           string
           maxUnavailable intstr.IntOrString
       }{
           {"percentage", intstr.FromString("30%")},
           {"integer", intstr.FromInt32(0)},
       }

       for _, tc := range cases {
           t.Run(tc.name, func(t *testing.T) {
               strategy := &workloadsv1alpha2.RolloutStrategy{
                   Type: workloadsv1alpha2.RollingUpdateStrategyType,
                   RollingUpdate: &workloadsv1alpha2.RollingUpdate{
                       MaxUnavailable: ptr.To(tc.maxUnavailable),
                       MaxSurge:       ptr.To(intstr.FromInt32(0)),
                       Partition:      ptr.To(intstr.FromInt32(0)),
                   },
               }
               _, err := validateRolloutStrategy(strategy, 0)
               assert.NoError(t, err)
           })
       }
   }
   ```

2. **Add a regression test for the preserved error path.**
   No test asserts that `replicas > 0` with both budgets at `0` still returns the invalid-strategy error. This guards against a future refactor accidentally removing the `replicas > 0` guard:

   ```go
   func TestValidateRolloutStrategyRejectsZeroBudgetWhenReplicasGtZero(t *testing.T) {
       strategy := &workloadsv1alpha2.RolloutStrategy{
           Type: workloadsv1alpha2.RollingUpdateStrategyType,
           RollingUpdate: &workloadsv1alpha2.RollingUpdate{
               MaxUnavailable: ptr.To(intstr.FromInt32(0)),
               MaxSurge:       ptr.To(intstr.FromInt32(0)),
               Partition:      ptr.To(intstr.FromInt32(0)),
           },
       }

       _, err := validateRolloutStrategy(strategy, 1)

       assert.Error(t, err)
       assert.Contains(t, err.Error(), "maxUnavailable may not be 0 when maxSurge is 0")
   }
   ```

Minor nit: the PR title and commit message have a typo — `"stratefy"` should be `"strategy"`.

vie-serendipity:
@diw-zw done

cheyang (approved):
LGTM. @diw-zw's two test requests are both in on `cd6c9d4c` and they do what they're meant to — I verified rather than eyeballed:

```
BASE 993491fb (production code) + this PR's test file grafted on:
  --- FAIL: TestValidateRolloutStrategyAllowsZeroReplicas/percentage
  --- FAIL: TestValidateRolloutStrategyAllowsZeroReplicas/integer
        Error: RollingUpdate is invalid: ...maxUnavailable may not be 0 when maxSurge is 0
  --- PASS: TestValidateRolloutStrategyRejectsZeroBudgetWhenReplicasGreaterThanZero

HEAD cd6c9d4c: all PASS
```

Both new subtests fail on the base with the exact error from the bug report, while the control passes on the base too — so the test file is detecting this specific bug rather than being broadly broken, and the control genuinely guards against the guard being dropped later. `pkg/reconciler` and `api/workloads/v1alpha2` are both green on the head.

Three things I checked that I think strengthen the case for this fix beyond the report:

**The webhook already established exactly this rule, so the reconciler was the inconsistent one.** `ValidateRollingUpdate` in `api/workloads/v1alpha2/rolebasedgroup_validation.go` skips a role when `*role.Replicas == 0`, and its doc comment says so outright: "validates each role's rolloutStrategy *when replicas is greater than 0*". So admission has always accepted scaling to 0 while the reconciler refused to carry it out — which is exactly the shape of a stuck scale-down. This PR aligns the two rather than inventing a new exemption, which to me is the strongest argument that the direction is right.

**The fix lands in the shared helper, so it covers both callers.** `validateRolloutStrategy` is called from `sts_reconciler.go:94` and `roleinstanceset_reconciler.go:87`, so RoleInstanceSet-backed roles get the fix too without a second edit. Testing the helper directly is the right level for this.

**`replicas > 0` is the precise boundary, not a rounded-off guess.** When `maxSurge == 0` the `maxUnavailable` percentage is scaled with `roundUp=true`, so for any `replicas >= 1` even a tiny percentage resolves to at least 1. `replicas == 0` is the only input that can make a percentage resolve to 0, so the guard admits exactly the broken case and nothing more.

I also checked for the same bug elsewhere: `lws_reconciler.go` only uses `GetScaledValueFromIntOrPercent` for `partition` and has no both-budgets-zero check, so there's no sibling copy to fix.

Non-blocking, and neither needs to hold up the merge:

- The `stratefy` → `strategy` typo @diw-zw flagged is still in the PR title and in the commit subject on `cd6c9d4c` — the "done" seems to have covered the tests only. The title is worth editing since it ends up in the squashed commit.
- The body still has the template placeholder `fixes #XXXX`. If there's no issue, `NONE` is the convention here; if there is one, wiring it up would close it on merge.

Approving. Nice tight fix with a real regression guard.

PR #432, opened 2026-08-24
Add in-place scheduling injection logs

googs1025:
## Summary
- Log when in-place scheduling skips affinity injection because no node binding exists
- Log successful Preferred and Required node affinity injection with mode, granularity, key, and target nodes

## Why
In-place scheduling bindings are currently kept in memory. After a controller restart or leader failover, a recreated pod may have `role-inplace-scheduling: Required` but no binding available in the new controller process. In that case the controller intentionally skips affinity injection, and the resulting pod has an empty node affinity.

Without an explicit log, operators only see Kubernetes events showing the replacement pod was created and scheduled. That makes it hard to distinguish scheduler behavior from the controller having no recovered binding. The new logs make this path visible and also record which nodes are injected when a binding is present.

## Testing
- `GOCACHE=/private/tmp/rbg-go-cache GOMODCACHE=/private/tmp/rbg-go-modcache go test ./pkg/reconciler/roleinstance/sync -count=1`

### DIFF
--- a/pkg/reconciler/roleinstance/sync/node_binding.go
+++ b/pkg/reconciler/roleinstance/sync/node_binding.go
@@ -332,6 +332,9 @@ func InjectInPlaceScheduling(pod *v1.Pod, instance *workloadsv1alpha2.RoleInstan
 	// 6. Load binding
 	nodes := store.Load(key)
 	if len(nodes) == 0 {
+		klog.InfoS("in-place scheduling: no node binding found, skip hostname affinity injection",
+			"pod", klog.KObj(pod), "instance", klog.KObj(instance),
+			"mode", mode, "granularity", granularity, "key", key)
 		// No binding — if avoid is configured, fold it into every existing
 		// required term (AND semantics). If no required terms exist, add it
 		// as a standalone term.
@@ -362,6 +365,9 @@ func InjectInPlaceScheduling(pod *v1.Pod, instance *workloadsv1alpha2.RoleInstan
 		}
 		nodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution = append(
 			nodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution, term)
+		klog.V(4).InfoS("in-place scheduling: injected preferred node affinity",
+			"pod", klog.KObj(pod), "instance", klog.KObj(instance),
+			"granularity", granularity, "key", key, "nodes", values)
 
 		// Fold avoid into every required term (AND semantics). If no
 		// required terms exist, a standalone avoid term is created.
@@ -387,6 +393,9 @@ func InjectInPlaceScheduling(pod *v1.Pod, instance *workloadsv1alpha2.RoleInstan
 		}
 		inPlaceExprs = append(inPlaceExprs, avoidExprs...)
 		foldIntoRequired(nodeAffinity, inPlaceExprs)
+		klog.V(4).InfoS("in-place scheduling: injected required node affinity",
+			"pod", klog.KObj(pod), "instance", klog.KObj(instance),
+			"granularity", granularity, "key", key, "nodes", values)
 	}
 }
 


pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -332,6 +332,9 @@ func InjectInPlaceScheduling(pod *v1.Pod, instance *workloadsv1alpha2.RoleInstan
 	// 6. Load binding
 	nodes := store.Load(key)
 	if len(nodes) == 0 {
+		klog.InfoS("in-place scheduling: no node binding found, skip affinity injection",
+			"pod", klog.KObj(pod), "instance", klog.KObj(instance),
+			"mode", mode, "granularity", granularity, "key", key)
```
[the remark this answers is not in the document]
googs1025: Fixed.

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -362,6 +366,9 @@ func InjectInPlaceScheduling(pod *v1.Pod, instance *workloadsv1alpha2.RoleInstan
 		}
 		nodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution = append(
 			nodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution, term)
+		klog.InfoS("in-place scheduling: injected preferred node affinity",
+			"pod", klog.KObj(pod), "instance", klog.KObj(instance),
+			"mode", mode, "granularity", granularity, "key", key, "nodes", values)
```
NoobDream2568: Since L369 already said "injected preferred node affinity", there's no need to log `mode`.

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -387,6 +394,9 @@ func InjectInPlaceScheduling(pod *v1.Pod, instance *workloadsv1alpha2.RoleInstan
 		}
 		inPlaceExprs = append(inPlaceExprs, avoidExprs...)
 		foldIntoRequired(nodeAffinity, inPlaceExprs)
+		klog.InfoS("in-place scheduling: injected required node affinity",
+			"pod", klog.KObj(pod), "instance", klog.KObj(instance),
+			"mode", mode, "granularity", granularity, "key", key, "nodes", values)
```
NoobDream2568: Same here. L397 already said "injected required node affinity", there's no need to log `mode`.

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -332,6 +332,10 @@ func InjectInPlaceScheduling(pod *v1.Pod, instance *workloadsv1alpha2.RoleInstan
 	// 6. Load binding
 	nodes := store.Load(key)
 	if len(nodes) == 0 {
+		klog.InfoS("in-place scheduling: no node binding found, skip hostname affinity injection",
+			"pod", klog.KObj(pod), "instance", klog.KObj(instance),
+			"mode", mode, "granularity", granularity, "key", key,
+			"avoidExpressions", len(avoidExprs))
```
NoobDream2568: Logging the length of `avoidExprs` provides limited observability value here. However, logging the full content directly may produce overly verbose logs.

So, maybe just drop it.

NoobDream2568 (changes requested):
Overall, this PR improves observability for the inplace-scheduling feature by logging affinity injection success and failure, which helps users troubleshoot through controller logs instead of inspecting Pod YAML directly.

It would be good to further refine the log content.

pkg/reconciler/roleinstance/sync/node_binding.go
googs1025: Fixed. Removed the redundant `mode` field from the preferred injection log because the message already identifies it as preferred.

pkg/reconciler/roleinstance/sync/node_binding.go
googs1025: Fixed. Dropped `avoidExpressions` from the no-binding log and kept the message focused on skipping hostname affinity injection.

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -362,6 +365,9 @@ func InjectInPlaceScheduling(pod *v1.Pod, instance *workloadsv1alpha2.RoleInstan
 		}
 		nodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution = append(
 			nodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution, term)
+		klog.InfoS("in-place scheduling: injected preferred node affinity",
```
cheyang: This fires on every successful injection, which is the normal happy path per pod. The other in-place scheduling traces in this file are gated behind `klog.V(4).InfoS` (see the "recording new node binding" and "skipped due to exclusive topology" logs), so an unconditional InfoS here will be noticeably louder than the rest of the subsystem on a busy controller. Consider dropping these two success logs to `klog.V(4)` to stay consistent and avoid steady-state log spam. The no-binding log above is an anomaly/failover path, so keeping that one at Info seems reasonable.

pkg/reconciler/roleinstance/sync/node_binding.go
```diff
@@ -387,6 +393,9 @@ func InjectInPlaceScheduling(pod *v1.Pod, instance *workloadsv1alpha2.RoleInstan
 		}
 		inPlaceExprs = append(inPlaceExprs, avoidExprs...)
 		foldIntoRequired(nodeAffinity, inPlaceExprs)
+		klog.InfoS("in-place scheduling: injected required node affinity",
```
cheyang: Same here as the preferred case: a V(4) level would line up with the existing trace logs in this file and keep the normal injection path from flooding controller logs.

pkg/reconciler/roleinstance/sync/node_binding.go
googs1025: Fixed.

pkg/reconciler/roleinstance/sync/node_binding.go
googs1025: Fixed.

cheyang (approved):
All prior suggestions have been addressed in this revision.

- The two success-path injection logs (preferred and required) are now properly gated behind `klog.V(4)`, matching the verbosity of other in-place scheduling traces in this file.
- The early-return "no node binding found" log stays at regular `InfoS` level, which makes sense for surfacing that diagnostic signal to operators.

LGTM — nice clean iteration.

PR #434, opened 2026-08-25
Implement kep-430

diw-zw:
refer to https://github.com/sgl-project/rbg/pull/430


pkg/scheduler/volcano/scheduler.go
```diff
@@ -0,0 +1,322 @@
...
+				},
+			},
+			MatchLabelKeys: []string{constants.RoleInstanceNameLabelKey},
+		})
+	}
+	return policies
+}
+
+// calculateGangMinimum computes the gang minimum as Σ(minReplicas × subGroupSize) for each role.
+func calculateGangMinimum(rbg *workloadsv1alpha2.RoleBasedGroup, gangStrategy *workloadsv1alpha2.GangSchedulingStrategy) int {
```
cheyang: `calculateGangMinimum` only sums roles that already exist in `rbg.Spec.Roles`, and it never checks `minReplicas` against the role's own replica count. Two problems follow from that.

A `minReplicas` key that matches no role (a typo, or a role that was renamed) contributes nothing, so the sum can be 0. `buildSubGroupPolicy` then returns an empty slice and the PodGroup is created with `minMember: 0`, which Volcano happily admits and schedules pod-by-pod with no gang guarantee at all. And `minReplicas[role]` greater than `*role.Replicas` produces a `minMember` larger than the pod count the RBG will ever create, so the gang can never fill and every pod stays Pending.

I reproduced both on a live Volcano 1.14.4 cluster: the typo'd role name gave `spec.minMember: 0`, and the over-count left all pods Pending. Please validate `minReplicas` in the CoordinatedPolicy webhook (keys must name existing roles, values must fall within `[1, role.replicas]`) and make sure a requested gang can never resolve to `minMember: 0`.

pkg/scheduler/volcano/scheduler.go
```diff
@@ -0,0 +1,322 @@
...
+		hasSubGroup := checkPodGroupCRDHasSubGroup(apiReader)
+		m.hasSubGroupPolicy.Store(hasSubGroup)
+	})
+
+	return m.createOrUpdate(ctx, rbg, gangStrategy)
+}
+
+// InjectPodSchedulingFields injects the Volcano PodGroup annotation and schedulerName
+// into the pod template spec.
+func (m *GangScheduler) InjectPodSchedulingFields(
```
cheyang, line 127: The new `role` parameter isn't read here. Whenever `gangStrategy` is non-nil this injects `schedulerName=volcano` and the PodGroup annotation into every role's pod template, including roles that were deliberately left out of `minReplicas`. Those pods then join the PodGroup and Volcano counts them against a `minMember` that never budgeted for them.

On the live cluster the excluded role still carried the PodGroup annotation. Either honor `role` and skip the roles outside the gang, or drop the parameter and document that gang applies to the whole group.

pkg/scheduler/common/gang_strategy.go
```diff
@@ -0,0 +1,55 @@
...
+// annotation compat path: when the rbg has the gang-scheduling annotation set
+// to "true", a default (empty) GangSchedulingStrategy is returned so that
+// callers can treat all gang-enabling paths uniformly via a non-nil pointer.
+//
+// Returns nil when gang scheduling is not enabled.
+func GetGangStrategy(ctx context.Context, c client.Reader, rbg *workloadsv1alpha2.RoleBasedGroup) *workloadsv1alpha2.GangSchedulingStrategy {
+	// 1. Check CoordinatedPolicy for an explicit gang strategy.
+	coordinatedPolicy := &workloadsv1alpha2.CoordinatedPolicy{}
+	if err := c.Get(ctx, types.NamespacedName{Name: rbg.Name, Namespace: rbg.Namespace}, coordinatedPolicy); err == nil {
+		for i := range coordinatedPolicy.Spec.Policies {
```
cheyang: This returns on the first policy rule that carries `Scheduling.Gang` and never looks at `policy.Roles`. So a rule scoped to a subset of roles via `Roles` is applied group-wide (the scoping field has no effect), and a CoordinatedPolicy that declares two gang rules (a natural way to express different per-role minimums, given `Roles` lives on the rule) has everything after the first silently dropped, with no error, event, or status condition.

Both are confirmed live. I'd either honor `Roles` and merge all gang rules, or reject a CoordinatedPolicy with more than one gang rule in the webhook.

pkg/scheduler/common/gang_strategy.go
```diff
@@ -0,0 +1,55 @@
...
+// GetGangStrategy returns the gang scheduling strategy for the given RBG.
+//
+// It first checks the CoordinatedPolicy (same name/namespace as the RBG) for a
+// Scheduling.Gang strategy. If none is found, it falls back to the legacy
+// annotation compat path: when the rbg has the gang-scheduling annotation set
+// to "true", a default (empty) GangSchedulingStrategy is returned so that
+// callers can treat all gang-enabling paths uniformly via a non-nil pointer.
+//
+// Returns nil when gang scheduling is not enabled.
+func GetGangStrategy(ctx context.Context, c client.Reader, rbg *workloadsv1alpha2.RoleBasedGroup) *workloadsv1alpha2.GangSchedulingStrategy {
```
cheyang: This is the only new consumer of the gang config and it builds a PodGroup, nothing more. The CoordinatedPolicy path never turns on the RoleInstance-level gang that the annotation path documents (`rbg.workloads.x-k8s.io/role-instance-gang-scheduling`): the orphan-pod check, and the atomic all-pods-of-an-instance recreation.

That matters more under `subGroupPolicy`, not less, because here a Volcano subGroup is exactly one RoleInstance. Recreating an instance's pods non-atomically is precisely what drops a subGroup below `subGroupSize`, so the feature that most needs the instance-level guarantee is the one running without it.

Separately, the annotation's doc comment says it "is derived automatically from the RBG-level `GangSchedulingAnnotationKey` annotation during RoleInstanceSet reconciliation", but nothing derives it: the two write sites only propagate a value that is already set. Please have this path set the RoleInstance-level flag on the roles it covers, and fix or drop the "derived automatically" comment.

pkg/scheduler/volcano/scheduler.go
```diff
@@ -0,0 +1,322 @@
...
+			Name:         role.Name,
+			SubGroupSize: ptr.To(workloadsv1alpha2.ComputeSubGroupSize(role)),
+			MinSubGroups: ptr.To(minReplicas),
+			LabelSelector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					constants.GroupNameLabelKey: rbg.Name,
+					constants.RoleNameLabelKey:  role.Name,
+				},
+			},
+			MatchLabelKeys: []string{constants.RoleInstanceNameLabelKey},
```
cheyang, line 335: `matchLabelKeys` is hardcoded to `rolebasedgroup.workloads.x-k8s.io/role-instance-name`. That label is written on the RoleInstanceSet path, so on current `main` (where every role is backed by a RoleInstanceSet) this holds, which is why I'm flagging it as minor rather than blocking.

The trap is latent: if a role backing that doesn't emit the per-instance label is reintroduced (the old Deployment/StatefulSet role types), Volcano can't partition the role and `subGroupSize`/`minSubGroups` describe a partition that doesn't exist, leaving the pods Pending. Worth deriving the key from the role's workload pattern, or rejecting `minReplicas` on roles that can't supply it, rather than emitting a constant unconditionally.

pkg/scheduler/common/gang_strategy.go
```diff
@@ -0,0 +1,55 @@
...
+	// 1. Check CoordinatedPolicy for an explicit gang strategy.
+	coordinatedPolicy := &workloadsv1alpha2.CoordinatedPolicy{}
+	if err := c.Get(ctx, types.NamespacedName{Name: rbg.Name, Namespace: rbg.Namespace}, coordinatedPolicy); err == nil {
```
cheyang: Any error from this `Get` is treated as "no policy" because of the `err == nil` guard, so a transient API error or an RBAC denial is indistinguishable from the policy being absent, and gang scheduling quietly degrades to the annotation path (or off). I'd propagate the error so the reconcile retries instead of silently downgrading a scheduling guarantee.

api/workloads/v1alpha2/coordinatedpolicy_types.go
```diff
@@ -58,6 +58,45 @@ type CoordinatedPolicyStrategy struct {
...
+	// the gang with their respective minimums. Roles absent from this
+	// map are excluded from gang constraints and scheduled normally.
+	//
+	// When the gang field is present but minReplicas is empty (nil),
+	// ALL roles participate and minMember equals GetGroupSize()
+	// (basic all-or-nothing gang).
+	//
+	// +optional
+	// +kubebuilder:validation:Type=object
+	MinReplicas map[string]int32 `json:"minReplicas,omitempty"`
```
cheyang, line 131: Zero and negative values pass through unvalidated: zero emits a meaningless `subGroupPolicy` entry (`minSubGroups=0`), and a negative value subtracts from `minMember`. There's no CRD-level minimum marker on the field either. Zero was accepted on the live cluster. A `+kubebuilder:validation:Minimum=1` marker plus the webhook check would close this.

cheyang (changes requested):
Thanks for pushing KEP-430 forward. The two-level gang design is the right direction and the Volcano `subGroupPolicy` wiring is close. I verified the changed behavior against a live Volcano 1.14.4 cluster (per-role `minReplicas` emitted as `subGroupPolicy`, `minMember` as the sum), so the comments below are backed by observed PodGroups rather than a code read alone.

A few things should be sorted before merge; I've left them inline. The common theme is input validation and silent degradation: a typo'd or over-large `minReplicas` can turn gang scheduling off or make it permanently unsatisfiable with no error, `CoordinatedPolicyRule.Roles` and multiple gang rules are silently ignored, and the new path skips the RoleInstance-level gang guarantee that the `subGroupPolicy` feature actually depends on.

Two more points that don't map to a single line:
- Nothing watches `CoordinatedPolicy`, so editing or deleting the gang strategy doesn't reconcile the owning RBG; the PodGroup keeps a stale `minMember`/`subGroupPolicy` until something else touches the RBG. I confirmed this live, and the e2e test even pokes `rbg.Spec` with a comment noting the controller doesn't watch CoordinatedPolicy. A watch mapping CoordinatedPolicy to the same-named RBG would also let that test drop the poke.
- Role-level gang only works on Volcano >= 1.14; on scheduler-plugins the same `minReplicas` config returns a hard runtime error. That's a reasonable first step, but it's worth documenting rather than surfacing as a runtime failure.

Happy to re-verify once these are addressed. The harness and the live steps are captured under `docs/verification/pr434-gang-scheduling/`.

pkg/scheduler/common/gang_strategy.go
cheyang: Any error from `c.Get(ctx, ...)` is silently treated as 'no CoordinatedPolicy exists' due to the `err == nil` guard. A transient API error or RBAC denial becomes indistinguishable from the policy being absent. This can cause the controller to delete an existing PodGroup or fall through to the weaker annotation path. Distinguish NotFound from other errors and requeue on non-NotFound failures.

pkg/scheduler/common/gang_strategy.go
```diff
@@ -0,0 +1,55 @@
...
+		for i := range coordinatedPolicy.Spec.Policies {
+			policy := &coordinatedPolicy.Spec.Policies[i]
+			if policy.Strategy.Scheduling != nil && policy.Strategy.Scheduling.Gang != nil {
```
cheyang: Returns on the first policy rule that carries `Scheduling.Gang` without checking `policy.Roles`. A rule scoped to a subset of roles is applied group-wide. Multiple gang rules in the same policy are silently collapsed. Respect the Roles field when selecting which gang strategy applies.

internal/controller/workloads/rolebasedgroup_controller.go
```diff
@@ -449,10 +450,17 @@ func (r *RoleBasedGroupReconciler) reconcilePodGroup(
...
+	if r.gangScheduler == nil {
 		return nil
 	}
-	return r.podGroupManager.ReconcilePodGroup(ctx, rbg, runtimeController, &watchedWorkload, r.apiReader)
+
+	// Fetch the gang scheduling strategy from the CoordinatedPolicy (if any).
+	// This is shared with PodReconciler via common.GetGangStrategy so that
+	// InjectPodSchedulingFields receives the gangStrategy directly, without
+	// relying on annotations as a compatibility flag.
+	gangStrategy := gangcommon.GetGangStrategy(ctx, r.client, rbg)
```
cheyang: GetGangStrategy reads CoordinatedPolicy here but SetupWithManager does not watch it. Creating, updating, or deleting `scheduling.gang` after the RBG exists won't trigger reconciliation, leaving the PodGroup stale. The e2e test even acknowledges this with a manual Spec poke. Add a Watches() mapping CoordinatedPolicy events to the same-named RBG.

pkg/scheduler/volcano/scheduler.go
```diff
@@ -0,0 +1,322 @@
...
+	}
+
+	// Inject schedulerName into pod spec
+	if pts.Spec == nil {
+		pts.Spec = &coreapplyv1.PodSpecApplyConfiguration{}
+	}
+	pts.Spec.WithSchedulerName(SchedulerName)
+
+	// Inject PodGroup annotation
+	pts.WithAnnotations(map[string]string{AnnotationKey: rbg.Name})
```
cheyang, line 148: When `gangStrategy` is non-nil, the annotation and schedulerName are injected for every role including those excluded from a non-empty `minReplicas` map. Roles documented as 'excluded from gang constraints' still get associated with the PodGroup.

api/workloads/v1alpha2/coordinatedpolicy_types.go
cheyang, line 131: Zero emits a meaningless subGroupPolicy entry (minSubGroups=0), negative values subtract from minMember, and values exceeding role.Replicas create permanently unsatisfiable gangs. There is no CRD-level minimum marker and no webhook validation.

pkg/scheduler/volcano/scheduler.go
```diff
@@ -0,0 +1,322 @@
...
+	}
+
+	// Check if the PodGroup CRD has subGroupPolicy field.
+	// This is done via sync.Once because SetupWithManager may have already
+	// loaded the CRD into watchedWorkload, causing the !loaded block above
+	// to be skipped entirely on every reconcile.
+	m.subGroupOnce.Do(func() {
+		hasSubGroup := checkPodGroupCRDHasSubGroup(apiReader)
+		m.hasSubGroupPolicy.Store(hasSubGroup)
+	})
```
cheyang: Caching with sync.Once makes the capability check permanent for the process lifetime. If the first schema read fails (CRD not yet installed during startup), per-role gangs remain unsupported until pod restart.

pkg/scheduler/volcano/scheduler.go
cheyang: An over-provisioned minReplicas (e.g., 5 for a role with 3 replicas) creates a PodGroup that is permanently unsatisfiable with no error or warning event.

pkg/scheduler/common/gang_strategy.go
cheyang: The API design (CoordinatedPolicyRule.Roles scopes per-rule) suggests per-role resolution, but the implementation returns a single group-wide strategy. Which is intended for KEP-430?

cheyang (changes requested):
## Review Summary

This PR implements KEP-430 (role-based gang scheduling) by refactoring `PodGroupManager` into a `GangScheduler` interface with Volcano subGroupPolicy support. The design direction is correct — the two-level gang model and backward-compatible annotation fallback are sound.

### Blocking

1. **CI broken**: `make generate` not run — two apply-configuration files missing from the commit
2. **Error swallowing in GetGangStrategy**: API errors silently treated as 'no policy', risking PodGroup deletion on transient failures
3. **Roles scoping ignored**: `GetGangStrategy` returns the first gang rule regardless of which roles it's scoped to
4. **No CoordinatedPolicy watch**: editing/deleting the gang strategy doesn't trigger RBG reconciliation

### Non-blocking

- InjectPodSchedulingFields annotates all roles even those excluded from minReplicas
- No minReplicas validation (zero, negative, or over-provisioned values accepted)
- sync.Once capability detection is permanently stale

### Positives

- Clean GangScheduler interface abstraction with proper Volcano/scheduler-plugins separation
- DeepCopy fix prevents a real shared-slice mutation bug (with regression test)
- Thorough e2e suite with dedicated Volcano CI workflow and proper label-filter separation
- Backward-compatible annotation path maintained

Note: These findings largely overlap with the existing maintainer review — confirming the same issues were independently identified.

cheyang:
## Suggestion: fail fast for unsupported workload types in `buildSubGroupPolicy`

`buildSubGroupPolicy` emits `MatchLabelKeys: [role-instance-name]` for every role that has a `minReplicas` entry. However, the `role-instance-name` label is **only written by the RoleInstanceSet reconciler** (`InitComponentPodLabels` → `GetSelectorMatchLabels`). For roles whose workload type is native `LeaderWorkerSet`, `Deployment`, or `StatefulSet`, this label is never present on pods — all pods collapse into a single subGroup, `MinSubGroups × SubGroupSize` can never be satisfied, and the PodGroup is **permanently unschedulable** with no error surfaced to the user.

Since full support for these workload types requires each of their reconcilers to produce the `role-instance-name` label (a non-trivial change that can land in a follow-up PR), the minimal fix for this PR is a guard that **rejects** the unsupported combination at reconcile time instead of silently generating a broken PodGroup.

Suggested patch (compile- and vet-verified):

```diff
diff --git a/pkg/scheduler/volcano/scheduler.go b/pkg/scheduler/volcano/scheduler.go
index 3acebe62..80db5837 100644
--- a/pkg/scheduler/volcano/scheduler.go
+++ b/pkg/scheduler/volcano/scheduler.go
@@ -153,7 +153,11 @@ func (m *GangScheduler) createOrUpdate(
 		}
 		// Calculate minMember as sum of (minReplicas × subGroupSize) for each role
 		minMember = int32(calculateGangMinimum(rbg, gangStrategy))
-		subGroupPolicy = buildSubGroupPolicy(rbg, gangStrategy)
+		var err error
+		subGroupPolicy, err = buildSubGroupPolicy(rbg, gangStrategy)
+		if err != nil {
+			return err
+		}
 	}
 
 	podGroup := &volcanoschedulingv1beta1.PodGroup{
@@ -231,7 +235,7 @@ func (m *GangScheduler) createOrUpdate(
 func buildSubGroupPolicy(
 	rbg *workloadsv1alpha2.RoleBasedGroup,
 	gangStrategy *workloadsv1alpha2.GangSchedulingStrategy,
-) []volcanoschedulingv1beta1.SubGroupPolicySpec {
+) ([]volcanoschedulingv1beta1.SubGroupPolicySpec, error) {
 	policies := make([]volcanoschedulingv1beta1.SubGroupPolicySpec, 0, len(gangStrategy.MinReplicas))
 	for i := range rbg.Spec.Roles {
 		role := &rbg.Spec.Roles[i]
@@ -239,6 +243,17 @@ func buildSubGroupPolicy(
 		if !exists {
 			continue
 		}
+		// SubGroupPolicy partitions pods by role-instance-name label, which is only
+		// written by the RoleInstanceSet reconciler. Other workload types (native LWS,
+		// Deployment, StatefulSet) do not produce this label, so the subGroup would
+		// collapse into one entry and leave pods permanently unschedulable.
+		if wt := role.GetWorkloadType(); wt != constants.RoleInstanceSetWorkloadType {
+			return nil, fmt.Errorf(
+				"role %q uses workload type %q which does not support per-role gang scheduling (subGroupPolicy); "+
+					"only RoleInstanceSet workload type is supported because it writes the %s label required for sub-group partitioning",
+				role.Name, wt, constants.RoleInstanceNameLabelKey,
+			)
+		}
 		policies = append(policies, volcanoschedulingv1beta1.SubGroupPolicySpec{
 			Name:         role.Name,
 			SubGroupSize: ptr.To(workloadsv1alpha2.ComputeSubGroupSize(role)),
@@ -252,7 +267,7 @@ func buildSubGroupPolicy(
 			MatchLabelKeys: []string{constants.RoleInstanceNameLabelKey},
 		})
 	}
-	return policies
+	return policies, nil
 }
```

This way:
- Users get a clear error in the RBG status (via `kubectl describe rbg`) instead of a silent Pending.
- A follow-up PR can lift this guard once native LWS / Deployment / StatefulSet reconcilers learn to write `role-instance-name`.
- No functional change for the supported path (RoleInstanceSet with any pattern — standalone, LWP, or custom).

pkg/scheduler/volcano/scheduler.go
```diff
@@ -0,0 +1,322 @@
...
+	rbg *workloadsv1alpha2.RoleBasedGroup,
+	gangStrategy *workloadsv1alpha2.GangSchedulingStrategy,
+) error {
+	logger := log.FromContext(ctx)
+	queue := rbg.Annotations[constants.GangSchedulingVolcanoQueueKey]
+	priorityClassName := rbg.Annotations[constants.GangSchedulingVolcanoPriorityClassKey]
+	desiredAnnotations := common.InheritPodGroupAnnotations(rbg.Annotations, volcanoschedulingv1beta1.AnnotationPrefix)
+
+	// Calculate minMember
+	minMember := int32(rbg.GetGroupSize())
```
NoobDream2568: It appears that the `roles` field in `CoordinatedPolicy` is ignored during gang scheduling validation and execution. For example:
```
kind: CoordinatedPolicy
metadata:
  name: inference-demo
spec:
  policies:
    - name: gang-scheduling
      roles: ["prefill", "decode"]
      strategy:
        scheduling:
          gang:
            minReplicas:
              prefill: 2
              decode: 1
              any-other-role: 1  # This role is NOT in the roles list
```
The controller currently processes all three, which contradicts the intent of the `roles` filter. Thus, it creates three subGroups.

Similar situation in the following example (assume a RBG with 3 roles):
```
kind: CoordinatedPolicy
metadata:
  name: inference-demo
spec:
  policies:
    - name: gang-scheduling
      roles: ["prefill", "decode"]
      strategy:
        scheduling:
          gang: {}
```
Here, `minMember` is calculated from all three roles instead of only "prefill" and "decode".

It is somehow surprising from an API perspective.

test/e2e/testcase/v1alpha2/gang_scheduling.go
```diff
@@ -0,0 +1,549 @@
...
+				if err := f.Client.Get(f.Ctx, client.ObjectKeyFromObject(cpolicy), cp); err != nil {
+					return err
+				}
+				cp.Spec.Policies[0].Strategy.Scheduling = nil
+				return f.Client.Update(f.Ctx, cp)
+			}, utils.Timeout, utils.Interval).Should(gomega.Succeed())
+
+			// Touch RBG Spec to trigger reconciliation (controller doesn't watch CoordinatedPolicy)
+			updateRbgV2(f, rbg, func(rbg *workloadsv1alpha2.RoleBasedGroup) {
+				rbg.Spec.Roles[0].Replicas = ptr.To(int32(1))
```
NoobDream2568: `Roles[0]` already has `replicas: 1` and this is a no-op.

However, the test case still passed due to pods and RoleInstanceSet are still churning state.

NoobDream2568:
Tested on Volcano v1.15.1 + controller built from this PR (--scheduler-name=volcano): prefill=4 / decode=2 replicas, scaling.maxSkew: "50%", scheduling.gang.minReplicas: {prefill: 4, decode: 2}.

Both `OrderReady` and `OrderScheduled` lead to a deadlock. While this behavior may be expected, there is currently no warning or document to indicate it.

cheyang (commented):
**Round 2 review** (iterative, focused on changes since 7621e01).

All four issues from the R1 review are thoroughly addressed:

| R1 issue | Resolution |
|---|---|
| CI broken / missing `make generate` output | Generated deepcopy, applyconfiguration, CRD YAML, and webhook manifests are committed |
| Error handling — silent gang downgrade on API failure | `GetGangStrategy` now returns `error`; `IncompatibleGangConfigError` separates permanent misconfig from transient failures (fixed requeue vs error backoff) |
| Role scoping — per-role minimums ignored scope | `MergeGangStrategies` drops keys outside the declaring rule's `Roles`; `RoleInGang` gates PodGroup annotation injection; excluded roles still get `schedulerName` |
| Watch gaps — CoordinatedPolicy edits not seen | Controller watches `CoordinatedPolicy` with `GenerationChangedPredicate`; CRD existence checked at startup |

New additions are well-designed: context-carried strategy resolution prevents mid-reconcile drift, admission webhook catches self-contained errors early, the `subGroupProbeTTL` probe is a clean upgrade from `sync.Once`, and the CI split into separate scheduler-plugins/volcano jobs is the right isolation.

**Remaining blocker:** the branch has a merge conflict (no CI can fire). After a rebase this should be ready for a final pass.

deploy/helm/rbgs/values.yaml
```diff
@@ -2,7 +2,7 @@
 
 global:
   # imagePullSecrets is shared by the controller Deployment and the crd-upgrade Job.
-  imagePullSecrets: []
+  imagePullSecrets: []
```
cheyang: `imagePullSecrets: [] ` has trailing whitespace — most linters flag this.

api/workloads/v1alpha2/coordinatedpolicy_types.go
```diff
@@ -58,6 +58,71 @@ type CoordinatedPolicyStrategy struct {
...
+	// +optional
+	Scheduling *SchedulingCoordinationStrategy `json:"scheduling,omitempty"`
+}
+
+// SchedulingCoordinationStrategy defines scheduling coordination for roles.
+// This is a domain that can hold multiple scheduling coordination strategies.
+type SchedulingCoordinationStrategy struct {
+	// Gang defines the gang scheduling coordination for roles.
+	// When present, gang scheduling is enabled for the RoleBasedGroup. Which roles
+	// are constrained depends on minReplicas: an empty map covers every role in the
```
cheyang: This still says an empty `minReplicas` map "covers every role in the group", but after the latest change an all-or-nothing rule only covers the roles its own `spec.policies[].roles` lists — `MergeGangStrategies` returns `GangStrategy{Roles: covered}`, and the doc on `GangSchedulingStrategy.MinReplicas` right below was updated to say exactly that. Worth fixing here too, since this is the field users read first and the two comments now contradict each other.

pkg/scheduler/volcano/scheduler.go
```diff
@@ -0,0 +1,412 @@
...
+	gangStrategy *common.GangStrategy,
+	apiReader client.Reader,
+) error {
+	logger := log.FromContext(ctx)
+	queue := rbg.Annotations[constants.GangSchedulingVolcanoQueueKey]
+	priorityClassName := rbg.Annotations[constants.GangSchedulingVolcanoPriorityClassKey]
+	desiredAnnotations := common.InheritPodGroupAnnotations(rbg.Annotations, volcanoschedulingv1beta1.AnnotationPrefix)
+
+	// Calculate minMember over the roles the gang covers
+	minMember, err := common.GangSize(rbg, gangStrategy)
```
cheyang: On the `minReplicas` path this result is thrown away a few lines down, but the error is not: if every covered role happens to sit at `replicas: 0`, `GangSize` fails with "resolved to minMember 0 ... the roles it covers are all scaled to zero" and we never reach `buildGangSpec`, which would have reported the far more useful "minReplicas for role %q is %d but the role only has %d replicas". Moving the `GangSize` call into an `else` branch keeps the better diagnostic and saves a pass over the roles.

pkg/scheduler/common/gang_strategy.go
```diff
@@ -0,0 +1,300 @@
...
+			if existing, ok := merged[roleName]; !ok || minReplicas > existing {
+				merged[roleName] = minReplicas
+			}
+		}
+	}
+
+	if !found {
+		return nil, nil
+	}
+	if allOrNothing {
```
cheyang: With rule A = `roles: [prefill]`, `gang: {}` and rule B = `roles: [decode]`, `gang.minReplicas: {decode: 2}`, `covered` ends up `{prefill, decode}` and the minimums are dropped, so decode's gang minimum jumps from 2 to its full replica count. Is absorbing B's roles into the all-or-nothing set the intent? It reads as the opposite of the branch just below, which refuses to fall back to all-or-nothing precisely because that "would be a stricter constraint than was asked for". Keeping `MinReplicas` for roles that no all-or-nothing rule named would honour both rules. If the current behaviour is deliberate, one line in the field doc about cross-rule strengthening would save someone a debugging session.

test/e2e/testcase/v1alpha2/gang_scheduling.go
```diff
@@ -0,0 +1,614 @@
...
+				_, err := getPodGroup(f, b.podGroupGVK, rbg.Name, rbg.Namespace)
+				return apierrors.IsNotFound(err)
+			}, utils.Timeout, utils.Interval).Should(gomega.BeTrue(),
+				"PodGroup should be deleted after removing scheduling.gang strategy")
+		})
+
+		// ============================================================
+		// Gang scheduling coexists with rollingUpdate and scaling
+		// ============================================================
+		ginkgo.It("gang scheduling coexists with rollingUpdate and scaling strategies", func() {
```
cheyang, line 340: This spec pairs gang with `strategy.rollingUpdate`, not `strategy.scaling`, so the deadlock `raiseScalingTargetsToGangMinimum` exists to break is only covered by the unit test on the helper itself. The interesting part sits downstream of the helper — that the raised target actually reaches the workload and the group leaves Pending — and that is what a unit test cannot show. Adding `Scaling: {maxSkew: 1}` to this policy (or a sibling spec) alongside the gang strategy would pin it end to end. Not blocking, but it is the one path here where the reasoning is stronger than the evidence.

cheyang (commented):
Round 3. The rebase onto `main` is clean — I checked the rebased commits against the pre-rebase series and the only content that moved is `go.mod`/`go.sum`/`vendor`, plus the `RoleInstanceSpec.DeepCopy` hunk that dropped out because #439 landed it upstream. No leftover `podGroupManager` wiring, and `make generate manifests` on this SHA produces no diff, so the committed generated output is in sync. All 13 checks are green, including the two new gang suites. That clears the only thing blocking this in round 2.

The new work since then is worth calling out: `raiseScalingTargetsToGangMinimum` closes a real deadlock between `maxSkew` pacing and a PodGroup that withholds scheduling, and the `defaultedSchedulerName` fix in `podSpecEqual` explains why the rollout partition never walked down. Both come with tests and a commit message that says why, which made this round much easier to review.

Nothing below blocks merge — one stale API doc comment, an error-message regression on the Volcano path, one merge-semantics question, and an e2e gap on the scaling-plus-gang interaction. From my side this is ready once you have looked at those; the changes-requested state still on the PR is from the earlier rounds and needs a maintainer to clear it.

cheyang (approved):
/lgtm
/approve

PR #435, opened 2026-08-26
fix: reject duplicate roleinstance component names

googs1025:
### Ⅰ. Motivation
Duplicate component names are currently accepted by API admission on the user-facing RoleBasedGroup paths and on direct RoleInstance creation. The controller later detects the duplicate names and reports `FailedScale=True` with `the name of component must be unique`.

Rejecting duplicate component names in the CRD schema returns the error earlier during admission for the objects users submit.

### Ⅱ. Modifications
- Mark v1alpha2 `RoleInstanceSpec.Components` as a Kubernetes map-list keyed by `name`.
- Mark v1alpha2 `CustomComponentsPattern.Components` and v1alpha1 `RoleSpec.Components` as Kubernetes map-lists keyed by `name`.
- Update the field comments to document that component names must be unique.
- Regenerate the RoleInstance, RoleBasedGroup, RoleBasedGroupSet CRDs and bundled kubectl manifest.
- Keep the existing controller-side duplicate-name check as defense in depth.

NONE

- Envtest case for rejecting duplicate names in v1alpha2 `RoleBasedGroup.spec.roles[].customComponentsPattern.components[]`.
- Envtest case for rejecting duplicate names in v1alpha1 `RoleBasedGroup.spec.roles[].components[]`.
- Envtest case for rejecting duplicate names in v1alpha2 `RoleInstance.spec.components[]`.

- `make manifests`
- `make fmt`
- `go test -mod=mod ./api/workloads/v1alpha1 ./api/workloads/v1alpha2 -count=1`
- `go test -mod=mod ./test/envtest/testcase/rbg -run '^$' -count=1`

I also attempted the focused envtest locally, but this machine is missing `/usr/local/kubebuilder/bin/etcd` and `setup-envtest use 1.31.0 --bin-dir ./bin -p path` timed out while downloading `envtest-releases.yaml` from `raw.githubusercontent.com`. The envtest cases are included for CI to run against the repository envtest setup.

This relies on Kubernetes structural schema map-list semantics to reject duplicate component names during admission. The controller validation remains unchanged as a fallback for clusters that have not applied the updated CRD.

For upgrade cases, objects that already contain duplicate component names are already semantically invalid and can still be caught by the controller defense. Applying the new CRD may cause subsequent writes to those invalid objects to fail schema validation on clusters without validation ratcheting behavior, so the duplicate component names should be corrected before or when those objects are updated.

## Checklist

### DIFF
--- a/api/workloads/v1alpha2/roleinstance_types.go
+++ b/api/workloads/v1alpha2/roleinstance_types.go
@@ -23,7 +23,7 @@ import (
 
 // RoleInstanceSpec defines the desired state of RoleInstance
 type RoleInstanceSpec struct {
-	// Components is a list of components, each of which specifies a component and the number of replicas and template for RoleInstance that match the component.
+	// Components is a list of components, each of which specifies a component and the number of replicas and template for RoleInstance that match the component. Component names must be unique.
 	// +listType=map
 	// +listMapKey=name
 	Components []RoleInstanceComponent `json:"components" patchStrategy:"merge" patchMergeKey:"name"`
--- a/api/workloads/v1alpha2/roleinstance_types.go
+++ b/api/workloads/v1alpha2/roleinstance_types.go
@@ -24,6 +24,8 @@ import (
 // RoleInstanceSpec defines the desired state of RoleInstance
 type RoleInstanceSpec struct {
 	// Components is a list of components, each of which specifies a component and the number of replicas and template for RoleInstance that match the component.
+	// +listType=map
+	// +listMapKey=name
 	Components []RoleInstanceComponent `json:"components" patchStrategy:"merge" patchMergeKey:"name"`
 
 	// RoleInstanceReadyPolicy specifies the policy for determining if the RoleInstance is ready.
--- a/config/crd/bases/workloads.x-k8s.io_roleinstances.yaml
+++ b/config/crd/bases/workloads.x-k8s.io_roleinstances.yaml
@@ -93,6 +93,9 @@ spec:
                   - template
                   type: object
                 type: array
+                x-kubernetes-list-map-keys:
+                - name
+                x-kubernetes-list-type: map
               readinessGates:
                 description: ReadinessGates is an optional list of PodReadinessGates
                   for the whole RoleInstance.
--- a/config/crd/bases/workloads.x-k8s.io_roleinstances.yaml
+++ b/config/crd/bases/workloads.x-k8s.io_roleinstances.yaml
@@ -54,7 +54,7 @@ spec:
               components:
                 description: Components is a list of components, each of which specifies
                   a component and the number of replicas and template for RoleInstance
-                  that match the component.
+                  that match the component. Component names must be unique.
                 items:
                   properties:
                     annotations:
--- a/test/envtest/testcase/rbg/basic_test.go
+++ b/test/envtest/testcase/rbg/basic_test.go
@@ -23,10 +23,13 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	workloadsv1alpha1 "sigs.k8s.io/rbgs/api/workloads/v1alpha1"
+	workloadsv1alpha2 "sigs.k8s.io/rbgs/api/workloads/v1alpha2"
 	"sigs.k8s.io/rbgs/test/envtest/testutil"
 )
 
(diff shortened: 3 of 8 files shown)


api/workloads/v1alpha2/roleinstance_types.go
```diff
@@ -24,6 +24,8 @@ import (
 // RoleInstanceSpec defines the desired state of RoleInstance
 type RoleInstanceSpec struct {
 	// Components is a list of components, each of which specifies a component and the number of replicas and template for RoleInstance that match the component.
+	// +listType=map
```
cheyang, line 27: This marker only constrains the RoleInstance object, which in the normal flow is written by the controller rather than by a user. The field people actually author is `spec.roles[].customComponentsPattern.components[]` on the RoleBasedGroup, and `CustomComponentsPattern.Components` (rolebasedgroup_types.go:406) still has no list markers — the generated RBG CRD shows that array with no `x-kubernetes-list-type`. From there `constructRoleInstanceTemplateByCustomComponentsPattern` (pkg/reconciler/roleinstanceset_reconciler.go:287) copies each entry straight into `roleInstanceTemplate.components`, and `RoleInstanceSetSpec.RoleInstanceTemplate` is `+kubebuilder:validation:Schemaless` with PreserveUnknownFields, so nothing validates it at the RoleInstanceSet level either.

So an RBG with two components named `worker` is still admitted. What changes is that the RoleInstanceSet controller's `client.Create` (pkg/reconciler/roleinstanceset/statefulmode/instance_ref_manager.go:54) now gets an Invalid error instead of the RoleInstance coming up with `FailedScale=True`. I haven't traced how far that create error surfaces to the user, but it isn't admission-time feedback on the object they submitted, which is what the description promises.

Could you either put the same `+listType=map` / `+listMapKey=name` on `CustomComponentsPattern.Components` (and v1alpha1 `roles[].components`), or do the uniqueness check in the RBG validator where the other user-facing checks already live (rolebasedgroup_validation.go)? While you're in this file, the field comment still doesn't mention that names must be unique — that text is what ends up in the generated API docs.

api/workloads/v1alpha2/roleinstance_types.go
```diff
@@ -24,6 +24,8 @@ import (
...
+	// +listMapKey=name
```
cheyang, line 28: Nothing tests the one behavior this PR introduces. The envtest harness already starts an API server against `config/crd/bases` (test/envtest/testutil/setup.go:69-74), so a case that creates a RoleInstance with two components sharing a name and expects an Invalid error is a handful of lines, and it pins the markers down — as things stand, losing either one in a later refactor of this struct would go unnoticed. Please add one under test/envtest/testcase.

config/crd/bases/workloads.x-k8s.io_roleinstances.yaml
```diff
@@ -93,6 +93,9 @@ spec:
                   - template
                   type: object
                 type: array
+                x-kubernetes-list-map-keys:
+                - name
+                x-kubernetes-list-type: map
```
cheyang, line 98: Worth a thought for the upgrade case: a cluster that already has RoleInstances with duplicate component names — exactly the situation described in the motivation — gets this schema applied underneath it. Where CRD validation ratcheting isn't in play, subsequent writes to those objects can start failing validation and the operator ends up unable to patch them. I haven't reproduced this, and such objects are already stuck in FailedScale, so it may well be a non-issue; if you've already checked it, a line in the PR description would save the next reader the same question.

cheyang (changes requested):
Two things I'd like resolved before this merges: the constraint sits on the derived RoleInstance rather than on the RoleBasedGroup field users actually write, and there's no test for the new rejection even though the envtest harness already installs these CRDs. Details inline.

Separately, and not your problem: on the previous head the `e2e-test` job failed on the webhook bootstrap assertion (`caBundle not injected into every webhook of rbgs-validating-webhook-configuration`). The object dump printed right after that failure shows a caBundle on both webhooks, so the check was racing the controller's patch rather than reacting to anything in this diff. e2e is re-running on the current head; no need to chase it.

api/workloads/v1alpha2/roleinstance_types.go
googs1025: Added envtest coverage under `test/envtest/testcase/rbg`: v1alpha2 RoleBasedGroup custom components, v1alpha1 RoleBasedGroup components, and direct v1alpha2 RoleInstance components now each assert duplicate component names are rejected with an Invalid error.

config/crd/bases/workloads.x-k8s.io_roleinstances.yaml
googs1025: Added an upgrade note to the PR description. The new CRD can make subsequent writes to existing invalid objects fail schema validation on clusters without validation ratcheting, so duplicate component names should be corrected before or when those objects are updated; the controller check remains as defense in depth.

api/workloads/v1alpha2/roleinstance_types.go
googs1025: Fixed by adding the same map-list markers to v1alpha2 `CustomComponentsPattern.Components` and v1alpha1 `RoleSpec.Components`, keeping the RoleInstance marker as well, and regenerating the CRDs plus bundled manifest. I also updated the field comments so the generated API docs mention that component names must be unique.

api/workloads/v1alpha2/roleinstance_types.go
```diff
@@ -23,7 +23,9 @@ import (
 
 // RoleInstanceSpec defines the desired state of RoleInstance
 type RoleInstanceSpec struct {
-	// Components is a list of components, each of which specifies a component and the number of replicas and template for RoleInstance that match the component.
+	// Components is a list of components, each of which specifies a component and the number of replicas and template for RoleInstance that match the component. Component names must be unique.
+	// +listType=map
+	// +listMapKey=name
 	Components []RoleInstanceComponent `json:"components" patchStrategy:"merge" patchMergeKey:"name"`
```
diw-zw, line 29: roleinstancesets has no admission coverage for this field — low reachability, but the failure mode is worse.

   Two gaps stack here: there's no webhook for the kind (`config/webhook/manifests.yaml` registers only `vrolebasedgroup.kb.io` and `vrolebasedgroupset.kb.io`), and the CRD schema is blind under spec.roleInstanceTemplate because it's schemaless (`roleinstanceset_types.go:46-48 → roleinstancesets.yaml:146-149`), so the new marker doesn't propagate there.

   In practice this is hard to reach: the controller copies component names verbatim from the now-validated RBG field (`pkg/reconciler/roleinstanceset_reconciler.go:287 and :325`), and a RoleInstance derived from a bad template would still be rejected at Create (`pkg/reconciler/roleinstanceset/statelessmode/sync/scale.go:252`). So it takes a hand-written or
   third-party RoleInstanceSet to trigger. But when it does trigger, the error surfaces as an Invalid from that Create — a reconcile error loop rather than a condition on the object. Not blocking; just noting the gap exists if you want to close it later.

test/envtest/testcase/rbg/basic_test.go
```diff
@@ -98,5 +101,109 @@ var _ = Describe("RoleBasedGroup Controller", func() {
...
+							},
+						},
+					},
+				},
+			}
+
+			expectInvalidCreate(rbg)
+		})
+
+		It("Should reject v1alpha1 components with duplicate names", func() {
```
diw-zw, line 138: `Should reject v1alpha1 components with duplicate names `(test/envtest/testcase/rbg/basic_test.go:138) does not test what it claims, and would still be a broken test even if the duplicate name were removed.

   A negative test like this is only meaningful if the object fails to create for the reason under test, and only that reason. This one fails for three independent reasons. The RoleSpec at :145-159 sets only Name,

   Replicas and Components — no Workload, no Template/TemplateRef — so the API server rejects it with:

      [spec.roles[0].workload.apiVersion: Invalid value: "": should match '^[a-z0-9]...',
       spec.roles[0].components[1]: Duplicate value: map[string]interface{}{"name":"worker"},
       spec.roles[0]: Invalid value: "object": template or templateRef must be set for non-InstanceSet workloads]

   The first and third errors have nothing to do with component names. The assertion at :208 is Expect(apierrors.IsInvalid(err)).To(BeTrue()), which only checks that some validation failed — it can't distinguish
   which field caused it. Either of the two unrelated errors is enough to satisfy it.

   The consequence: rename the second component to worker-2 so there is no duplicate at all, and the test still passes, because the object is independently invalid. The same holds if the map-list markers are
   dropped from the CRD entirely — I verified that directly by deleting x-kubernetes-list-type: map from the v1alpha1 components block in config/crd/bases/workloads.x-k8s.io_rolebasedgroups.yaml and re-running the
   suite: still ok. So this spec would keep passing if the feature it guards were reverted.

   Two changes are needed, and only fixing one of them isn't enough:

   1. :208 — assert on the field path, not the error category, e.g. Expect(err.Error()).To(ContainSubstring("spec.roles[0].components[1]: Duplicate value")). Since expectInvalidCreate is shared by all three specs,
   it's probably cleanest to take the expected substring as a parameter.
   2. :145-159 — make the object otherwise valid (set Workload and Template/TemplateRef) so that the duplicate name is the only thing wrong with it. Otherwise you're still asserting validation behaviour on an
   object that could never be created anyway.

   For contrast, the other two specs (:105 v1alpha2 and :167 RoleInstance) build valid objects and their error lists contain only the single Duplicate value entry, so they do exercise the intended behaviour —
   though they'd still benefit from the tighter assertion in (1) to stay meaningful under future changes.

cheyang (approved):
/lgtm
/approve

PR #436, opened 2026-08-28
fix: align template ref and update strategy validation

googs1025:
### I. Motivation
Two validation paths were inconsistent with the v1alpha2 API shape:

- `templateRef.patch` is optional in the API type and template resolution already supports nil or empty patches, but controller validation rejected `templateRef` without a patch.
- v1alpha2 update strategy type fields accepted arbitrary strings because their CRD schemas had no enum constraint.

### II. Modifications
- Allow `templateRef` without `patch` in controller-side role template reference validation.
- Add kubebuilder enum validation for:
  - `spec.roles[].rolloutStrategy.rollingUpdate.type`
  - `RoleInstanceSet.spec.updateStrategy.type`
- Regenerate CRDs and the bundled kubectl manifest.
- Update controller validation tests and add envtest coverage for API acceptance/rejection behavior.

### III. Does this pull request fix one issue?
NONE

### IV. List the added test cases (unit test/integration test) if any, please explain if no tests are needed.
- Updated `TestRoleInstanceSetReconciler_ValidateRoleTemplateReferences` to expect `templateRef` without `patch` to pass.
- Added envtest coverage that v1alpha2 RBG with `roleTemplates` and `templateRef` but no `patch` becomes Ready.
- Added envtest coverage that invalid RBG rolling update type is rejected by API validation.
- Added envtest coverage that invalid RIS update strategy type is rejected by API validation.

### V. Describe how to verify it
- `make fmt`
- `git diff --check`
- `go test ./api/workloads/v1alpha2 ./pkg/reconciler -count=1`
- `go test ./test/envtest/testcase/rbg -run '^$' -count=1`
- Attempted `make test-envtest`, but the local run could not fetch envtest assets from GitHub and then failed before running specs because `kube-apiserver` was unavailable in PATH.

This intentionally only uses CRD enum validation for update strategy type rejection, without adding extra webhook/controller defense-in-depth validation.

## Checklist

### DIFF
--- a/api/workloads/v1alpha2/rolebasedgroup_types.go
+++ b/api/workloads/v1alpha2/rolebasedgroup_types.go
@@ -118,6 +118,7 @@ const (
 type RollingUpdate struct {
 	// Type indicates the type of the InstanceSetUpdateStrategy.
 	// Default is InPlaceIfPossible.
+	// +kubebuilder:validation:Enum={RecreatePod,InPlaceIfPossible,InPlaceOnly}
 	Type UpdateStrategyType `json:"type,omitempty"`
 
 	// Partition indicates the ordinal at which the role should be partitioned for updates.
--- a/api/workloads/v1alpha2/roleinstanceset_types.go
+++ b/api/workloads/v1alpha2/roleinstanceset_types.go
@@ -119,6 +119,7 @@ type RoleInstanceSetScaleStrategy struct {
 type RoleInstanceSetUpdateStrategy struct {
 	// Type indicates the type of the RoleInstanceSetUpdateStrategy.
 	// Default is InPlaceIfPossible.
+	// +kubebuilder:validation:Enum={RecreatePod,InPlaceIfPossible,InPlaceOnly}
 	Type UpdateStrategyType `json:"type,omitempty"`
 
 	// Partition is the desired number of RoleInstances in old revisions.
--- a/api/workloads/v1alpha2/roletemplate_validation.go
+++ b/api/workloads/v1alpha2/roletemplate_validation.go
@@ -135,13 +135,6 @@ func validateRoleTemplateFields(
 			)
 		}
 
-		// RawExtension cannot be inspected by CEL, so validate in controller.
-		if !hasTemplatePatch {
-			return fmt.Errorf(
-				"spec.roles[%d]: templateRef.patch is required when templateRef is set (if no overrides needed, set patch to empty object: {})",
-				index,
-			)
-		}
 		return nil
 	}
 
--- a/test/envtest/testcase/rbg/basic_test.go
+++ b/test/envtest/testcase/rbg/basic_test.go
@@ -24,6 +24,7 @@ import (
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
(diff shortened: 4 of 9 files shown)


api/workloads/v1alpha2/rolebasedgroup_types.go
```diff
@@ -118,6 +118,7 @@ const (
 type RollingUpdate struct {
 	// Type indicates the type of the InstanceSetUpdateStrategy.
 	// Default is InPlaceIfPossible.
+	// +kubebuilder:validation:Enum={RecreatePod,InPlaceIfPossible,InPlaceOnly}
```
diw-zw, line 121: Upgrade hazard: this enum tightens the schema for a field that previously accepted any string. `rollingUpdate.type` has no CRD default and no defaulting webhook — the "Default is InPlaceIfPossible" behavior only exists in-memory when the RBG controller generates child RISes (pkg/reconciler/roleinstanceset_reconciler.go:229) — so stored objects can legitimately contain `type: ""` (templated/Helm manifests) or arbitrary strings (v1alpha1 has no enum either). After this CRD ships:

- status updates keep working (the status subresource isn't validated against spec), so such RBGs keep looking Ready;
- but every spec write is rejected with 422 — including controller-driven ones: RBGScalingAdapter.updateRoleReplicas (internal/controller/workloads/rolebasedgroupscalingadapter_controller.go:511) and the RBGS controller's update (internal/controller/workloads/rolebasedgroupset_controller.go:465) — so scaling silently breaks for those RBGs.

Please consider: (a) a release note + migration guidance (patch the field to a valid value or remove it), and/or (b) a defaulting webhook normalizing `""` -> `InPlaceIfPossible` before validation (matching the documented default). Note a CEL transition rule can't grandfather this while keeping the enum: rules referencing oldSelf are skipped on CREATE, and the enum rejects `""` unconditionally.

api/workloads/v1alpha2/roleinstanceset_types.go
```diff
@@ -119,6 +119,7 @@ type RoleInstanceSetScaleStrategy struct {
 type RoleInstanceSetUpdateStrategy struct {
 	// Type indicates the type of the RoleInstanceSetUpdateStrategy.
 	// Default is InPlaceIfPossible.
+	// +kubebuilder:validation:Enum={RecreatePod,InPlaceIfPossible,InPlaceOnly}
```
diw-zw, line 122: Same upgrade concern as the RBG enum: directly-created RoleInstanceSets can carry stored `updateStrategy.type` values this enum now rejects (e.g. `""`), and they'll fail with 422 on their next spec update while status keeps succeeding. RISes generated by the RBG controller are safe, because the controller defaults `""` -> InPlaceIfPossible before creating them. Suggest the same migration note / defaulting consideration here.

api/workloads/v1alpha2/roletemplate_validation.go
```diff
@@ -135,13 +135,6 @@ func validateRoleTemplateFields(
...
 		}
 
-		// RawExtension cannot be inspected by CEL, so validate in controller.
-		if !hasTemplatePatch {
-			return fmt.Errorf(
-				"spec.roles[%d]: templateRef.patch is required when templateRef is set (if no overrides needed, set patch to empty object: {})",
-				index,
-			)
-		}
 		return nil
```
diw-zw, line 138: This removal reverses the KEP-8 validation table, which says `templateRef Set -> templatePatch Required` (keps/8-reduce-yaml-duplication/README.md:102-107). Please update the KEP accordingly (this is also the unchecked "Update the documentation" checklist item) rather than silently dropping the requirement.

Also worth stating in the PR description — it's the strongest argument for this change: the removed check rejects objects that are perfectly valid in v1alpha1. A v1alpha1 RBG with `templatePatch: {}` (which v1alpha1's own validation error message tells users to write) converts to a hub object with `Patch=nil`, because convertTemplateRefV1alpha1ToV2 drops empty patches (api/workloads/v1alpha1/rolebasedgroup_conversion.go:229). The v1alpha2 controller then rejected it with "patch is required". So the check was actively breaking upgraded objects, not just being strict.

Related: v1alpha1 still requires the patch, so this creates a semantic divergence between versions — worth an explicit note that v1alpha1 remains frozen with the old semantics.

diw-zw (commented):
Overall the change is well-tested and the direction (aligning validation with the implemented + documented optional-patch API) is right — I verified unit tests, envtest, and `make manifests` reproducibility locally.

Two things to add before merging:

1. **Upgrade-path coverage.** All new tests exercise fresh creations only. Given the enum tightening (see inline comments), upgrade coverage would be valuable: create an object under the previous CRD schema with `type: ""` or an invalid type string, upgrade to the new CRD, and assert the intended behavior (still updatable, or a documented failure mode). The release-gate e2e suite would be the natural home.

2. **Release note / migration guidance** for the enum tightening, unless the defaulting-webhook approach suggested in the inline comments is adopted instead.

api/workloads/v1alpha2/rolebasedgroup_types.go
cheyang, line 121: Tightening this field to an enum only covers the v1alpha2 write path. v1alpha1 is still served (`served: true, storage: false`) and its copy of `rollingUpdate.type` has no enum, and CRD validation runs against the schema of the version in the request — the object converted to the storage version is not re-validated. So a v1alpha1 create can still get an out-of-enum value persisted into the v1alpha2 hub form.

That is not hypothetical here, because the two versions disagree on the value names. v1alpha1 documents `Recreate` (`api/workloads/v1alpha1/constant.go:232`), v1alpha2 renamed it to `RecreatePod`, and `convertRolloutStrategyV1alpha1ToV2` casts the string verbatim with no mapping (`api/workloads/v1alpha1/rolebasedgroup_conversion.go:245`). A v1alpha1 RBG with `rollingUpdate.type: Recreate` therefore ends up stored as `Recreate` under v1alpha2. From that point:

- every v1alpha2 spec write on that RBG returns 422, including the controller-driven ones in `rolebasedgroupscalingadapter_controller.go:511` and `rolebasedgroupset_controller.go:465`, so scaling quietly stops working;
- if the role resolves to `workloads.x-k8s.io/v1alpha2/RoleInstanceSet`, the child RIS is built with `updateStrategy.type: Recreate` (`pkg/reconciler/roleinstanceset_reconciler.go:228` only defaults the empty string), and the new RIS enum rejects that create outright — the role never rolls out at all.

So as written the enum both fails to close the hole for v1alpha1 clients and breaks a configuration they can legitimately create today. I'd rather see the value mapped in `convertRolloutStrategyV1alpha1ToV2` (and normalised in the RIS reconciler next to the existing empty-string default) than rely on the enum alone; widening the enum to accept the v1alpha1 alias would also work if you want to keep it schema-only.

While you're in this struct: `+kubebuilder:default=InPlaceIfPossible` is worth adding too. `partition`, `maxUnavailable`, `maxSurge` and `rolloutStrategy.type` all carry CRD defaults, so `type` is the odd one out, and the default the doc comment promises currently only exists in controller memory.

api/workloads/v1alpha2/roleinstanceset_types.go
cheyang, line 122: One correction to the existing comment on this line: generated RISes are safe for `""`, which `pkg/reconciler/roleinstanceset_reconciler.go:228-230` defaults to `InPlaceIfPossible` before the apply, but they are not safe for a v1alpha1-originated `Recreate`. That value is passed straight into `WithType(...)`, so with this enum in place the RIS create itself starts failing rather than just a later RBG spec update. Whatever normalisation you add is probably best placed right next to that existing empty-string default so both cases are handled in one spot.

test/envtest/testcase/rbg/basic_test.go
```diff
@@ -186,6 +187,111 @@ var _ = Describe("RoleBasedGroup Controller", func() {
...
+					return err
+				}
+				if !apimeta.IsStatusConditionTrue(created.Status.Conditions, string(workloadsv1alpha2.RoleBasedGroupReady)) {
+					return fmt.Errorf("RoleBasedGroup %s/%s Ready condition is not true: %v", testNs, rbgName, created.Status.Conditions)
+				}
+				return nil
+			}, timeout, interval).Should(Succeed())
+		})
+
+		It("Should reject v1alpha2 role rolloutStrategy with invalid update type", func() {
```
cheyang, line 236: Both new rejection cases use `DefinitelyNotARealStrategy`, which confirms the enum is wired into the CRD but not much beyond that. The values that decide whether this is safe to ship are `""` (what templated/Helm manifests tend to emit for an `omitempty` string) and `Recreate` (valid under v1alpha1). Could you add those two, plus one case asserting each of the three accepted values actually goes through?

An update-path case would help more than another create case, since the create path is not what breaks for existing objects: create with a valid type, then patch an unrelated spec field, and assert the write succeeds.

cheyang (dismissed):
The templateRef half of this checks out — `applyStrategicMergePatch` already treats a nil or empty patch as a no-op (`api/workloads/v1alpha2/helper.go:440-446`), so the check you removed was rejecting objects the resolver handles fine, and the v1alpha1 `templatePatch: {}` -> `Patch=nil` conversion made it actively harmful rather than merely strict.

My concern is with the enum. It only lands on the v1alpha2 schema, while v1alpha1 is still served with no enum and the v1alpha1 -> v1alpha2 conversion casts the strategy string without mapping `Recreate` to `RecreatePod`. That makes the new validation both bypassable and breaking for v1alpha1 users, and in the RoleInstanceSet case it blocks the child object from being created at all. Details inline.

diw-zw (approved):

PR #438, opened 2026-08-28
Fix stateful RoleInstance identity handling

cyf-2002:
### Ⅰ. Motivation

Stateful RoleInstances currently shallow-copy the nested `RoleInstanceSpec`. Their component slices and Pod template metadata maps can therefore be shared.

When identity labels are injected for a later RoleInstance, labels on an earlier instance may be overwritten.

For example, the following Pod belongs to RoleInstance `prefill-0`, but contains identity labels from `prefill-1`:

```yaml
metadata:
  name: multi-role-custom-components-gang-prefill-0-worker-0
  labels:
    rbg.workloads.x-k8s.io/role-instance-name: multi-role-custom-components-gang-prefill-0
    rbg.workloads.x-k8s.io/role-instance-index: "1"
    statefulset.kubernetes.io/pod-name: multi-role-custom-components-gang-prefill-1
```

The expected RoleInstance index is `0`.

The shallow copy occurs when constructing a RoleInstance:

```go
Spec: setToUse.Spec.RoleInstanceTemplate.RoleInstanceSpec
```

The first commit fixes this by deep-copying the spec before injecting instance-specific metadata.

### Ⅱ. Modifications

This PR contains two separate commits.

#### 1. Fix shared RoleInstance specs

Deep-copy each `RoleInstanceSpec` before modifying component templates:

```go
Spec: *setToUse.Spec.RoleInstanceTemplate.RoleInstanceSpec.DeepCopy()
```

This prevents one RoleInstance from modifying another RoleInstance or the source RoleInstanceSet template.

#### 2. Discuss the Pod name label semantics

Component Pod templates currently inherit:

```text
statefulset.kubernetes.io/pod-name=<RoleInstance name>
```

For example:

```text
Pod name:    multi-role-custom-components-gang-prefill-0-worker-0
Label value: multi-role-custom-components-gang-prefill-0
```

According to the Kubernetes
[StatefulSet Pod name label semantics](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-name-label),
this label identifies the actual Pod name.

The second commit therefore proposes:

- Keep `statefulset.kubernetes.io/pod-name` on the RoleInstance.
- Stop propagating it to component Pod templates.
- Continue propagating `rbg.workloads.x-k8s.io/role-instance-index`.

NONE

### Ⅳ. Added test cases

- `TestInstanceSpecIsolated`
  - Verifies that RoleInstances have independent component templates.
  - Verifies that the source RoleInstanceSet template is not mutated.

- `TestPodLabelsOmitRINSName`
  - Verifies that the RoleInstance keeps its identity label.
  - Verifies that component Pod templates do not inherit the
    StatefulSet Pod name label.

### Ⅴ. Verification

```shell
go test ./pkg/reconciler/roleinstanceset/statefulmode -count=1
```

### VI. Discussion

The DeepCopy change is a correctness fix.

The second commit changes existing component Pod labels and needs maintainer agreement. Which behavior should RBG use?

1. Remove `statefulset.kubernetes.io/pod-name` from component Pods.
2. Set it to the actual component Pod name during Pod creation.
3. Keep the current RoleInstance-name value for compatibility.

The commits are separated so the second change can be revised or dropped without affecting the DeepCopy fix.

### DIFF
--- a/pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control_test.go
+++ b/pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control_test.go
@@ -33,9 +33,31 @@ import (
 	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/ptr"
 
+	"sigs.k8s.io/rbgs/api/workloads/constants"
 	workloadsv1alpha2 "sigs.k8s.io/rbgs/api/workloads/v1alpha2"
 	instanceinplace "sigs.k8s.io/rbgs/pkg/inplace/instance/inplaceupdate"
 )
+
+func TestInstanceSpecIsolated(t *testing.T) {
+	set := buildSet("trainer", 2, nil, nil)
+	set.Spec.RoleInstanceTemplate.RoleInstanceSpec.Components = []workloadsv1alpha2.RoleInstanceComponent{
+		{Name: "worker"},
+	}
+
+	first := newVersionedInstance(set, set, testUpdateRev, testUpdateRev, 0, nil)
+	second := newVersionedInstance(set, set, testUpdateRev, testUpdateRev, 1, nil)
+	indexKey := constants.RoleInstanceIndexLabelKey
+
+	if got := first.Spec.Components[0].Template.Labels[indexKey]; got != "0" {
+		t.Fatalf("first component index = %q, want 0", got)
+	}
+	if got := second.Spec.Components[0].Template.Labels[indexKey]; got != "1" {
+		t.Fatalf("second component index = %q, want 1", got)
+	}
+	if _, ok := set.Spec.RoleInstanceTemplate.RoleInstanceSpec.Components[0].Template.Labels[indexKey]; ok {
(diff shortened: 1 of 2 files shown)


diw-zw:
Good catch! We actually spotted this issue and fixed it during the release. 
Regarding the StatefulSet label issue, feel free to open an issue so we can discuss it together.

cyf-2002:
> Good catch! We actually spotted this issue and fixed it during the release.说得好！我们实际上发现了这个问题并在发布时修复了它。 Regarding the StatefulSet label issue, feel free to open an issue so we can discuss it together.关于 StatefulSet 标签的问题，请随时打开一个问题报告，我们可以一起讨论这个细节。


The deep-copy change has already fixed the original issue. The only remaining question is whether this label still needs to be preserved:

```text
statefulset.kubernetes.io/pod-name
```
Since RBG already provides a dedicated label:
```text
rbg.workloads.x-k8s.io/role-instance-name
```

I think statefulset.kubernetes.io/pod-name may no longer be necessary.

PR #446, opened 2026-08-30
fix(e2e): gate the upgrade baseline on pods, not the Ready condition

diw-zw:
## Summary
   `Release Test` on main fails intermittently in the upgrade suite's phase 3:

    ```
    pods are no longer in the phase they were in before the upgrade:
      - up-v1a1/lw: pod up-v1a1-lw-0-1 went from Pending to Running
    ```

    The upgrade did not do this. The baseline did: `before` was captured while that pod was
    still `Pending`, and it started about a minute later. This waits on the pods themselves
    before taking the baseline.

## Root cause

    Phase 1 gated the baseline on status only, and an RBG's `Ready` condition inherits
    whatever the workload underneath it calls ready. The chain, bottom up:

    1. LWS v0.7.0 counts a group ready when the leader pod is Running and Ready **and** the
       worker StatefulSet is "ready" (`pkg/controllers/leaderworkerset_controller.go:418`).
    2. That `StatefulsetReady` compares `*sts.Spec.Replicas` against `sts.Status.Replicas` —
       pods *created* — not against `Status.ReadyReplicas`
       (`pkg/utils/statefulset/statefulset_utils.go:48-51`). A `Pending` worker pod counts.
    3. RBG passes that straight up: the role's `Replicas`/`ReadyReplicas` are
       `lws.Status.*` verbatim (`pkg/reconciler/lws_reconciler.go:133-145`).
    4. The `Ready` condition requires `*role.Replicas == rs.Replicas == rs.ReadyReplicas`
       (`internal/controller/workloads/rolebasedgroup_controller.go:736-748`) → `1 == 1 == 1`
       → `Ready=True`.

    CI installs LWS v0.7.0 (`release-test.yml:23`), the version read above. So a `Pending`
    worker beside a `Running` leader makes the whole RBG report Ready, and RBG has no way to
    tell the difference.

    [The failing run](https://github.com/sgl-project/rbg/actions/runs/33303713013) shows
    exactly that: `before` has `up-v1a1` at `ready=true`, `role lw ready=1`, and its worker at
    `phase=Pending rv=1705`. The live dump 80s later has it `Running rv=2290`, with the
    `resourceVersion` unmoved in between — the pod was starting for the first time, not being
    restarted by the upgrade.

    The only pod-level guard was "every role has at least one `Running` pod", asserted on
    `before`. A `Running` leader beside a `Pending` worker satisfies it.

## What changed

    `waitAllPodsRunning` runs after `waitAllRBGsReady`, before `before` is captured, and waits
    until every role has its full expected pod count and every one of those pods is `Running`.

    - Pods come from `listPodFactsForRole`, the same helper `captureRBG` uses, so the gate
      waits on exactly the pods the snapshot will record.
    - `expectedPodsForRole` computes `replicas × pods-per-replica`: `1` for
      `standalonePattern`, `leaderWorkerPattern.size` for leader-worker, `Σ components[].size`
      for `customComponentsPattern`. A pattern it cannot count is an error, not a guess of one
      pod per replica — a guess would silently under-count a pattern added later.
    - Terminating objects are skipped, as in `waitAllRBGsReady`: the prune probe of the first
      phase 1 spec can still be present and will never have pods again.

    Scope, stated honestly: the reproduced failure is the `Pending` pod, and that is what the
    phase check catches. **No workload has been observed opening the Ready gate with pods
    missing outright** — the condition also requires `*role.Replicas == status.Replicas`. The
    count is there because it does not depend on any workload's definition of ready, so it
    holds whatever a later version decides that word means.

    No detector was touched. `checkStillReady` still treats any phase change as a finding,
    including `Pending -> Running`, so an upgrade that really does restart or add a pod is
    still reported.

### DIFF
--- a/test/e2e/upgrade/specs.go
+++ b/test/e2e/upgrade/specs.go
@@ -923,23 +923,28 @@ func waitAllRBGsReady(f *framework.Framework, except ...string) {
 }
 
 // waitAllPodsRunning waits until every role of every RoleBasedGroup in the namespace has
-// its full complement of pods and every one of them is Running, except for the named
-// RBGs.
+// its full expected pod count and every one of those pods is Running, except for the
+// named RBGs.
 //
-// The Ready gate above cannot establish either half. An RBG's Ready condition is only as
-// good as the workload status it is derived from, and for a LeaderWorkerSet role that is
-// LWS's own group accounting: LeaderWorkerSetReconciler.CheckWorkloadReady is
-// `ReadyReplicas == Replicas`, which is also true before LWS has populated its status at
-// all, and both the role status and the condition have been observed reporting the single
-// group ready while one of its worker pods was still Pending.
+// The Ready gate above cannot establish this, because an RBG's Ready condition inherits
+// whatever the workload underneath it calls ready. For a LeaderWorkerSet role that is
+// LWS's group accounting, and LWS v0.7.0 counts a group ready when the leader pod is
+// Running and Ready and the worker StatefulSet merely has its pods *created*: its
+// StatefulsetReady compares Spec.Replicas against Status.Replicas, not against
+// Status.ReadyReplicas. A Pending worker beside a Running leader therefore makes the
+// whole RBG report Ready, which is how up-v1a1-lw-0-1 reached the baseline still Pending
+// and started a minute later.
 //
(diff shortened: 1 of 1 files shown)


cheyang (approved):
/lgtm
/approve

cheyang (approved):
/lgtm
/approve

PR #447, opened 2026-08-31
fix: validate RoleBasedGroup role dependencies at admission

googs1025:
### Ⅰ. Motivation
Reject invalid RoleBasedGroup role dependency graphs during admission instead of accepting the object and surfacing dependency errors later during reconciliation.

RoleBasedGroupSet templates are validated as well because the set controller copies template roles into derived RoleBasedGroups; admitting an invalid template would otherwise move the failure from the original object to child reconciliation.

### Ⅱ. Modifications
- Add `ValidateRoleDependencies(rbg *RoleBasedGroup) error` for unknown dependency roles, self-dependencies, and dependency cycles.
- Reuse the dependency validation for RoleBasedGroupSet templates with the correct `spec.groupTemplate.spec.roles` field path.
- Call role dependency validation from RoleBasedGroup and RoleBasedGroupSet create/update admission paths.
- Add unit coverage for unknown role dependencies, self-dependencies, cyclic dependencies, RBG admission wiring, and RBGSet admission wiring.

NONE

- `TestValidateRoleDependencies`
- `TestRoleBasedGroupValidator_ValidateCreateRoleDependencies`
- `TestRoleBasedGroupValidator_ValidateUpdateRoleDependencies`
- `TestRoleBasedGroupSetValidator_ValidateCreateRoleDependencies`
- `TestRoleBasedGroupSetValidator_ValidateUpdateRoleDependencies`

```bash
go test -count=1 ./api/workloads/v1alpha2
```

This keeps the validation in the API package and does not make admission depend on the controller dependency manager.

`ValidateUpdate` validates the complete role dependency graph on every main-resource update. That means pre-existing invalid objects admitted before this webhook will need their dependency spec fixed before non-status updates can be accepted; status subresource updates are unaffected.

## Checklist

### DIFF
--- a/api/workloads/v1alpha2/rolebasedgroup_admission.go
+++ b/api/workloads/v1alpha2/rolebasedgroup_admission.go
@@ -57,6 +57,9 @@ func (v *RoleBasedGroupValidator) ValidateCreate(_ context.Context, obj runtime.
 	if err := ValidateRollingUpdate(rbg); err != nil {
 		allErrs = append(allErrs, err)
 	}
+	if err := ValidateRoleDependencies(rbg); err != nil {
+		allErrs = append(allErrs, err)
+	}
 	if !v.EnableDeprecatedWorkloadTypes {
 		if err := validateNoDeprecatedWorkloadTypes("spec.roles", rbg.Spec.Roles); err != nil {
 			allErrs = append(allErrs, err)
@@ -80,6 +83,9 @@ func (v *RoleBasedGroupValidator) ValidateUpdate(ctx context.Context, oldObj, ne
 
 	var allErrs []error
 	if err := ValidateRollingUpdate(rbg); err != nil {
+		allErrs = append(allErrs, err)
+	}
+	if err := ValidateRoleDependencies(rbg); err != nil {
 		allErrs = append(allErrs, err)
 	}
 	if err := ValidateScalingAdapterReplicas(ctx, v.Client, oldRBG, rbg); err != nil {
--- a/api/workloads/v1alpha2/rolebasedgroup_validation.go
+++ b/api/workloads/v1alpha2/rolebasedgroup_validation.go
@@ -19,6 +19,7 @@ package v1alpha2
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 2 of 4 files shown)


diw-zw (commented):
## Review

The motivation is sound: invalid role dependency graphs (unknown roles, self-dependencies, cycles) are currently accepted by admission and only surface later as `InvalidRoleDependency` events during reconciliation (`pkg/dependency/dependency.go` `SortRoles`, surfaced at `internal/controller/workloads/rolebasedgroup_controller.go:486`). Rejecting them at admission is the right call.

I verified the DFS cycle detection against the controller's `dependencyOrder` semantics — both reject unknown deps, self-deps, and multi-node cycles. The `path` slice aliasing in `visit` is safe, and `utilerrors.NewAggregate(nil)` returns nil so empty role lists pass. Tests pass locally and CI is green.

### Main findings

**1. RBGSet templates bypass the new validation (back door).**

`RoleBasedGroupSetValidator` (`api/workloads/v1alpha2/rolebasedgroupset_admission.go`) only calls `validateNoDeprecatedWorkloadTypes`, not `ValidateRoleDependencies`. Since `spec.groupTemplate.spec` is a full `RoleBasedGroupSpec` (including `dependencies`), a set with an invalid dependency graph is admitted, and `newRBGForSet` (`internal/controller/workloads/rolebasedgroupset_controller.go:538`) copies the roles verbatim into the derived RBG — whose Create/Update is then rejected by this webhook (failurePolicy=fail). The set controller ends up looping on a reconcile error that only appears in logs, with no event or condition. This is exactly the "surfaced later during reconciliation" problem this PR aims to eliminate, just through a different path.

Suggestion: give `ValidateRoleDependencies` a `fieldPath` parameter (mirroring `validateNoDeprecatedWorkloadTypes`) and wire it into the RBGSet validator for `spec.groupTemplate.spec.roles`.

**2. Upgrade consideration: pre-existing invalid RBGs can no longer be updated at all.**

`ValidateUpdate` runs the check unconditionally on every update. Objects with invalid deps that were admitted before this webhook existed (or on clusters with `--enable-webhooks=none`) will now have every main-resource write rejected — including controller-driven writes such as the discovery-mode annotation patch in `ensureDiscoveryConfigMode` (`internal/controller/workloads/rolebasedgroup_controller.go:380`, which runs before `SortRoles`), causing an API-error loop on each requeue for legacy objects that lack the annotation. Status subresource writes are unaffected.

This is probably acceptable (those objects are already broken), but it's worth documenting in the PR description, and consider validating only when the `spec.roles` dependencies actually changed on update.

### Minor / optional

- Test gaps: multi-error aggregation, cycles of length 3+, cycle rendering when the cycle starts mid-path (`appendCycle`'s main branch), combined unknown-dep + cycle, and `ValidateUpdate` with unknown/self deps.
- Dead code: `appendCycle`'s fallback branch is unreachable (`visiting` implies the role is on the current path), and the `graph[role.Name] = nil` seeding is redundant.
- The doc comment on `ValidateRoleDependencies` doesn't mention the self-dependency check; the 3-color DFS and `appendCycle` truncation are non-obvious and could use a why-comment, matching the style of `validateNoDeprecatedWorkloadTypes`.
- The cycle error message has no field path/index and only the first cycle is reported — UX nit.

None of these are blocking, but #1 is worth addressing before merge since it leaves a path where invalid graphs still fail only at reconcile time.

googs1025:
Thanks for the review. I updated the admission validation to cover RoleBasedGroupSet templates as well, so both RoleBasedGroup and RoleBasedGroupSet create/update paths now validate role dependencies consistently.

diw-zw (approved):

PR #451, opened 2026-09-02
feat(helm): record installed chart version

diw-zw:
### Ⅰ. Motivation

Record the installed Helm chart version so that subsequent upgrades can identify the version currently deployed in the cluster.

### Ⅱ. Modifications

- Add a post-install/post-upgrade ConfigMap named `<release-name>-chart-version`.
- Persist `.Chart.Version` in `data.chartVersion`.
- Retain the marker between upgrades and remove the previous marker before creating the next one.
- Document the chart version marker in the Helm chart README.


deploy/helm/rbgs/templates/chart-version.yaml
```diff
@@ -0,0 +1,16 @@
...
+metadata:
+  name: {{ $markerName }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+  annotations:
+    # A fixed-name hook must remove the previous marker before it can write the new one.
+    # It is otherwise retained as the source-version record for the next upgrade.
+    "helm.sh/hook": post-install,post-upgrade
+    "helm.sh/hook-delete-policy": before-hook-creation
```
cheyang: Helm 3 doesn't track hook resources as part of the release manifest, so this ConfigMap likely survives `helm uninstall`. If the marker is meant to say "what's currently deployed", leaving it behind after uninstall is misleading — the next chart installed under the same release name will find a stale marker until the post-hook overwrites it. Either add a `pre-delete` cleanup hook or document the retention in the README so operators know the ConfigMap outlives the release.

deploy/helm/rbgs/templates/chart-version.yaml
```diff
@@ -0,0 +1,16 @@
...
+data:
+  chartVersion: {{ .Chart.Version | quote }}
```
cheyang: Checking intent: `.Chart.Version` here is `0.8.0`, `.Chart.AppVersion` is `0.8.0-16705159`. If the eventual consumer is upgrade gating or "what code is actually running", AppVersion is arguably the more useful signal. If it's chart-schema compatibility, Chart.Version is right. Either way worth one line in the PR body.

deploy/helm/rbgs/templates/chart-version.yaml
```diff
@@ -0,0 +1,16 @@
+{{- $markerName := printf "%s-chart-version" .Release.Name }}
```
cheyang: `<release>-chart-version` is inconsistent with the other hook resources in this chart — `crd-upgrade.yaml` names its Job / ServiceAccount / ClusterRole / ClusterRoleBinding with a `rbgs-` prefix (`rbgs-crds-upgrade`). This is the only hook that skips it.

Including the chart name also makes the ConfigMap self-describing under `kubectl get cm`, and keeps the marker's identity stable across a chart rename or split.

```suggestion
{{- $markerName := printf "%s-rbgs-chart-version" .Release.Name }}
```

If you take this, please also update the README bullet in the same PR — it currently describes the marker without naming it, so either spell out the new pattern (`<release>-rbgs-chart-version`) or keep the description generic on purpose.

deploy/helm/rbgs/templates/chart-version.yaml
```diff
@@ -0,0 +1,16 @@
...
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ $markerName }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+  annotations:
+    # A fixed-name hook must remove the previous marker before it can write the new one.
```
cheyang: nit: this reads as if `before-hook-creation` is what retains the marker across upgrades. `before-hook-creation` is Helm's default delete policy — it only handles removing the previous run's resource. Retention actually comes from *not* setting `hook-succeeded`. Something like "no `hook-succeeded` — the marker must outlive the hook so the next upgrade can read it" says it more directly.

deploy/helm/rbgs/templates/chart-version.yaml
```diff
@@ -0,0 +1,16 @@
+{{- $markerName := printf "%s-chart-version" .Release.Name }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ $markerName }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
```
cheyang: nit: only `app.kubernetes.io/managed-by` is set. Peer templates in this chart (e.g. `webhook/manifests.yaml`) also carry `app.kubernetes.io/name`. Consider adding the conventional `app.kubernetes.io/name`, `app.kubernetes.io/instance: {{ .Release.Name }}`, and `helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version }}` so the marker is self-describing and easy to select.

cheyang (changes requested):
Blocking: no test asserts the new template renders. The repo already has `test/chart/deprecated_workload_types_test.go` and `make test-chart`, and #449 (which this marker was extracted from) shipped `test/chart/chart_version_test.go` alongside. A minimum here would be a `helm template` assertion — the ConfigMap renders as `<release>-chart-version`, the hook annotations match what the file claims, and `data.chartVersion` equals `Chart.yaml`'s `version`.

Also, since #449 is closed and the reader is not in this PR, is there a tracking issue for the consumer side? Worth linking from the PR body so the marker doesn't sit unused indefinitely.

Inline: naming consistency with the other hook resources in this chart, uninstall cleanup, one comment wording nit, standard labels, and a `Chart.Version` vs `Chart.AppVersion` question.

cheyang (approved):
/approve
/lgtm

PR #459, opened 2026-09-07
fix: count scheduled Pods in role replica units

cyf-2002:
### Ⅰ. Motivation

`OrderScheduled` compares `ScheduledReplicas` against `RoleStatus.Replicas`, but the controller supplies a raw scheduled Pod count while multi-Pod role replicas are counted as RoleInstances. With two current replicas containing four Pods each, four scheduled Pods already pass the `4 >= 2` check and allow the next scaling batch, even though half the current batch is still unscheduled.

### Ⅱ. Modifications

Normalize the scheduled Pod count by the role's Pods-per-replica value using the existing `ComputeSubGroupSize` helper. The progression check therefore requires eight scheduled Pods for two four-Pod replicas. Preserve the single-Pod behavior and return zero for patterns that produce no Pods to avoid division by zero. Document the multi-Pod progression threshold.

This is a count-unit correction using the existing role-level Pod query and current role pattern; it does not add per-instance scheduling status or change the scaler's batch calculation.

NONE.

### Ⅳ. List the added test cases

A controller regression test exercises `CalculateScalingForAllCoordination` with standalone, leader-worker, custom-component, default component-size, empty-component, and zero-sized-component patterns. It verifies that partially scheduled batches wait and fully scheduled batches advance without requiring Pod readiness.

Before the fix, the leader-worker and custom-component cases with 4/8 and 7/8 scheduled Pods failed: the target advanced to four replicas instead of remaining at two. The same cases pass after the fix.

Passed locally with Go 1.26.1:

```sh
go test ./internal/controller/workloads ./pkg/coordination/coordinationscaling -count=1
go vet ./internal/controller/workloads ./pkg/coordination/coordinationscaling
golangci-lint run --new-from-rev=upstream/main ./internal/controller/workloads/... ./pkg/coordination/coordinationscaling/...
BASE_REF=upstream/main make copyright-check
```

The lint check used the repository-pinned golangci-lint v2.13.1. Changed Go files were formatted with gofmt; `git diff --check` passed.

The regression test uses the controller-runtime fake client. No real-cluster E2E was run.

## Checklist

- [x] Format changed Go code.

### DIFF
--- a/internal/controller/workloads/rolebasedgroup_controller.go
+++ b/internal/controller/workloads/rolebasedgroup_controller.go
@@ -1375,12 +1375,21 @@ func (r *RoleBasedGroupReconciler) CalculateScalingForAllCoordination(
 	return result, nil
 }
 
-// getScheduledReplicas queries the number of scheduled pods (with nodeName) for a given role.
+// getScheduledReplicas converts scheduled Pods into the replica units used by RoleStatus.
 func (r *RoleBasedGroupReconciler) getScheduledReplicas(
 	ctx context.Context,
 	rbg *workloadsv1alpha2.RoleBasedGroup,
 	roleName string,
 ) (int32, error) {
+	role, err := rbg.GetRole(roleName)
+	if err != nil {
+		return 0, err
+	}
+	podsPerReplica := workloadsv1alpha2.ComputeSubGroupSize(role)
+	if podsPerReplica == 0 {
+		return 0, nil
+	}
+
 	podList := &corev1.PodList{}
 	labelSelector := client.MatchingLabels{
 		constants.GroupNameLabelKey: rbg.Name,
@@ -1398,7 +1407,10 @@ func (r *RoleBasedGroupReconciler) getScheduledReplicas(
 		}
 	}
 
-	return scheduled, nil
+	// CurrentReplicas counts instances for multi-Pod patterns. Keep the comparison
+	// in replica units so the next batch waits for currentReplicas * podsPerReplica
+	// scheduled Pods, rather than advancing after only currentReplicas Pods.
+	return scheduled / podsPerReplica, nil
 }
 
 // CalculateRollingUpdateForAllCoordination calculates rolling update strategies for all coordination policies.
--- a/internal/controller/workloads/rolebasedgroup_controller.go
+++ b/internal/controller/workloads/rolebasedgroup_controller.go
@@ -1337,7 +1337,7 @@ func (r *RoleBasedGroupReconciler) CalculateScalingForAllCoordination(
 			}
 
 			// Query scheduled replicas from pods
-			scheduled, err := r.getScheduledReplicas(ctx, rbg, roleName)
+			scheduled, err := r.getScheduledReplicas(ctx, rbg, roleName, current)
 			if err != nil {
 				return nil, fmt.Errorf("failed to query scheduled replicas for role %s: %w", roleName, err)
 			}
@@ -1380,14 +1380,17 @@ func (r *RoleBasedGroupReconciler) getScheduledReplicas(
 	ctx context.Context,
 	rbg *workloadsv1alpha2.RoleBasedGroup,
 	roleName string,
+	currentReplicas int32,
 ) (int32, error) {
 	role, err := rbg.GetRole(roleName)
 	if err != nil {
-		return 0, err
+		// A separate policy may still reference a role removed from the RBG.
+		return 0, nil
 	}
 	podsPerReplica := workloadsv1alpha2.ComputeSubGroupSize(role)
 	if podsPerReplica == 0 {
-		return 0, nil
+		// With no Pods to schedule, every current replica satisfies OrderScheduled.
+		return currentReplicas, nil
 	}
 
 	podList := &corev1.PodList{}
--- a/internal/controller/workloads/rolebasedgroup_controller.go
+++ b/internal/controller/workloads/rolebasedgroup_controller.go
@@ -1336,18 +1336,19 @@ func (r *RoleBasedGroupReconciler) CalculateScalingForAllCoordination(
 				}
 			}
 
-			// Query scheduled replicas from pods
-			scheduled, err := r.getScheduledReplicas(ctx, rbg, roleName, current)
+			// Count scheduled Pods and the Pods expected from the current replicas.
+			scheduled, expected, err := r.getPodSchedulingCounts(ctx, rbg, roleName, current)
 			if err != nil {
-				return nil, fmt.Errorf("failed to query scheduled replicas for role %s: %w", roleName, err)
+				return nil, fmt.Errorf("failed to query Pod scheduling counts for role %s: %w", roleName, err)
 			}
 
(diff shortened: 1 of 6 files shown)


internal/controller/workloads/rolebasedgroup_controller.go
```diff
@@ -1375,12 +1375,21 @@ func (r *RoleBasedGroupReconciler) CalculateScalingForAllCoordination(
...
-// getScheduledReplicas queries the number of scheduled pods (with nodeName) for a given role.
+// getScheduledReplicas converts scheduled Pods into the replica units used by RoleStatus.
 func (r *RoleBasedGroupReconciler) getScheduledReplicas(
 	ctx context.Context,
 	rbg *workloadsv1alpha2.RoleBasedGroup,
 	roleName string,
 ) (int32, error) {
+	role, err := rbg.GetRole(roleName)
+	if err != nil {
+		return 0, err
```
diw-zw: `getScheduledReplicas` now calls `rbg.GetRole(roleName)` first and returns an error when the role is not found, which makes `CalculateScalingForAllCoordination` fail entirely — emitting a `FailedCalculateScaling` warning event and failing the reconcile on every retry. CoordinatedPolicy is a separate CR, so removing a role from the RBG without updating the policy is an easy operational mistake. Before this change that scenario was silently harmless (desired=0, current=0, skipped by the scaler); after it, the reconcile keeps failing. Consider returning 0 (or skipping the role) instead of a hard error when the role is not found, to preserve the previous fault-tolerant semantics.

internal/controller/workloads/rolebasedgroup_controller.go
```diff
@@ -1375,12 +1375,21 @@ func (r *RoleBasedGroupReconciler) CalculateScalingForAllCoordination(
...
+	}
+	podsPerReplica := workloadsv1alpha2.ComputeSubGroupSize(role)
+	if podsPerReplica == 0 {
```
diw-zw: When components are empty or all sized zero, `podsPerReplica == 0` so this returns 0, and `0 < CurrentReplicas` makes `canProceedToNextBatch` wait forever — a coordination deadlock. The old code was equally blocked (0 Pods scheduled), so this is not a regression introduced by this PR; however, if CEL validation allows such patterns, consider skipping the progression check for roles that produce no Pods, and mention this behavior in the doc.

internal/controller/workloads/rolebasedgroup_controller_test.go
```diff
@@ -2390,3 +2390,114 @@ func TestReconcileIncompatibleGangConfig(t *testing.T) {
 	assert.Zero(t, result.RequeueAfter)
 	assert.Contains(t, drainEvents(), "Warning FailedReconcilePodGroup podgroup apiserver hiccup")
 }
+
+// OrderScheduled must wait for every Pod in the current replica batch, including
+// when a role replica creates more than one Pod.
+func TestCalculateScalingForAllCoordination_OrderScheduled(t *testing.T) {
```
diw-zw, line 2396: The regression cases cover all patterns and the truncation behavior well. Minor nit: this relies on the fake client and does not exercise real cluster timing (no E2E, noted in the PR) — acceptable for this change.

internal/controller/workloads/rolebasedgroup_controller.go
cyf-2002: Fixed in 8e91ade6. A missing role now returns `0, nil`, preserving the previous behavior for stale policy references while still propagating Pod list errors. Added regression cases with the removed role's status absent or still reporting replicas; both verify that the remaining role can advance. These cases failed before the fix and pass now.

internal/controller/workloads/rolebasedgroup_controller.go
cyf-2002: Confirmed that the current CRD validation and admission webhook do not reject empty components or all-zero component sizes. Fixed in 8e91ade6: when `podsPerReplica == 0`, the helper returns the current replica count so `OrderScheduled` is satisfied without waiting for nonexistent Pods. The normal `maxSkew` batch limits and `OrderReady` checks remain in effect. Updated the feature documentation and regression tests; the zero-Pod cases failed before the fix and now pass, including checks that batch limits and readiness gating are preserved.

NoobDream2568:
So what we're expecting here is something like `scheduledReplicas` — `OrderReady` already gates on `readyReplicas`. This PR derives it by counting scheduled Pods: **the next batch can only start once all Pods of the current batch are scheduled**. In the normal case, this is equivalent to `scheduledReplicas`.

Could we skip Pods with `DeletionTimestamp != nil` here? It is one line, and it would make this consistent with `OrderReady` side.

internal/controller/workloads/rolebasedgroup_controller.go
```diff
@@ -1375,12 +1375,21 @@ func (r *RoleBasedGroupReconciler) CalculateScalingForAllCoordination(
 	return result, nil
 }
 
-// getScheduledReplicas queries the number of scheduled pods (with nodeName) for a given role.
+// getScheduledReplicas converts scheduled Pods into the replica units used by RoleStatus.
```
diw-zw: Minor follow-up on semantics: the value returned here is a lower bound, not an exact count of fully scheduled replicas. With podsPerReplica=10 and 15 scheduled Pods, `15/10 = 1` whether the Pods are distributed 10+5 or 5+5+5. That is fine for the only consumer, `canProceedToNextBatch`, because floor only reaches `CurrentReplicas` when every Pod of the current batch is scheduled — the exact boundary it checks (ceil would wrongly advance a partially scheduled batch). However, `RoleScalingState.ScheduledReplicas` in scaler.go is documented as "Number of replicas that have been scheduled", which overstates the semantics. Consider updating that field comment to something like "replica units (a lower bound)" so future consumers (logging, skew display, other strategies) do not treat it as an exact per-replica count.

cyf-2002: Would it make sense to express the OrderScheduled gate directly in Pod counts? We could replace RoleScalingState.ScheduledReplicas with ScheduledPods and ExpectedPods, where ExpectedPods = CurrentReplicas * podsPerReplica, and proceed when ScheduledPods >= ExpectedPods. This would avoid implying that the quotient represents an exact count of fully scheduled replicas.
For roles that produce no Pods, both counts would be zero, so the scheduling gate would naturally be satisfied without returning a synthetic replica count. Replica-based batch sizing and the OrderReady check would remain unchanged. We would still need to exclude terminating Pods and handle negative component sizes correctly.
Would you prefer this approach over keeping the replica-unit conversion and clarifying its documentation?

cyf-2002: Implemented in `0084bd44`: `OrderScheduled` now compares `ScheduledPods` with `ExpectedPods` directly, avoiding ambiguous replica conversion and naturally handling zero-Pod roles. Expected counts include Pods not yet created; terminating Pods are excluded, and negative component sizes contribute zero. Replica-based batch limits and `OrderReady` are unchanged. Regression tests pass locally; no real-cluster E2E was run.

diw-zw (approved):

PR #461, opened 2026-09-08
fix: leave Pod Ready condition to kubelet; controller writes only readiness gates

### LINKED ISSUE #460: [Bug] Pod Ready flaps because the controller writes a kubelet-owned condition
### Checklist

### Describe the bug

`Ready` is kubelet-owned (see `PodConditionsByKubelet` in `pkg/kubelet/types`). Writing it from a
controller makes the controller a second writer, and kubelet's status manager cache then diverges
from the API server. On its next periodic sync kubelet republishes the stale value it has cached,
overwriting what the controller wrote.

This produces two distinct defects.

### Symptom A — a ready Pod briefly reports not-ready (`removeNotReadyKey`)

A newly created Pod reaches `Ready=True`, then **0-10 s later** flips to `Ready=False` for ~30 ms
and back. It cascades Pod → RoleInstance → RoleInstanceSet → RoleBasedGroup, so a RoleBasedGroup
that has just reported `Ready=True` reports `Ready=False` again for no reason. Anything consuming
readiness sees a serving replica stop serving: a rolling-update availability budget, an endpoints
consumer, or a test asserting that ready replicas never drop.

For example:

```text
12:52:14.021 POD  chain-1-test-role-0  Ready True   uid=f73a12eb  <- came up, serving normally
12:52:14.029 RBG  chain-1              Ready True
               ... 9.2 s pass; nothing is changed on this group, no rollout in flight ...
12:52:23.242 POD  chain-1-test-role-0  Ready False  uid=f73a12eb  <- the flap, same Pod
12:52:23.250 RI   chain-1-test-role-0  False,False
12:52:23.258 RIS  chain-1-test-role    readyReplicas 1 -> 0
12:52:23.269 RBG  chain-1              Ready False
12:52:23.309 RI   chain-1-test-role-0  True,True
12:52:23.347 POD  chain-1-test-role-0  Ready True   uid=f73a12eb
12:52:23.359 RBG  chain-1              Ready True
```

### Symptom B — a Pod deliberately taken out of rotation briefly rejoins it (`addNotReadyKey`)

`markPodNotReady` is computed **once per RoleInstance** and applied to every Pod
that instance owns (`pkg/reconciler/roleinstance/instance_status.go`, `updatePodsLifeCycle`):

```go
allPodRuntimeReady := r.allPodsRuntimeReady(ctx, instance, pods)   // false if ANY pod is not ready
markPodNotReady = inInplaceUpdating || !allPodRuntimeReady
...
UpdatePodLifecycle(instance, pod, markPodNotReady)                 // same value for every pod
```

`IsPodRuntimeReady` requires `Phase=Running` plus `ContainersReady` and `InPlaceUpdateReady` both
True. So when one Pod of a multi-Pod RoleInstance is unhealthy, its **healthy siblings** are marked
not-ready too — deliberately, since `RoleInstanceReadyOnAllPodReady` is the default policy — and
each of them gets `addNotReadyKey`, which writes `InstancePodReady=False` plus `Ready=False` while
its own container stays `ContainersReady=True`.

Seconds later kubelet republishes its cached `Ready=True`, so a Pod the controller intentionally
pulled out of rotation **reappears as ready and can be re-added to the endpoints**. For
`leaderWorkerPattern` — that means the leader rejoins rotation while its worker is missing.

Measured: `leaderWorkerPattern` with `size: 2` (one leader + one worker in a single RoleInstance),
worker Pod deleted:

```text
10:39:58.987 sibling-prefill-0-0  Ready=True   gate=True   CTN=True   <- steady, leader healthy
10:40:02.667 === delete sibling-prefill-0-1 (the worker) ===
10:40:04.478 sibling-prefill-0-0  Ready=False  gate=False  CTN=True   <- controller pulls the healthy leader out
10:40:12.578 sibling-prefill-0-0  Ready=True   gate=False  CTN=True   <- STALE True republished
10:40:12.603 sibling-prefill-0-0  Ready=False  gate=False  CTN=True   <- corrected 25ms later
10:40:44.812 sibling-prefill-0-0  Ready=False  gate=True   CTN=True   <- worker back
10:40:44.853 sibling-prefill-0-0  Ready=True   gate=True   CTN=True
```

The leader's UID never changes and `ContainersReady` stays `True` throughout — it is healthy the
whole time.

### Reproduction

Only a Pod recreation is needed.

Watch with `kubectl get -w`, not a polling loop: the flap lasts 25-35 ms and polling will miss it.

### Variant A — minimal, reproduces symptom A

```bash
cat <<'EOF' | kubectl apply -f -
kind: RoleBasedGroup
metadata:
  name: flap
spec:
  roles:
    - name: role-1
      replicas: 1
      standalonePattern:
        template:
          spec:
            containers:
              - name: nginx
                image: registry.cn-hangzhou.aliyuncs.com/acs-sample/nginx:latest
                imagePullPolicy: IfNotPresent
EOF

kubectl wait --for=condition=Ready pod/flap-role-1-0 --timeout=180s
# Let kubelet's status cache settle across at least one 10s tick before provoking it.
sleep 20
```

Start the watcher, then delete the Pod so the RoleInstance recreates it:

```bash
kubectl get pod -w --no-headers -o custom-columns=\
'READY:.status.conditions[?(@.type=="Ready")].status,MSG:.status.conditions[?(@.type=="Ready")].message,IU:.status.conditions[?(@.type=="InPlaceUpdateReady")].status,IP:.status.conditions[?(@.type=="InstancePodReady")].status,CTN:.status.conditions[?(@.type=="ContainersReady")].status,UID:.metadata.uid' \
  | while IFS= read -r l; do printf '%s %s\n' "$(date +%H:%M:%S.%N | cut -c1-12)" "$l"; done \
  | tee /tmp/repro.log &

sleep 4
kubectl delete pod flap-role-1-0 --wait=false
```

Follow the **replacement** Pod, i.e. the second UID in the log. Actual output:

```text
11:21:15.623  Ready=<none>  <none>                                 IU=True  IP=<none>  <- created
11:21:15.641  Ready=False   containers with unready status: [nginx]         CTN=False
11:21:15.662  Ready=False   containers with unready status: [nginx] IP=False CTN=False  <- gate created
11:21:16.535  Ready=False   the status of pod readiness gate "InstancePodReady"
                            is not "True", but False                IP=False CTN=True   <- kubelet
11:21:16.567  Ready=True    <none>                                  IP=True  CTN=True   <- controller writes Ready
11:21:22.573  Ready=False   the status of pod readiness gate "InstancePodReady"
                            is not "True", but False                IP=True  CTN=True   <- STALE +6.0s
11:21:22.605  Ready=True    <none>                                  IP=True  CTN=True   <- fixed +32ms
```

The `22.573` line is the bug: its message claims `InstancePodReady` is not `True` while the same
object reports `IP=True`.

### Environment

rbg version: v0.8.0
Kubernetes: v1.35.0 (kind)

The behaviour depends on kubelet's `syncPeriod = 10 * time.Second` and on
`NeedToReconcilePodReadiness` comparing Status *and* Message, both verified against v1.35.0. Other
kubelet versions were not tested, but neither constant is recent.

NoobDream2568:

fixes #460 

The two cases differ by an order of magnitude, so quote the right one:

| Situation | Kubelet picks up the gate change in | Measured |
|-----------|-------------------------------------|----------|
| Gate transition on an already-settled Pod (Repro 2) | tens of milliseconds | 17 ms, 36 ms |
| A newly created Pod reaching `Ready` for the first time (Repro 1) | about one second | 992 ms, 981 ms |

The ~1 s case is worth knowing about: it is what a rolling update or a scale-up pays per Pod. It
looks like kubelet's regular sync cadence rather than the RECONCILE path, because a Pod that is
still starting up has a busy pod worker. Pods come up in parallel, so the wall-clock cost of a
scale-up is roughly one second total; a strictly serial recreate rollout pays it once per group.

| | before | after |
|---|---|---|
| Controller writes | gate + `Ready` | gate only |
| Kubelet's first reaction | none (it agrees with us) | writes the correct `Ready` |
| At the next 10s tick | stale value republished for 25-35 ms | nothing |
| Cost of the fix | — | 17-36 ms on a settled Pod, ~1 s on a new one |

#### Two remaining callers that are *not* bugs

`UpdatePodReadyCondition` is still called from the in-place update paths:

- `pkg/inplace/pod/inplaceupdate/inplace_update.go`, in `updateCondition` when a gate is set False
  at the start of an update
- `pkg/reconciler/roleinstance/inplaceupdate/inplaceupdate.go`, in `updateInstanceReadyCondition`

Both are safe because each is immediately followed, in the same reconcile, by a write that
`podsDifferSemantically` **does** compare:

| Caller | Followed by | Effect on kubelet |
|--------|-------------|-------------------|
| `inplace_update.go` `Update()` | `updatePodInPlace` patches the Pod **spec** (container image) | spec change → `needUpdate` → `HandlePodUpdates` → full pod sync → cache refreshed |
| `inplaceupdate.go` `onlyUpdateRevision` | sets `controller-revision-hash` **label** + `UpdatePod` | label change → same |

The rule is therefore: **writing `Ready` is only harmful when it is the sole effect of the
reconcile.** A pure status write leaves kubelet's cache stale; a status write accompanied by a
spec or label change forces kubelet to re-sync and refresh it.

## Checklist

### DIFF
--- a/pkg/inplace/pod/readiness/pod_readiness_utils.go
+++ b/pkg/inplace/pod/readiness/pod_readiness_utils.go
@@ -22,7 +22,6 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/util/retry"
-	util "sigs.k8s.io/rbgs/pkg/inplace/pod"
 	podadapter "sigs.k8s.io/rbgs/pkg/inplace/pod/clientadapter"
 )
 
@@ -61,8 +60,9 @@ func addNotReadyKey(adp podadapter.Adapter, pod *v1.Pod, msg Message, condType v
 			condition.LastTransitionTime = metav1.Now()
 		}
 
-		// set pod ready condition to "False"
-		util.UpdatePodReadyCondition(newPod)
+		// Write only the gate condition. Ready is kubelet-owned: kubelet derives it from
+		// ContainersReady plus the readiness gates. So writing it anyway is harmful:
+		// might shortly flip the pod back to the state we just moved it away from.
 		if err = adp.UpdatePodStatus(newPod); err != nil {
 			return err
 		}
@@ -96,8 +96,7 @@ func removeNotReadyKey(adp podadapter.Adapter, pod *v1.Pod, msg Message, condTyp
 		}
 		condition.Message = messages.dump()
 		condition.LastTransitionTime = metav1.Now()
-		// Re-evaluate pod Ready condition after updating readiness gate
-		util.UpdatePodReadyCondition(newPod)
+		// Ready is left to kubelet here too; see the note in addNotReadyKey.
 		if err = adp.UpdatePodStatus(newPod); err != nil {
 			return err
 		}
--- a/pkg/inplace/pod/readiness/pod_readiness_utils_test.go
+++ b/pkg/inplace/pod/readiness/pod_readiness_utils_test.go
@@ -17,8 +17,11 @@ limitations under the License.
 package readiness
 
 import (
+	"encoding/json"
 	"fmt"
+	"reflect"
 	"testing"
+	"time"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -26,33 +29,62 @@ import (
 	podadapter "sigs.k8s.io/rbgs/pkg/inplace/pod/clientadapter"
 )
 
-// fakeAdapter is a minimal in-memory Adapter for unit tests.
-type fakeAdapter struct {
-	pod *v1.Pod
+var gateCondType = constants.InstancePodReadyConditionType
+
+// recordingAdapter is an in-memory Adapter that records the exact pod object
+// passed to every UpdatePodStatus call. Deliberately, it does nothing else.
+//
+// A previous version of this fake additionally recomputed the Ready condition
+// the way kubelet would, immediately after every write. That made every
+// assertion on Ready converge to the same value whether or not the controller
+// itself had written Ready, so the tests passed both with and without the
+// write they were meant to guard against. Here, any Ready value observed in a
+// recorded write is a value the controller put there itself.
+type recordingAdapter struct {
+	pod    *v1.Pod
+	writes []*v1.Pod
 }
 
-func (f *fakeAdapter) GetPod(namespace, name string) (*v1.Pod, error) {
+var _ podadapter.Adapter = &recordingAdapter{}
+
+func (f *recordingAdapter) GetPod(namespace, name string) (*v1.Pod, error) {
 	if f.pod == nil || f.pod.Namespace != namespace || f.pod.Name != name {
 		return nil, fmt.Errorf("pod %s/%s not found", namespace, name)
 	}
 	return f.pod.DeepCopy(), nil
 }
 
-func (f *fakeAdapter) UpdatePod(pod *v1.Pod) error {
+func (f *recordingAdapter) UpdatePod(pod *v1.Pod) error {
 	f.pod = pod.DeepCopy()
 	return nil
 }
 
-func (f *fakeAdapter) UpdatePodStatus(pod *v1.Pod) error {
+func (f *recordingAdapter) UpdatePodStatus(pod *v1.Pod) error {
+	f.writes = append(f.writes, pod.DeepCopy())
 	f.pod = pod.DeepCopy()
 	return nil
 }
 
-var _ podadapter.Adapter = &fakeAdapter{}
+// kubeletManagedReady returns a deliberately distinctive Ready condition.
+// Ready is kubelet-owned: kubelet derives it from ContainersReady plus the
+// readiness gates, and a controller that also writes it makes kubelet's
+// status cache diverge from the API server, so kubelet's next 10s tick
+// republishes a stale value (the observed True->False->True flap). The
+// marker Reason/Message make any controller write to Ready show up in a
+// DeepEqual even when the Status it writes happens to equal the current one.
+func kubeletManagedReady(status v1.ConditionStatus) v1.PodCondition {
+	return v1.PodCondition{
+		Type:               v1.PodReady,
+		Status:             status,
+		Reason:             "KubeletManaged",
+		Message:            "owned by kubelet; the controller must not rewrite this condition",
+		LastTransitionTime: metav1.NewTime(time.Date(2026, time.January, 1, 0, 0, 0, 0, time.UTC)),
+	}
+}
 
-// newReadyPod creates a pod that simulates the state after KWOK's pod-ready
-// stage fires: all conditions True, both readiness gates present and True.
-func newReadyPod() *v1.Pod {
+// servingPod is a pod that has fully come up: containers ready, both gates
+// satisfied, and kubelet has published Ready=True.
+func servingPod() *v1.Pod {
 	return &v1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: "test-ns",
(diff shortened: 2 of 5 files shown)


diw-zw:
Small note on the description: `podsDifferSemantically` is kubelet's upstream function (used on kubelet's sync path to decide whether a spec/label change forces a full pod sync), not a function in this repo — phrasing it as "kubelet's `podsDifferSemantically`" would save readers a grep.

The argument itself checks out: I verified both remaining callers — `Update()` writes the gate-False condition immediately before `updatePodInPlace` patches the pod spec, and `onlyUpdateRevision` sets the revision-hash label via `UpdatePod` — so in both cases kubelet is forced to re-sync and refresh its status cache. The readiness + inplace-update test suites pass locally on this branch.

pkg/inplace/pod/readiness/pod_readiness_utils.go
```diff
@@ -61,8 +60,9 @@ func addNotReadyKey(adp podadapter.Adapter, pod *v1.Pod, msg Message, condType v
 			condition.LastTransitionTime = metav1.Now()
 		}
 
-		// set pod ready condition to "False"
-		util.UpdatePodReadyCondition(newPod)
+		// Write only the gate condition. Ready is kubelet-owned: kubelet derives it from
```
diw-zw, line 63: The fix is right for real clusters — kubelet owns `Ready` and republishes its cached value, so removing this write fixes the second-writer bug. One gap to consider for simulated environments, though: in KWOK there is no kubelet to derive `Ready` from the gate, and the `pod-ready` stage in `test/stress/templates/kwok-stage.yaml` sets `Ready=True` unconditionally without ever evaluating readiness gates.

After this change, in KWOK/stress environments `addNotReadyKey` (gate=False) leaves `Ready=True` — possibly permanently, since on current main the controller's `Ready=False` write is what actually took the pod out of rotation there. Notably, the removed test `TestRemoveNotReadyKey_UpdatesPodReadyCondition` existed precisely because of KWOK behavior ("pods stay Ready=False permanently in KWOK environments" — i.e. nothing in KWOK corrects `Ready` on its own).

Could the `pod-ready` stage be made gate-aware (its selector already uses expressions over conditions), or should the PR at least call out the fidelity loss for stress tests? Otherwise take-out-of-rotation silently becomes invisible in the stress environment.

pkg/inplace/pod/readiness/pod_readiness_utils.go
```diff
@@ -61,8 +60,9 @@ func addNotReadyKey(adp podadapter.Adapter, pod *v1.Pod, msg Message, condType v
...
+		// ContainersReady plus the readiness gates. So writing it anyway is harmful:
+		// might shortly flipping the pod back to the state we just moved it away from.
```
diw-zw: Nit: "might shortly flipping the pod back" → "might shortly flip the pod back".

pkg/inplace/pod/readiness/pod_readiness_utils.go
NoobDream2568: I'll try fix it in `kwok-stage.yaml`

test/stress/templates/kwok-stage.yaml
```diff
@@ -111,6 +121,12 @@ spec:
         operator: 'NotIn'
         values:
           - "True"
+      # Only promote to Ready while the controller has not gated the pod out of rotation.
+      # Emulates kubelet deriving Ready from the readiness gates (see file header).
+      - key: '[.status.conditions.[]? | select((.type == "InPlaceUpdateReady" or .type == "InstancePodReady") and .status == "False")] | if length > 0 then "False" else "True" end'
```
cheyang, line 130: Blocking merge on the KWOK companion, not the production code. The gate-aware check you added here in `pod-ready` is correct — but `pod-running` (line 84, unchanged by this PR) still writes `Ready=True` unconditionally on the Pending→Running transition, with no gate check. So a pod the controller has gated out (any readiness gate False) can still be briefly exposed as Ready by `pod-running`, then later flipped back to False by `pod-ready-blocked` — exactly the flap class this PR removes in production, so the stress sim does not faithfully guard against its own regression. Cheapest fix: have `pod-running` write `Ready=False` (set ContainersReady/Initialized/phase=Running only) and let only this gate-aware `pod-ready` stage promote to `Ready=True`, so every Ready=True promotion goes through the gate check. Please fold this in before merge — the KWOK companion is part of this change.

test/stress/templates/kwok-stage.yaml
```diff
@@ -143,6 +159,42 @@ spec:
...
         {{ end }}
 ---
+# Gate-aware counterpart of pod-ready: when the controller has written any readiness gate to
+# False (markPodNotReady / take-out-of-rotation), flip Ready to False exactly like a kubelet
+# would. Only the Ready condition is patched; Pod statuses are merged with a strategic merge
+# patch for Pods, so the controller-owned gate conditions are preserved.
+apiVersion: kwok.x-k8s.io/v1alpha1
+kind: Stage
+metadata:
+  name: pod-ready-blocked
```
cheyang, line 173: This new `pod-ready-blocked` stage is not in the sibling teardown script: `test/stress/scripts/teardown-kwok.sh` (lines 29-32) deletes `pod-initialize`, `pod-running`, `pod-ready`, `pod-delete` but omits `pod-ready-blocked`, so the stage leaks after uninstall. Add `kubectl delete stage pod-ready-blocked --ignore-not-found=true` next to the other deletes. (teardown-kwok.sh isn't in this PR's diff, so I'm anchoring the note here on the stage that's missing its cleanup line.)

pkg/inplace/pod/readiness/pod_readiness_utils.go
cheyang, line 63: Confirmed the fix is correct. The premise (issue #460) reproduces at the code/unit layer: base writes the kubelet-owned `Ready` via `util.UpdatePodReadyCondition` here and in `removeNotReadyKey`, making the controller a second writer whose value kubelet's status-manager cache then republishes stale on its 10s sync. Removing both calls (this + line ~99) so the controller writes only the gate is the right fix — it restores the single-writer invariant. Harness-bites check: the new contract tests fail against buggy base (controller rewrites Ready) and pass on this PR head. (I could not trigger the live 25-35ms flap on the ACK cluster — its installed v0.8.0 controller is stuck on unrelated legacy `restartPolicy` data, not because the flap is absent.)

pkg/inplace/pod/readiness/pod_readiness_utils_test.go
```diff
@@ -17,42 +17,74 @@ limitations under the License.
...
 	podadapter "sigs.k8s.io/rbgs/pkg/inplace/pod/clientadapter"
 )
 
-// fakeAdapter is a minimal in-memory Adapter for unit tests.
-type fakeAdapter struct {
-	pod *v1.Pod
+var gateCondType = constants.InstancePodReadyConditionType
+
+// recordingAdapter is an in-memory Adapter that records the exact pod object
+// passed to every UpdatePodStatus call. Deliberately, it does nothing else.
```
cheyang, line 35: Nice improvement on the test harness. The old `fakeAdapter` recomputed `Ready` the way kubelet would after every write, so it passed both with and without the bug it was meant to guard — the tests never actually pinned the controller's own Ready write. `recordingAdapter` recording writes verbatim, plus the `kubeletManagedReady` marker condition with a distinctive Reason/Message, makes any controller write to Ready show up in `reflect.DeepEqual` even when the Status happens to match. This is the kind of test that would have caught the regression originally.

cheyang (changes requested):
Production fix is correct and should merge; the KWOK companion needs two small fixes first. Premise confirmed: base writes the kubelet-owned `Ready` from `addNotReadyKey`/`removeNotReadyKey` (a second writer), so kubelet republishes the stale value on its 10s sync — the True→False→True flap. Removing the two `UpdatePodReadyCondition` calls so the controller writes only the readiness gate restores the single-writer invariant; the rewritten tests now actually bite (the old fakeAdapter masked the bug — verified: tests fail on buggy base, pass on this head). On the two remaining `UpdatePodReadyCondition` callers in the in-place paths: the justification is sound in the happy path (each is followed in the same function by a spec/label write that forces kubelet to re-sync), but those are separate, non-atomic API calls, not a transaction — so `Ready=False` can land while the spec/label write fails. I would not delete those callers in this PR, but the failure path is worth a separate follow-up. Two KWOK items to fold in before merge: (1) `pod-running` still writes `Ready=True` unconditionally and bypasses the new gate-aware check, so a gated-out pod can be briefly exposed Ready — have `pod-running` leave `Ready=False` and let only the gate-aware `pod-ready` promote; (2) `pod-ready-blocked` is missing from `teardown-kwok.sh` and leaks after uninstall. Neither touches production code. Verified on branch verify/pr461-pod-ready-kubelet — evidence under docs/verification/pr461-pod-ready-kubelet/.

test/stress/templates/kwok-stage.yaml
cheyang, line 130: Blocking merge on the KWOK companion, not the production code. The gate-aware check you added here in `pod-ready` is correct — but `pod-running` (line 84, unchanged by this PR) still writes `Ready=True` unconditionally on the Pending→Running transition, with no gate check. So a pod the controller has gated out (any readiness gate False) can still be briefly exposed as Ready by `pod-running`, then later flipped back to False by `pod-ready-blocked` — exactly the flap class this PR removes in production, so the stress sim does not faithfully guard against its own regression. Cheapest fix: have `pod-running` write `Ready=False` (set ContainersReady/Initialized/phase=Running only) and let only this gate-aware `pod-ready` stage promote to `Ready=True`, so every Ready=True promotion goes through the gate check. Please fold this in before merge — the KWOK companion is part of this change.

test/stress/templates/kwok-stage.yaml
cheyang, line 173: This new `pod-ready-blocked` stage is not in the sibling teardown script: `test/stress/scripts/teardown-kwok.sh` (lines 29-32) deletes `pod-initialize`, `pod-running`, `pod-ready`, `pod-delete` but omits `pod-ready-blocked`, so the stage leaks after uninstall. Add `kubectl delete stage pod-ready-blocked --ignore-not-found=true` next to the other deletes. (teardown-kwok.sh isn't in this PR's diff, so I'm anchoring the note here on the stage that's missing its cleanup line.)

pkg/inplace/pod/readiness/pod_readiness_utils.go
cheyang, line 63: Confirmed the fix is correct. The premise (issue #460) reproduces at the code/unit layer: base writes the kubelet-owned `Ready` via `util.UpdatePodReadyCondition` here and in `removeNotReadyKey`, making the controller a second writer whose value kubelet's status-manager cache then republishes stale on its 10s sync. Removing both calls (this + line ~99) so the controller writes only the gate is the right fix — it restores the single-writer invariant. Harness-bites check: the new contract tests fail against buggy base (controller rewrites Ready) and pass on this PR head. (I could not trigger the live 25-35ms flap on the ACK cluster — its installed v0.8.0 controller is stuck on unrelated legacy `restartPolicy` data, not because the flap is absent.)

pkg/inplace/pod/readiness/pod_readiness_utils_test.go
cheyang, line 35: Nice improvement on the test harness. The old `fakeAdapter` recomputed `Ready` the way kubelet would after every write, so it passed both with and without the bug it was meant to guard — the tests never actually pinned the controller's own Ready write. `recordingAdapter` recording writes verbatim, plus the `kubeletManagedReady` marker condition with a distinctive Reason/Message, makes any controller write to Ready show up in `reflect.DeepEqual` even when the Status happens to match. This is the kind of test that would have caught the regression originally.

cheyang (changes requested):
Production fix is correct and should merge; the KWOK companion needs two small fixes first. Premise confirmed: base writes the kubelet-owned `Ready` from `addNotReadyKey`/`removeNotReadyKey` (a second writer), so kubelet republishes the stale value on its 10s sync — the True→False→True flap. Removing the two `UpdatePodReadyCondition` calls so the controller writes only the readiness gate restores the single-writer invariant; the rewritten tests now actually bite (the old fakeAdapter masked the bug — verified: tests fail on buggy base, pass on this head). On the two remaining `UpdatePodReadyCondition` callers in the in-place paths: the justification is sound in the happy path (each is followed in the same function by a spec/label write that forces kubelet to re-sync), but those are separate, non-atomic API calls, not a transaction — so `Ready=False` can land while the spec/label write fails. I would not delete those callers in this PR, but the failure path is worth a separate follow-up. Two KWOK items to fold in before merge: (1) `pod-running` still writes `Ready=True` unconditionally and bypasses the new gate-aware check, so a gated-out pod can be briefly exposed Ready — have `pod-running` leave `Ready=False` and let only the gate-aware `pod-ready` promote; (2) `pod-ready-blocked` is missing from `teardown-kwok.sh` and leaks after uninstall. Neither touches production code. Verified on branch verify/pr461-pod-ready-kubelet — evidence under docs/verification/pr461-pod-ready-kubelet/.

NoobDream2568:
@cheyang Thanks for suggestions. TBH, I'm not very familiar with KWOK, but I'm trying my best to fix the issues. The RBGs Rolling Update feature could be introduced after this PR gets merged.

cheyang (approved):
Approving. Round 2 (05d465d7) addressed both KWOK companion findings from the prior review and production code is unchanged. `pod-running` now writes `Ready=False` so only the gate-aware `pod-ready` stage promotes a pod to Ready=True — the bypass where a gated-out pod could be briefly exposed Ready is closed. `teardown-kwok.sh` now deletes `pod-ready-blocked`, so the new stage no longer leaks after uninstall. The core fix is unchanged and still proven: base wrote the kubelet-owned `Ready` from `addNotReadyKey`/`removeNotReadyKey` (a second writer → kubelet republished the stale value on its 10s sync), and removing those two `UpdatePodReadyCondition` calls so the controller writes only the readiness gate restores the single-writer invariant; the rewritten tests bite (verified: they fail on buggy base, pass on this head). Unit layer green on 05d465d7; regression scan of the pod-running Ready=False change shows no stuck-state path. No blockers or majors remain. Verified on branch verify/pr461-pod-ready-kubelet — evidence under docs/verification/pr461-pod-ready-kubelet/. (Note: two duplicate CHANGES_REQUESTED reviews from the prior round are still on the PR; GitHub can't delete submitted reviews via API — please dismiss one manually.)

PR #463, opened 2026-09-09
fix(api): heal legacy update strategy types via a mutating webhook

diw-zw:
## Summary
Follow-up to #436, which tightened `spec.roles[].rolloutStrategy.rollingUpdate.type` (RoleBasedGroup / RoleBasedGroupSet) and `spec.updateStrategy.type` (RoleInstanceSet) from free-form strings to CRD enums. That enum breaks writes to objects stored by older releases:

- pre-enum CRDs accepted anything, so stored objects can carry `""` or the v1alpha1 spelling `Recreate` (v1alpha2 spells it `RecreatePod`; the field converts as a literal cast between versions);
- after upgrade, any spec write to such an object fails with 422 — on k8s >= 1.31 validation ratcheting only ignores the invalid value while it stays unchanged, so the first spec change still fails, and older clusters fail even earlier;
- v1alpha1 is still served without an enum, and the conversion webhook did not map `Recreate` -> `RecreatePod`, so writing through v1alpha1 kept producing the invalid value.

Instead of a migration that rewrites stored objects (bumping generations and rolling pods), this PR makes the values self-heal on write: mutating-webhook defaulters for RoleBasedGroup, RoleBasedGroupSet and RoleInstanceSet normalize the type on every create/update (`""` -> `InPlaceIfPossible`, `Recreate` -> `RecreatePod`, valid values pass through). Objects that are never written again are left untouched.

## Changes

- **API** (`api/workloads/v1alpha2`): add `NormalizeUpdateStrategyType` and `LegacyRecreateUpdateStrategyType`; new `RoleBasedGroupDefaulter` / `RoleBasedGroupSetDefaulter` / `RoleInstanceSetDefaulter` mutating webhooks (verbs `create;update`).
- **Conversion** (`api/workloads/v1alpha1`): map `Recreate` <-> `RecreatePod` in both directions; empty stays empty (RestartPolicy precedent) so defaultingapplies later instead of being baked into stored objects.
- **Cert bootstrap**: the controller now patches caBundle onto `rbgs-mutating-webhook-configuration` as well — cert-manager helper, WebhookCertReconciler watch, RBAC, kustomize/helm manifests, registration in `main.go`.
- **Reconciler**: RoleInstanceSet construction normalizes the strategy type too, so clusters running with webhooks disabled handle legacy values.
- **e2e upgrade suite** (v0.7.0 -> current): new fixtures storing `Recreate ` and `""`; asserts the stored values survive the upgrade untouched, then heal to`RecreatePod` / `InPlaceIfPossible` on a metadata-only write.
- **envtest**: all three valid strategy types are accepted; update-path coverage.
- **CI**: the bootstrap gate now waits for caBundle on both webhook configurations; the upgrade-lane failure dump prints the mutating caBundle; the helm README lists the new webhook configuration.
- **KEP-8**: validation table and flowcharts updated — `templatePatch` is optional in v1alpha2 when `templateRef` is set (v1alpha1 keeps requiring it).


pkg/reconciler/roleinstanceset_reconciler.go
```diff
@@ -253,9 +253,9 @@ func (r *RoleInstanceSetReconciler) constructRoleInstanceSetApplyConfiguration(
 
 	if role.RolloutStrategy != nil && role.RolloutStrategy.RollingUpdate != nil {
 		rollingUpdate := role.RolloutStrategy.RollingUpdate
-		if rollingUpdate.Type == "" {
-			rollingUpdate.Type = workloadsv1alpha2.InPlaceIfPossibleUpdateStrategyType
-		}
+		// Normalize legacy values ("" and v1alpha1's "Recreate") so objects that
+		// predate the CRD enum still work when webhooks are disabled.
+		rollingUpdate.Type = workloadsv1alpha2.NormalizeUpdateStrategyType(rollingUpdate.Type)
```
cheyang, line 258: You normalized here for the RBG→RIS link so clusters with webhooks disabled still heal — but the RBGS→child-RBG link has the same shape and is uncovered. `updateExistingRBGs` (`internal/controller/workloads/rolebasedgroupset_controller.go:458`) and `newRBGForSet` (`:529`) copy `rbgset.Spec.GroupTemplate.Spec.Roles` into children verbatim, and `rolesEqual` (`:343`) compares with `reflect.DeepEqual` — so a legacy `Recreate` stored on a pre-webhook RBGS template vs. a webhook-healed `RecreatePod` child is always "different". The RBGS defaulter only fires on RBGS spec create/update, and the controller's only RBGS write is `Status().Update` (`:323`), so a legacy RBGS template is never healed; every reconcile re-issues a child `Update` that the webhook normalizes back to `RecreatePod` → parent(`Recreate`) ≠ child(`RecreatePod`) → `needsUpdate` stays true → the RBGS layer never converges on the legacy-upgrade path, which is this PR's headline scenario. The RIS itself stays valid (you normalize before SSA here), so the stated break is fixed and there's no data corruption — but please mirror this normalization at the RBGS→RBG boundary: deep-copy `GroupTemplate.Spec.Roles` and run `NormalizeUpdateStrategyType` on each `RolloutUpdate.Type` before assigning into children; optionally treat `Recreate`≡`RecreatePod` as equal in `rolesEqual` so a spelling-only delta doesn't rewrite unchanged children.

test/e2e/upgrade/fixtures.go
```diff
@@ -305,6 +315,49 @@ func buildMidRolloutFixture(ns string) *workloadsv1alpha2.RoleBasedGroup {
...
 
+// legacyStrategyRole is the role of both legacy-strategy fixtures. It is shared so
+// the recorded RoleInstanceSet repair in snapshot.go names the right owner.
+const legacyStrategyRole = "worker"
+
+// buildLegacyStrategyFixture stores the v1alpha1 spelling "Recreate" of the update
+// strategy type, which v0.7.0's enum-less CRD accepted and its controller copied
+// verbatim into the RoleInstanceSet. One replica, because the strategy type only
+// shapes how a rollout proceeds and nothing here rolls.
+func buildLegacyStrategyFixture(ns string) *workloadsv1alpha2.RoleBasedGroup {
```
cheyang, line 332: This legacy `Recreate` fixture is a standalone RBG only. There's no legacy `RoleBasedGroupSet` fixture (GroupTemplate carrying `Recreate`), so the RBGS→child divergence path described on `roleinstanceset_reconciler.go:258` is never exercised. Please add a legacy-RBGS fixture (GroupTemplate carrying `Recreate`) and assert that repeated reconcile does not re-update the children.

test/envtest/testcase/rbg/basic_test.go
```diff
@@ -292,6 +293,88 @@ var _ = Describe("RoleBasedGroup Controller", func() {
 
 			expectInvalidCreate(ris)
 		})
+
+		It("Should accept all valid v1alpha2 update strategy types", func() {
```
cheyang, line 297: These cases only create valid strategy types and modify valid objects — they don't (and can't, since `test/envtest/testutil/setup.go` wires no admission webhooks: no `WebhookInstallOptions`, no webhook server) prove the webhook heal that is the PR's core mechanism. Please add a webhook-enabled integration test covering `Recreate`/`""` writes on RBG, RBGS, and RIS, plus a v1alpha1 write.

api/workloads/v1alpha2/roleinstanceset_default.go
```diff
@@ -0,0 +1,46 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// RoleInstanceSetDefaulter normalizes legacy update-strategy type values on
+// RoleInstanceSet writes. Objects stored before the updateStrategy.type enum
+// was introduced can carry "" or the v1alpha1 "Recreate" spelling; normalizing
+// them here keeps such objects writable instead of failing schema validation.
+//
+// +kubebuilder:webhook:path=/mutate-workloads-x-k8s-io-v1alpha2-roleinstanceset,mutating=true,failurePolicy=fail,sideEffects=None,groups=workloads.x-k8s.io,resources=roleinstancesets,verbs=create;update,versions=v1alpha2,name=mroleinstanceset.kb.io,admissionReviewVersions=v1
```
cheyang, line 32: `failurePolicy: Fail` on the new RIS mutating webhook extends the existing Fail-webhook pattern already used by RBG/RBGS, and the cert-manager refactor in this commit keeps its caBundle in sync via the shared `patchOneWebhookConfiguration` path. Bootstrap is gated on the caBundle in CI. Not a new risk class — noting for completeness, no change required.

cheyang (changes requested):
Requesting changes — not on the design (the fix direction is right and the stated break is genuinely fixed), but on one legacy-upgrade gap the PR leaves open.

**What's fixed (verified):** After #436's enum on `RoleInstanceSet.updateStrategy.type`, a `Recreate`/`""` value 422s on RIS writes, and the v0.8.0 controller copies the legacy value verbatim into the RIS so it's never created. This PR heals that: the RBG reconciler normalizes before SSA (`pkg/reconciler/roleinstanceset_reconciler.go:258`), and admission deflectors heal RBG/RBGS/RIS writes. The RIS-creation break is resolved — confirmed live on ACK (k8s v1.36.1) in a prior round, plus unit + envtest.

**What's still open (major):** The RBGS→child-RBG link is not normalized. `rolesEqual` (`internal/controller/workloads/rolebasedgroupset_controller.go:343`) compares with `reflect.DeepEqual`, and `updateExistingRBGs` (:458) / `newRBGForSet` (:529) copy the parent `GroupTemplate` verbatim. The RBGS defaulter only fires on RBGS spec create/update, and the controller's only RBGS write is `Status().Update`, so a legacy pre-webhook RBGS whose stored template still carries `Recreate` is never healed. After the webhook goes live, each controller `Update` of a child is healed `Recreate`→`RecreatePod`, leaving parent(`Recreate`) ≠ child(`RecreatePod`) → `needsUpdate` stays true → the RBGS layer never converges on the legacy-upgrade path. The RIS stays valid (reconciler normalizes before SSA), so there's no data corruption — but this is the PR's headline scenario. The non-convergence holds regardless; whether it's a tight hot-loop or slow churn depends on whether a no-op webhook-healed `Update` re-queues via `Owns`, which I couldn't confirm statically (needs a webhook-wired envtest or live run). Suggested fix: deep-copy `GroupTemplate.Spec.Roles` + `NormalizeUpdateStrategyType` before assigning to children, mirroring the RBG→RIS reconciler normalization.

**Coverage gap (major):** The only suite that runs the upgrade path (`RunUpgradeSpecs`) is in `release-test.yml`, which triggers on `/release-test` (a release gate, not per-PR CI) and runs against the chart's default published images, not this PR's head. The per-PR envtest wires no admission webhooks (`test/envtest/testutil/setup.go`), and the upgrade fixtures carry legacy `Recreate` only on standalone RBGs, never on an RBGS `GroupTemplate`. So the divergence above is uncovered at every layer. Please add (a) a webhook-enabled test for `Recreate`/`""` healing on RBG/RBGS/RIS, and (b) a legacy-RBGS convergence test asserting repeated reconcile does not re-update children.

**Note on a prior bot review's CI claim:** the 13 per-PR checks are all `COMPLETED`/`SUCCESS`; `mergeStateStatus: BLOCKED` reflects only the missing human approval, not a CI failure. The real CI concern is the structural exclusion of the upgrade lane above.

Direction is good — this is about closing the RBGS link, not reworking the approach.

cheyang (approved):
/approve
/lgtm

cheyang (approved):
Approving — the round-2 change request is resolved.

B1 (RBGS parent/child strategy non-convergence on the legacy-upgrade path) is fixed in commit f79aa490. The fix matches the shape the verification harness proved necessary: rolesEqual now deep-copies both sides and normalizes RollingUpdate.Type on each before comparing, and updateExistingRBGs / newRBGForSet write normalizedGroupTemplateRoles(rbgset) so children are created already canonical. Crucially the fix normalizes in the compare path as well as the copy path — copy-only normalization is insufficient because the un-healed legacy parent would otherwise keep needsUpdate true forever. I verified this directly: the reproduction contracts now pass (needsUpdate=false; converges across repeated reconciles), and the canaries that asserted the buggy write-back of Recreate flip to fail (child is now RecreatePod) — exactly the predicted fixed signature.

Both coverage gaps from round 2 are addressed with in-tree tests that pass: (1) a webhook-wired envtest (test/envtest/testcase/webhook) registers the RBG/RBGS/RIS mutating webhooks and proves the admission heal at the integration layer — TestWebhookDefaulting passes; (2) a new legacySetV2 upgrade fixture carries legacy Recreate on an RBGS GroupTemplate with a convergence assertion, covering the exact divergence scenario end-to-end. The RBGS defaulter change is comment-only; the cert-manager changes (conn.Close error handling, MutatingWebhookConfigurationName constant, retry helpers) address the earlier review notes.

Full unit + envtest suite is green with no regressions (controller workloads, api conversion, reconciler, webhook, and the new webhook envtest). The RIS-creation break the PR targets remains fixed (reconciler normalizes before SSA), and there is no data-corruption path.

One non-blocking limitation, carried over and not a merge blocker: the live hot-loop rate on the ACK cluster was not re-observed, because doing so would require mutating live nginx-cluster/test-rbg RIS spec data (the legacy restartPolicy object-vs-string form) without per-action authorization. The unit layer and the webhook-wired envtest now cover the convergence mechanism deterministically, so the live layer is confirmatory rather than load-bearing.

Evidence and the re-verification harness: https://github.com/cheyang/rbg/tree/verify/pr463-legacy-strategy-heal/docs/verification/pr463-legacy-strategy-heal (.last-reviewed = 9c95973c).

PR #466, opened 2026-09-13
fix: validate warmup actions and target RBG failures

googs1025:
### Ⅰ. Motivation

`RoleBasedGroupWarmup` accepted empty image references and could create invalid Pods. A missing target RBG caused repeated NotFound retries with uninformative status.

### Ⅱ. Modifications

- Added CRD validation `images.items.minLength: 1`.
- Added controller validation for empty/whitespace images and empty customized actions.
- Invalid actions now enter `Failed` with `InvalidWarmupSpec` and create no Pods.
- Missing target RBG now enters `Failed` with `InvalidTarget`, records completion, and does not requeue.
- Validation runs before node discovery, including zero-node selectors.
- Added regression tests and updated `keps/129-rbg-warmup/README.md`.

NONE

### Ⅳ. List the added test cases

- Empty image reference fails with `InvalidWarmupSpec`.
- Empty image list and empty customized containers fail before node discovery.
- Missing target RBG fails with `InvalidTarget` and no requeue.
- Existing Warmup controller, Pod construction, timeout, backoff, and status tests pass.

```bash
env -u GOROOT GOCACHE=/private/tmp/rbg-warmup-fixes-cache GOMODCACHE=/private/tmp/rbg-go-modcache-125 GOFLAGS=-mod=mod go test ./internal/controller/workloads ./api/workloads/v1alpha2 -run 'Test(BuildWarmupPod|ComputePermanentlyFailedNodes|CollectPendingNodes|RequeueForTimeout|Reconcile)|TestRoleBasedGroup' -count=1\n```\n\nFocused tests pass. `make manifests` regenerated the CRD and kubectl manifests. Image `rolebasedgroup/rbgs-controller:v0.9.0-warmup-fix` was built and deployed to Minikube; the controller reached `1/1 Running`.\n\n### VI. Special notes for reviews\n\n- This is a one-shot Warmup API: the referenced RBG must exist when reconciled. If absent, the Warmup fails; fix the target and create a new Warmup.\n- No deferred waiting mode is introduced. It should be an explicit future API option if needed.\n- Live server-side YAML validation was limited by intermittent Minikube API Server timeouts; deterministic controller/API tests pass.\n- All commits include DCO sign-off.\n\n## Checklist\n\n- [x] Format code with `gofmt`.\n- [x] Add unit tests.\n- [x] Update Warmup KEP documentation.

### CODE REVIEW

File: internal/controller/workloads/rolebasedgroupwarmup_controller.go [hunk 1]
```diff
@@ -176,6 +183,10 @@ func (r *RoleBasedGroupWarmupReconciler) reconcileUnfinished(ctx context.Context
 
 	desiredNodes, err := r.getDesiredNodesToWarmup(ctx, *warmup)
 	if err != nil {
+		if apierrors.IsNotFound(err) && warmup.Spec.TargetRoleBasedGroup != nil {
```
cheyang, line 186: `getDesiredNodesToWarmup` reads the RBG through the cached client, so this can turn a transient cache miss into a permanent failure. If an RBG and its Warmup are created close together, the Warmup may reconcile before the RBG informer has observed the target even though it already exists in the API server. I reproduced that split view with the cached client returning `NotFound` while `apiReader` could read the RBG; this branch still set `InvalidTarget` and made the one-shot resource unrecoverable. Please confirm `NotFound` with the uncached reader (or keep the retry behavior) before recording a terminal failure, and cover the cached-miss/API-hit case.

internal/controller/workloads/rolebasedgroupwarmup_controller.go
```diff
@@ -168,6 +169,12 @@ func (r *RoleBasedGroupWarmupReconciler) reconcileFinished(ctx context.Context,
 func (r *RoleBasedGroupWarmupReconciler) reconcileUnfinished(ctx context.Context, warmup *workloadsv1alpha2.RoleBasedGroupWarmup) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
 
+	if err := validateWarmupSpec(warmup.Spec); err != nil {
+		logger.Error(err, "invalid warmup specification")
+		return ctrl.Result{}, r.failWarmupJob(ctx, warmup, nil, nil, nil, nil,
```
cheyang: Validation runs before `listWarmupPods`, and this call passes nil pod slices into `failWarmupJob`. Existing objects can still reach this path with active pods—for example, a running Warmup can be updated to a whitespace-only image, which the CRD accepts and the controller rejects. I reproduced that case: reconciliation set the Warmup to `Failed` with `active=0`, but the existing warmup pod remained running. Please list the owned pods before failing validation and pass them through the normal cleanup path, with a regression test that starts from an active pod.

cheyang (dismissed):
Two terminal-failure paths still need correction before merge: a cached `NotFound` can falsely fail a one-shot Warmup whose target already exists, and invalid-spec handling can leave an existing warmup pod running while reporting `active=0`.

internal/controller/workloads/rolebasedgroupwarmup_controller.go
```diff
@@ -176,10 +182,23 @@ func (r *RoleBasedGroupWarmupReconciler) reconcileUnfinished(ctx context.Context
...
+			return r.waitForTarget(ctx, warmup)
+		}
 		logger.Error(err, "failed to get desired nodes to warm up")
 		return ctrl.Result{}, err
 	}
 
+	if err := validateWarmupActions(desiredNodes); err != nil {
+		logger.Error(err, "invalid warmup actions")
+		return ctrl.Result{}, r.failWarmupJob(ctx, warmup, activePods, succeededPods, failedPods, desiredNodes,
+			"InvalidWarmupSpec", err.Error())
```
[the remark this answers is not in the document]
googs1025: Done — validation now runs before target-node discovery, including zero-node matches.

internal/controller/workloads/rolebasedgroupwarmup_controller.go
googs1025: Done — existing Warmup Pods are listed before validation and are cleaned up on invalid-spec failure, with regression coverage.

internal/controller/workloads/rolebasedgroupwarmup_controller.go
googs1025: Done — cached NotFound now falls back to apiReader before failing, with regression coverage.

api/workloads/v1alpha2/rolebasedgroupwarmup_types.go
```diff
@@ -24,17 +24,19 @@ import (
...
 
 const (
-	WarmupJobPhaseNone      WarmupJobPhase = ""
-	WarmupJobPhaseRunning   WarmupJobPhase = "Running"
-	WarmupJobPhasePaused    WarmupJobPhase = "Paused"
-	WarmupJobPhaseCompleted WarmupJobPhase = "Completed"
-	WarmupJobPhaseFailed    WarmupJobPhase = "Failed"
+	WarmupJobPhaseNone             WarmupJobPhase = ""
+	WarmupJobPhaseRunning          WarmupJobPhase = "Running"
+	WarmupJobPhaseWaitingForTarget WarmupJobPhase = "WaitingForTarget"
```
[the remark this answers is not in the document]
googs1025: Done — WaitingForTarget was removed; the Warmup KEP now documents the final Failed/InvalidTarget behavior.

internal/controller/workloads/rolebasedgroupwarmup_controller.go
```diff
@@ -226,6 +245,57 @@ func (r *RoleBasedGroupWarmupReconciler) reconcileUnfinished(ctx context.Context
...
+		Status:             metav1.ConditionFalse,
+		ObservedGeneration: warmup.Generation,
+		Reason:             "RoleBasedGroupNotFound",
+		Message:            fmt.Sprintf("target RoleBasedGroup %q was not found", warmup.Spec.TargetRoleBasedGroup.Name),
+	}
+	apimeta.SetStatusCondition(&warmup.Status.Conditions, condition)
+	warmup.Status.Phase = workloadsv1alpha2.WarmupJobPhaseWaitingForTarget
+	if err := r.Status().Update(ctx, warmup); err != nil {
+		return ctrl.Result{}, err
+	}
```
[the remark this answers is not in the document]
googs1025: Done — WaitingForTarget was removed in the final design, so this status-update loop no longer exists.

api/workloads/v1alpha2/rolebasedgroupwarmup_types.go
```diff
@@ -35,6 +35,7 @@ type ImagePreloadAction struct {
 	// Images specifies the container images to be preloaded onto target nodes.
 	// Each entry must be a valid image reference (e.g., "registry.example.com/app:v1.0").
 	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:items:MinLength=1
```
diw-zw, line 38: Validation is asymmetric: `minLength: 1` is added only for `imagePreload.images`. `customizedAction.containers[]` gets no equivalent constraint (the CRD still only requires `name`), and `validateWarmupActions` never inspects a customized container's `image`. As a result a customized container with a missing/blank image passes both admission and controller validation and only fails when the Pod is actually built - a path that is currently swallowed silently (see the comment on `validateWarmupActions`). Please validate customized container images as well, or reject them at the CRD layer.

internal/controller/workloads/rolebasedgroupwarmup_controller.go
```diff
@@ -226,6 +237,40 @@ func (r *RoleBasedGroupWarmupReconciler) reconcileUnfinished(ctx context.Context
...
+				if len(action.ImagePreload.Images) == 0 {
+					return fmt.Errorf("node %q has imagePreload with no images", nodeName)
+				}
+				for _, image := range action.ImagePreload.Images {
+					if strings.TrimSpace(image) == "" {
+						return fmt.Errorf("node %q has an empty image reference", nodeName)
+					}
+				}
+			}
+			if action.CustomizedAction != nil && len(action.CustomizedAction.Containers) == 0 {
```
diw-zw: Pod-create failures are swallowed here, leaving the Warmup permanently stuck with an empty phase.

`validateWarmupActions` only checks that the customized container *list* is non-empty; it never checks each container's `image`, and `customizedAction.containers[]` does not require `image` in the CRD. So a customized container without `image` (or with a whitespace image) passes admission + this validation, `buildWarmupPod` produces a container with an empty image, and `createPodsForNodes` then hits `r.Create`. For any error other than `AlreadyExists` it only logs + emits an event and `continue`s - no error return, no requeue, no terminal phase.

Verified on a real cluster (PR image `rbgs-controller:pr-466-1`):
1. Apply a Warmup whose only action is a customized container without an image:
   `spec.targetNodes.nodeNames: [<node>]`, `spec.targetNodes.customizedAction.containers: [{name: c1, command: ["true"]}]`
2. Admission accepts it. Controller logs, once per reconcile:
   `Failed to create warmup Pod for node <node>: Pod "..." is invalid: spec.containers[0].image: Required value`
3. `kubectl get rbgwarmup` shows `PHASE=""` / `desired=1` / `active=0` *forever*; no condition is set and reconcile stops (no requeue). With a whitespace image the message is instead `spec.containers[0].image: Invalid value: " ": must not have leading or trailing whitespace`.
4. Adding `policies.globalTimeoutSeconds: 10` does not help: `status.startTime` is never set (the timeout branch requires it), so the timeout never fires and the object hangs indefinitely.

This contradicts the PR's own motivation ("accepted empty image references and could create invalid Pods"): the invalid Pod is never created, but the Warmup becomes a permanent zombie - `kubectl wait`/monitoring never sees a terminal phase.

Suggested fix: validate each customized container's `image` here, and either propagate the create error out of `createPodsForNodes` (so the request is requeued) or fail the Warmup after N consecutive create failures.

internal/controller/workloads/rolebasedgroupwarmup_controller.go
```diff
@@ -174,8 +175,18 @@ func (r *RoleBasedGroupWarmupReconciler) reconcileUnfinished(ctx context.Context
...
+		logger.Error(err, "invalid warmup specification")
+		return ctrl.Result{}, r.failWarmupJob(ctx, warmup, activePods, succeededPods, failedPods, nil,
+			"InvalidWarmupSpec", err.Error())
+	}
+
 	desiredNodes, err := r.getDesiredNodesToWarmup(ctx, *warmup)
 	if err != nil {
+		if apierrors.IsNotFound(err) && warmup.Spec.TargetRoleBasedGroup != nil {
+			return ctrl.Result{}, r.failWarmupJob(ctx, warmup, activePods, succeededPods, failedPods, nil,
+				"InvalidTarget", err.Error())
```
diw-zw, line 188: A transient or ordering-related NotFound becomes an unrecoverable terminal failure.

The new `apiReader` fallback fixes the "warmup's cached RBG read missed but the RBG already exists" case, but it does not cover the case where the target RBG is created *shortly after* the Warmup - which is normal for `kubectl apply -f` of a combined manifest, or for any GitOps/controller ordering. Since the Warmup is one-shot, never requeued and never retried, the failure is permanent.

Verified on a real cluster with the PR image:
- Warmup created first, target RBG created a few seconds later -> Warmup goes `Failed/InvalidTarget`; creating the RBG afterwards does not recover it.
- A single manifest containing the Warmup before the RBG, applied with one `kubectl apply -f` -> the RBG is created fine but the Warmup stays `Failed/InvalidTarget` permanently.

Please consider a bounded wait/requeue for a missing target (the earlier `WaitingForTarget` approach, keeping the status-update de-duplication from eab9293b so it does not self-trigger) instead of a hard one-shot failure, or otherwise make the failure recoverable.

internal/controller/workloads/rolebasedgroupwarmup_controller.go
```diff
@@ -444,14 +489,22 @@ func (r *RoleBasedGroupWarmupReconciler) getDesiredNodesToWarmup(ctx context.Con
...
+				logger.Error(fallbackErr, "Failed to confirm RoleBasedGroup with API reader", "name", rbgTarget.Name, "namespace", warmup.Namespace)
+				return nil, fmt.Errorf("failed to get RoleBasedGroup %s/%s: %w", warmup.Namespace, rbgTarget.Name, fallbackErr)
+			}
+			reader = r.apiReader
 		}
 
 		// List all Pods belonging to this RoleBasedGroup
 		podList := &corev1.PodList{}
-		if err := r.List(ctx, podList,
+		if err := reader.List(ctx, podList,
```
diw-zw, line 520: An RBG whose Pods are not scheduled yet makes the Warmup report false success.

The loop below skips Pods with `pod.Spec.NodeName == ""`. If the Warmup reconciles before the target RBG's Pods get bound to a node, `desiredNodes` ends up empty, so `updateStatus` sets `phase=Completed` with reason `NoNodesMatched` - i.e. "nothing to warm up". Nothing was pre-pulled, but the job reports success, and being one-shot it can never correct itself.

Verified on a real cluster with the PR image: create an RBG whose Pod cannot be scheduled immediately (e.g. a Pod template with `resources.requests.cpu: "1000"` on a full node) and right after it a Warmup via `targetRoleBasedGroup`:
```
rbg-pending-worker-0   Pending   <no node assigned>
warmup: phase=Completed  desired=0  conditions: Complete/NoNodesMatched
```
Per the KEP workflow, the user sees `Completed` and then triggers the RBG upgrade - but the images were never preloaded, silently defeating the whole feature. Please distinguish "the RBG has no Pods for the listed roles" from "Pods exist but are not scheduled yet", and in the latter case wait/requeue (or record a distinct condition) instead of completing.

internal/controller/workloads/rolebasedgroupwarmup_controller.go
```diff
@@ -226,6 +237,40 @@ func (r *RoleBasedGroupWarmupReconciler) reconcileUnfinished(ctx context.Context
...
+}
+
+func validateWarmupSpec(spec workloadsv1alpha2.RoleBasedGroupWarmupSpec) error {
+	actionsByTarget := make(map[string][]workloadsv1alpha2.WarmupActions)
+	if spec.TargetNodes != nil {
+		actionsByTarget["targetNodes"] = []workloadsv1alpha2.WarmupActions{spec.TargetNodes.WarmupActions}
+	}
+	if spec.TargetRoleBasedGroup != nil {
+		for roleName, actions := range spec.TargetRoleBasedGroup.Roles {
+			actionsByTarget[fmt.Sprintf("role %q", roleName)] = []workloadsv1alpha2.WarmupActions{actions}
```
diw-zw: Minor: these pseudo keys are surfaced as node names in validation messages, e.g. `node "targetNodes" has an empty image reference` or `node "role \"worker\"" has ...`, which is misleading when debugging. In addition, `spec.targetRoleBasedGroup.roles` is a map, so when several roles are invalid the reported role is nondeterministic. Consider phrasing the messages in terms of the target reference and sorting role names for deterministic output.

keps/129-rbg-warmup/README.md
```diff
@@ -176,6 +176,7 @@ At least one of `imagePreload` or `customizedAction` must be specified (enforced
 | `Complete`       | True   | `WarmupCompleted`              | All nodes reached a terminal state and failures are within tolerance.                                                                                                   |
 | `Complete`       | True   | `NoNodesMatched`               | No nodes matched the target configuration; job completed immediately.                                                                                                   |
 | `Failed`         | True   | `GlobalTimeoutExceeded`        | `globalTimeoutSeconds` exceeded; active Pods were deleted.                                                                                                              |
+| `Failed`         | True   | `InvalidTarget`                | The referenced RoleBasedGroup does not exist; the one-shot Warmup must be recreated after fixing the target.                                                            |
```
diw-zw, line 180: Please also document the new `Failed/InvalidWarmupSpec` reason here - this PR emits it for empty/whitespace images and empty customized actions, but the condition table only adds `InvalidTarget`. Also, the "does not retry / must be recreated" semantics for `InvalidTarget` should mention what happens when the target RBG exists but has no scheduled Pods yet, since that currently completes as `NoNodesMatched` (false success).

internal/controller/workloads/rolebasedgroupwarmup_controller_test.go
```diff
@@ -817,3 +821,204 @@ func TestReconcile_RespectsParallelism(t *testing.T) {
 		t.Fatalf("expected 1 pod (parallelism=1), got %d", len(podList.Items))
 	}
 }
+
+func TestReconcile_InvalidImageFailsWarmup(t *testing.T) {
```
diw-zw, line 827: The new tests cover empty image / empty list / empty customized containers, but they miss the two failure modes that actually reproduce on a real cluster:

1. A Warmup whose `customizedAction.containers[].image` is empty or whitespace must still reach a terminal phase and must never create an invalid Pod. Today it hangs with `phase=""` because the Pod-create rejection is swallowed (see the comment on `validateWarmupActions`).
2. A `targetRoleBasedGroup` whose Pods exist but have `spec.nodeName == ""` must not complete as `NoNodesMatched`; today it reports `Completed` while doing nothing.

Adding these two cases as regression tests would have caught both issues.

internal/controller/workloads/rolebasedgroupwarmup_controller.go
googs1025: Skipping an unscheduled Pod is intentional because it does not have a node to warm yet, but I agree that reporting `Completed` can be misleading. This behavior predates this PR; I would like to address readiness and requeue semantics in the same focused follow-up as the missing-target case.

internal/controller/workloads/rolebasedgroupwarmup_controller.go
googs1025: Fixed in `d379f04a`. Customized container images are validated before discovery and Pod creation, and owner-reference or non-`AlreadyExists` create errors are now returned so controller-runtime retries them. Regression coverage includes empty/whitespace images and create failures.

api/workloads/v1alpha2/rolebasedgroupwarmup_types.go
googs1025: Addressed in `d379f04a` and `af6adf03`. The generated CRD now applies an item-scoped CEL rule requiring each customized container to have an image, while controller validation uses `TrimSpace` to catch whitespace and legacy-invalid objects. This prevents invalid Pods without exceeding the CEL cost budget.

internal/controller/workloads/rolebasedgroupwarmup_controller_test.go
googs1025: Added regression coverage for empty and whitespace-only customized container images, including the terminal `InvalidWarmupSpec` status and no invalid Pod creation. The unscheduled RBG Pod case is intentionally deferred with the target-readiness discussion.

internal/controller/workloads/rolebasedgroupwarmup_controller.go
googs1025: Fixed in `d379f04a`. Validation errors now use explicit spec paths such as `spec.targetNodes` and `spec.targetRoleBasedGroup.roles["worker"]`, and role names are sorted before validation for deterministic output.

keps/129-rbg-warmup/README.md
googs1025: The KEP now documents `Failed/InvalidWarmupSpec` and its cleanup behavior. I have left the unscheduled-target contract undocumented for now because that readiness behavior is being discussed separately below.

internal/controller/workloads/rolebasedgroupwarmup_controller.go
googs1025: This is a valid ordering case. I am keeping the current one-shot `InvalidTarget` behavior in this focused validation fix and would prefer to handle bounded waiting and recovery in a dedicated follow-up, since that changes the public state machine and timeout semantics.

diw-zw:
## Verified against the new head `af6adf03`

I re-ran my cluster/API-server repros against `d379f04a` + `af6adf03`. Results:

**Resolved**
- Fixed-Pod-create-error (my earlier comment): create/owner-reference errors are now returned, so controller-runtime retries and the Warmup no longer sits at `phase=""`. Confirmed by `TestReconcile_PropagatesPodCreateError`, `TestReconcile_PropagatesOwnerReferenceError`.
- Customized container image validation: I installed the generated CRD into a real API server (envtest) and confirmed item-scoped CEL now rejects a missing/blank image in both `targetNodes` and `targetRoleBasedGroup.roles`. Controller `TrimSpace` validation catches whitespace.
- Validation message paths / deterministic role ordering: confirmed (`spec.targetNodes`, `spec.targetRoleBasedGroup.roles["..."]`, sorted).
- KEP: `Failed/InvalidWarmupSpec` now documented.

**Still open**
- Unscheduled-target false success, and the one-shot `InvalidTarget` ordering case: acknowledged as a follow-up. I re-ran those repros at this head and they still reproduce - see the new comments I added below.

One new risk introduced by the CRD CEL rule is described in a separate comment.

api/workloads/v1alpha2/rolebasedgroupwarmup_types.go
```diff
@@ -50,6 +51,7 @@ type CustomizedAction struct {
 	// image preload containers. The warmup is considered complete when all containers
 	// finish successfully.
 	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:items:XValidation:rule="has(self.image) && self.image != ''",message="customized action container image must not be empty"
```
diw-zw: ### Upgrade hazard: this item-scoped CEL rule can soft-freeze pre-existing objects

Verified on a real API server (envtest): I installed the CRD **at the base commit `41e40cb0`**, created a Warmup that the old schema accepted (a `customizedAction` container without `image`), then upgraded the CRD in place to this PR's version, and exercised updates:

| operation on the pre-existing object | k8s 1.31 / 1.32 API server | k8s 1.33 / 1.34 API server |
|---|---|---|
| GET (read) | accepted | accepted |
| `PATCH .../status` (merge) | **REJECTED** | accepted |
| `PUT .../status` (what `Status().Update` does) | **REJECTED** | accepted |
| spec patch that fixes `image` | accepted | accepted |
| DELETE | accepted | accepted |

Rejection message:
```
RoleBasedGroupWarmup "legacy" is invalid:
spec.targetNodes.customizedAction.containers[0]: Invalid value: "object":
customized action container image must not be empty
```

Why this matters: an existing Warmup whose spec is invalid under the new rule can no longer have its **status** written. The controller would fail to transition `phase`/`conditions`, so `kubectl wait` never returns and any repair logic that reads `status` cannot make progress. The object is only recoverable by deleting it or by patching the spec (which an operator would not know to do, since they cannot see the condition).

Note the repo's own CI targets **k8s 1.31** (`ENVTEST_K8S_VERSION: "1.31.0"`, `KUBERNETES_VERSION: v1.31.12`), which is exactly the affected range, while RBG 0.9.0 also runs on ACK clusters at 1.36. Two things to fix:

1. The rule rejects `image == ""` but **accepts whitespace** (`image: " "` and `image: "\t"` are both accepted by admission; only the controller's `TrimSpace` catches them). So the intended invariant is only partially enforced at the API layer, and the controller still has to be the backstop.
2. Consider `has(self.image) && !self.image.trim().isEmpty()`? CEL here has no `trim()`, so the practical options are: keep the item rule but relax it (e.g. apply it only on create via `oldSelf`-based rules) or drop the CEL rule and rely on the controller validation plus a documented `InvalidWarmupSpec` terminal state - so pre-existing objects keep a writable status.

Any of those choices is fine; the point is that a schema-only rule applied to already-stored objects is not free on 1.31/1.32.

internal/controller/workloads/rolebasedgroupwarmup_controller.go
```diff
@@ -444,14 +502,22 @@ func (r *RoleBasedGroupWarmupReconciler) getDesiredNodesToWarmup(ctx context.Con
...
+				logger.Error(fallbackErr, "Failed to confirm RoleBasedGroup with API reader", "name", rbgTarget.Name, "namespace", warmup.Namespace)
+				return nil, fmt.Errorf("failed to get RoleBasedGroup %s/%s: %w", warmup.Namespace, rbgTarget.Name, fallbackErr)
+			}
+			reader = r.apiReader
 		}
 
 		// List all Pods belonging to this RoleBasedGroup
 		podList := &corev1.PodList{}
-		if err := r.List(ctx, podList,
+		if err := reader.List(ctx, podList,
```
diw-zw, line 520: ### (still open) Unscheduled target Pods are silently reported as success - re-verified at `af6adf03`

Thanks for the reply. I re-ran the repro at this head; the behaviour is unchanged, so I am leaving the concern on the record.

**Cluster repro (PR image `rbgs-controller:pr-466-1`)**
1. Create a target RBG whose Pod cannot be scheduled immediately, e.g. a template with `resources.requests.cpu: "1000"` on a full node.
2. While its Pod is still `Pending`, create a Warmup with `targetRoleBasedGroup` pointing at that RBG.
3. Observe:

```
rbg-pending-worker-0   Pending   <no node assigned>
warmup:  phase=Completed   desired=0   conditions: Complete/NoNodesMatched
```

**Unit repro at this head** (same loop, a `Pod` with `spec.nodeName == ""` and a `targetRoleBasedGroup` spec) prints:

```
RESULT phase="Completed" desired=0 active=0 requeueAfter=0s
  condition Complete/NoNodesMatched: No nodes matched the target configuration
*** FALSE SUCCESS: warmup reports Completed while nothing was warmed up ***
```

I understand skipping an unscheduled Pod is intentional (nothing to warm yet), and you are deferring the readiness/requeue semantics. The reason I would still like it handled in this PR is that the failure mode is a **false positive that the user cannot detect**: the KEP workflow is "wait for `Completed`, then trigger the RBG upgrade", so the user proceeds believing images are pre-warmed when nothing was pre-warmed, and - being one-shot - the resource can never correct itself. If it is deferred, could the KEP at minimum state that `Completed/NoNodesMatched` does not guarantee any node was warmed, and that `targetRoleBasedGroup` requires all listed-role Pods to be scheduled before creating the Warmup?

internal/controller/workloads/rolebasedgroupwarmup_controller.go
```diff
@@ -174,8 +175,18 @@ func (r *RoleBasedGroupWarmupReconciler) reconcileUnfinished(ctx context.Context
...
 
+	if err := validateWarmupSpec(warmup.Spec); err != nil {
+		logger.Error(err, "invalid warmup specification")
+		return ctrl.Result{}, r.failWarmupJob(ctx, warmup, activePods, succeededPods, failedPods, nil,
+			"InvalidWarmupSpec", err.Error())
+	}
+
 	desiredNodes, err := r.getDesiredNodesToWarmup(ctx, *warmup)
 	if err != nil {
+		if apierrors.IsNotFound(err) && warmup.Spec.TargetRoleBasedGroup != nil {
```
diw-zw, line 186: ### (still open) `InvalidTarget` is unrecoverable - re-verified at `af6adf03`

Agreed that bounded waiting is a larger change; recording the re-verification so the impact is explicit.

**Unit repro at this head**
1. Create only the Warmup (`targetRoleBasedGroup: later-rbg`).
2. Reconcile -> `phase="Failed"`, `requeueAfter=0s`, condition `Failed/InvalidTarget`.
3. Now create the target RBG and reconcile again -> the Warmup is **still** `phase="Failed"`; a terminal phase short-circuits `reconcileFinished`, so it never picks up the target.

Real cluster, same result, including when the Warmup and the RBG come from a single `kubectl apply -f` manifest (RBG applied second): the RBG is created fine, the Warmup stays `Failed/InvalidTarget` permanently.

This is asymmetric with what this PR just fixed: the cached-`NotFound`/`apiReader` fallback handles "RBG already exists but the cache missed", but not "RBG is created moments later", which is arguably the more common ordering in GitOps. Since the follow-up will revisit the state machine anyway, would it be cheaper to keep the failure recoverable now (e.g. requeue instead of terminating, or at least honour a retry if the target appears), and leave the *timeout/expiry* semantics to the follow-up?

keps/129-rbg-warmup/README.md
```diff
@@ -176,6 +176,8 @@ At least one of `imagePreload` or `customizedAction` must be specified (enforced
 | `Complete`       | True   | `WarmupCompleted`              | All nodes reached a terminal state and failures are within tolerance.                                                                                                   |
 | `Complete`       | True   | `NoNodesMatched`               | No nodes matched the target configuration; job completed immediately.                                                                                                   |
 | `Failed`         | True   | `GlobalTimeoutExceeded`        | `globalTimeoutSeconds` exceeded; active Pods were deleted.                                                                                                              |
+| `Failed`         | True   | `InvalidWarmupSpec`            | A warmup action is invalid, such as an empty image or an empty customized action; active Pods are deleted and no new Pods are created.                                  |
+| `Failed`         | True   | `InvalidTarget`                | The referenced RoleBasedGroup does not exist; the one-shot Warmup must be recreated after fixing the target.                                                            |
```
diw-zw, line 180: Thanks for documenting `InvalidWarmupSpec`. Two follow-ups for the condition table:

1. `InvalidWarmupSpec` currently applies to pre-existing objects too, and on k8s 1.31/1.32 such an object can no longer have its **status** written (see the comment on `rolebasedgroupwarmup_types.go:54`), so the condition may never actually be recorded for an already-running Warmup. Worth a note about legacy objects if the CEL rule stays.
2. The table still does not state that `Complete/NoNodesMatched` does not imply any node was warmed - which is exactly what happens when the target RBG exists but its Pods are not scheduled yet.

diw-zw (approved):

PR #469, opened 2026-09-14
test(upgrade): assert expected post-upgrade content instead of counting rewrites

diw-zw:
## What

Replace the upgrade suite's generation-bump bookkeeping with expected-content
assertions: each recorded rewrite is a function applied to the before snapshot
to yield the exact expected after spec, compared with cmp.Diff. The heal of the
legacy "Recreate" strategy spelling (#463) is encoded as expected content on the
RoleBasedGroupSet child and its RoleInstanceSets.

## Why

Generation counts could not distinguish content rewrites from writes that only
advance generation, required fragile per-controller-start scaling, and reported
an object name without the field that changed. The expected/after comparison is
exact in both directions, idempotent across controller restarts, and fails with
a diff that names the field — including a stale recording when a registered
rewrite no longer happens.


test/e2e/upgrade/specs.go
```diff
@@ -472,8 +476,8 @@ func RunUpgradeSpecs(f *framework.Framework) {
 			gomega.Expect(child.Spec.Roles[0].RolloutStrategy).ToNot(gomega.BeNil())
 			gomega.Expect(child.Spec.Roles[0].RolloutStrategy.RollingUpdate).ToNot(gomega.BeNil())
 			gomega.Expect(child.Spec.Roles[0].RolloutStrategy.RollingUpdate.Type).To(
-				gomega.Equal(workloadsv1alpha2.LegacyRecreateUpdateStrategyType),
-				"the stored legacy strategy type in the child RoleBasedGroup was rewritten",
+				gomega.Equal(workloadsv1alpha2.RecreatePodUpdateStrategyType),
```
cheyang, line 479: This asserts the legacy-set child's stored `rolloutStrategy.rollingUpdate.type` is healed to `RecreatePod` on upgrade, but that heal does not happen via the mechanism the PR cites. `updateExistingRBGs` is only reached for children where `needsUpdate` is true, and `needsUpdate` delegates to `rolesEqual`, which (#463's convergence fix) deep-copies and normalizes **both** the parent template and the child before `reflect.DeepEqual`. The child was written by v0.7.0 with `Recreate` copied verbatim from a legacy template, and the parent template is also `Recreate`, so after normalization they compare equal → `needsUpdate` returns false → the child is never re-applied → its stored value stays `Recreate`. This also regresses the prior assertion (`== LegacyRecreateUpdateStrategyType`, "leaves the child alone"), which matched the real behavior.

I proved this at the unit level on a checkout of this PR's base: `needsUpdate(parent=Recreate, child=Recreate)` returns false (the existing suite's "legacy strategy spelling differs only" case is the same gate). `normalizedGroupTemplateRoles` does yield `RecreatePod`, so the barrier is the closed `needsUpdate` gate, not a missing normalize.

Either revert this assertion to `== LegacyRecreateUpdateStrategyType` (the child is not healed on upgrade), or — if healing the existing legacy child on upgrade is actually desired — make that a separate behavior change in the reconciler (it's out of scope for a tests-only PR, and re-opening `needsUpdate` for legacy spellings risks the B1 hot-loop #463 fixed).

test/e2e/upgrade/snapshot.go
```diff
@@ -936,22 +922,33 @@ var controllerStartRewrites = recordedRewrites{
...
-		"RoleInstanceSet/" + legacySetChildRISName(): 1,
+		// also wrote with the legacy spelling copied verbatim from the template, so
+		// it is healed on the upgraded controller's first reconcile too.
+		"RoleInstanceSet/" + legacySetChildRISName(): healStoredStrategyType,
+		// The same heal reaches the child RoleBasedGroup itself: the RBGS controller
+		// re-applies the child from its groupTemplate with the strategy type
+		// normalized, so the stored child spec changes on its first reconcile. A
+		// top-level RoleBasedGroup has no writer above it, so up-legacy keeps the
+		// legacy spelling -- only the child is re-applied.
+		"RoleBasedGroup/" + legacySetChildName(): healRoleStrategyTypes,
```
cheyang, line 941: The recorded `healRoleStrategyTypes` for `RoleBasedGroup/up-legacy-set-0` asserts the child's after-spec is `RecreatePod`. Since the child is not re-applied on upgrade (see the phase-3 comment), `compareWithExpected(before=Recreate, heal→RecreatePod, after=Recreate)` produces a non-empty `cmp.Diff` and `checkOwnersStable` flags the child as a finding — so this recording makes the upgrade spec fail rather than pass. Drop this entry (the child stays `Recreate`).

test/e2e/upgrade/snapshot.go
```diff
@@ -936,22 +922,33 @@ var controllerStartRewrites = recordedRewrites{
...
+	},
+
+	// Healing the child's stored spec moved the RBG-layer revision hash, so the
+	// controller stamps one new revision for it. The hash moves but no pod does:
+	// the RoleInstanceSet layer is repaired in place and its revision is stable,
+	// which the RoleInstanceSet entries above assert through content. No revision
+	// may be removed, and any count other than one is still reported with its
+	// content diff.
+	revisionAdds: map[string]int{
+		legacySetChildName(): 1,
```
cheyang, line 951: `revisionAdds[up-legacy-set-0] = 1` is contingent on the child's spec being healed (that heal is what moves the RBG-layer revision hash). Since the child is not re-applied, no new revision is stamped and this entry is stale — the "recorded new ControllerRevision(s) never appeared" branch of `checkNoRevisionExplosion` fires. Drop this entry too.

cheyang (dismissed):
Requesting changes. The headline of this PR — encoding the #463 legacy-strategy heal as expected post-upgrade content — is correct for the RoleInstanceSet layer (the webhook heals stored RISes on write) but does not hold for the legacy-set **child RoleBasedGroup**, and the suite records the opposite.

The refuted premise: phase 3 asserts the legacy-set child is healed `Recreate → RecreatePod` on the upgraded controller's first reconcile, and `snapshot.go` records `healRoleStrategyTypes` + `revisionAdds=1` for it, citing the RBGS controller re-applying the child via `updateExistingRBGs`. But `updateExistingRBGs` is gated by `needsUpdate`, and `needsUpdate → rolesEqual` normalizes **both** parent and child before `DeepEqual` (#463's convergence fix). The legacy child (stored `Recreate`, as v0.7.0 wrote it) under a legacy parent (`Recreate`) normalizes equal, so `needsUpdate` returns false, the child is never re-applied, and its stored type stays `Recreate`.

This is the same gate #463 added on purpose — to stop re-issuing child updates for spelling-only deltas — so the two are in tension: the fix that converges the post-heal steady state also prevents the initial re-apply that would heal the child.

I proved this at the unit level on the PR's base (reconciler logic is tests-only-identical at base and head): `needsUpdate(parent=Recreate, child=Recreate)` returns false, while `normalizedGroupTemplateRoles` does yield `RecreatePod` — so the cause is the closed gate, not a missing normalize. `newRBGForSet` and a direct `updateExistingRBGs` call do normalize (the existing tests prove it), but the reconcile loop only calls `updateExistingRBGs` for `needsUpdate=true` children, which the legacy child is not. That's the gap.

Net effect: the recorded `healRoleStrategyTypes` would make `checkOwnersStable` flag the child, phase 3's `child==RecreatePod` fails (and regresses the prior correct `==Recreate` assertion), and `revisionAdds=1` is stale. Since `RunUpgradeSpecs` only runs via `/release-test` against published images — not in per-PR CI, not against the PR head — these wrong expectations wouldn't be caught at PR time and would surface as a red release gate later.

The generation→content migration itself is sound and the pure functions (`compareWithExpected`, `healStoredStrategyType`, `describeAddedRevisionChange`, `revisionDataDiff`) are correct; the set-template-stays-`Recreate` assertion is right; dropping the per-start LeaderWorkerSet allowance is right (that patch is content-equal). Only the recorded child-heal is wrong.

Suggested fix: revert the phase-3 child assertion to `== LegacyRecreateUpdateStrategyType` ("leaves the child alone") and drop the `healRoleStrategyTypes` + `revisionAdds` entries for `RoleBasedGroup/up-legacy-set-0`. Evidence and the unit harness are on branch `verify/pr469-upgrade-content-assert` under `docs/verification/pr469-upgrade-content-assert/`.

cheyang (approved):
# Approving — prior REQUEST_CHANGES review was wrong (retracted)

I posted a REQUEST_CHANGES on this PR claiming the legacy-set child `up-legacy-set-0` is **not** healed on upgrade and that phase-3 (`child == RecreatePod`), the recorded `healRoleStrategyTypes`, and `revisionAdds=1` were all wrong. **That was incorrect, and I've dismissed that review.** A live ACK upgrade e2e disproves all three findings.

## Live evidence (2026-09-16)

v0.7.0 → upgraded controller (built from this PR's base `c500c963`, i.e. the production code the PR's tests assume; image `ttl.sh/cheyang-rbgs-pr469:pr469c`), on the `up-legacy-set` fixture:

| object | before | after | |
| --- | --- | --- | --- |
| child RBG `up-legacy-set-0` `rollingUpdate.type` | `Recreate` | **`RecreatePod`** | healed |
| child generation | 1 | **2** | re-applied |
| ControllerRevisions | `…-0-66f859d984-1` | `…-0-66f859d984-1`, **`…-0-6bfc548975-2`** | +1 |
| RBGSet template `type` | `Recreate` | `Recreate` | untouched ✓ |

→ `specs.go:479` (`child == RecreatePod`) **passes**, `snapshot.go:941` `healRoleStrategyTypes` is **correct**, `snapshot.go:951` `revisionAdds=1` is **correct**.

## Why my unit-level refutation was wrong

My harness (`pr469_legacy_child_heal_premise_test.go`) built the parent template role and child role with only `rollingUpdate.type` set, so after `rolesEqual` normalizes the `type` field on both sides, `reflect.DeepEqual` held → `needsUpdate=false` → "child never re-applied". **But the real v0.7.0-written child carries `maxSurge:0 / maxUnavailable:1 / partition:0`** (v0.7.0 defaulting populates them), while the template role has those fields **nil**. `rolesEqual`'s `normalizeRolloutUpdateTypes` only normalizes `type` — it does not touch `maxSurge`/`maxUnavailable`/`partition` — so the two sides are not `DeepEqual` → `needsUpdate=true` → `updateExistingRBGs` re-applies `normalizedGroupTemplateRoles(rbgset)` → the mutating `RoleBasedGroupDefaulter` webhook heals `Recreate → RecreatePod`. My unit test's input was unrepresentative of real cluster state; the live result is the gold standard and it's unambiguous.

## Note on the full `make test-e2e-upgrade`

The full suite couldn't complete on the ACK cluster I used: it fails in phase 1 on the `up-lws` fixture because the LeaderWorkerSet CRD (`leaderworkerset.x-k8s.io`) isn't installed there (`install.go` lists it as a cluster prereq the suite does not provide). `--ginkgo.fail-fast` stopped before phase 3, so the decisive observation came from a targeted manual repro of the legacy-set premise (same v0.7.0 install, same fixture, same upgraded image) rather than the full suite run. The premise under test here — the legacy-set child heal — has no LWS dependency, so the repro is a valid substitute.

The cluster was restored to its prior v0.8.0 install afterward.

Apologies for the noise. The generation→content migration and the pure functions (`compareWithExpected`, `healStoredStrategyType`, `describeAddedRevisionChange`, `revisionDataDiff`) are sound, and the child-heal expectations are correct. Approving.

PR #470, opened 2026-09-15
fix: retry unhealthy update targets in Parallel rollouts

### LINKED ISSUE #464: [Bug] Stateful RoleInstanceSet rollout can stall after fresh unhealthy instances exhaust maxUnavailable
### Checklist

### Describe the bug

A Stateful RoleInstanceSet rollout may stop progressing when its Pod template is updated again while the Pods from the previous update are still NotReady. The rollout does not automatically resume unless another event triggers reconciliation.

This can be fixed by scheduling a requeue when the unhealthy window expires. It is also worth reconsidering whether the additional fixed 10-second window is necessary, since probe `failureThreshold` and `periodSeconds` already provide failure debouncing.

### Reproduction

  Create a Stateful RoleInstanceSet with 2 replicas, maxSurge: 0, and maxUnavailable: 2.

  1. Add health probes.
  2. Before Pods become Ready, remove the probes.
  3. The Pods may retain the probes and the rollout stalls until another event triggers reconciliation.

### Environment

  - RBG: v0.8.0
  - Workload mode: Stateful RoleInstanceSet
  - Update strategy: InPlaceIfPossible

jiangwu25:
### Ⅰ. Motivation

In Parallel mode, a rollout can stop when unhealthy instances at the old revision use up the `maxUnavailable` budget. These update targets can be replaced after ten seconds, but the controller does not schedule a check when that wait ends. It needs another event before it can continue.

### Ⅱ. Modifications

Schedule a retry when an unhealthy update target reaches the end of its ten-second wait. Keep any earlier retry that is already set.

If the budget blocks a healthy target, check the lower targets too. An unhealthy lower target can then request a retry or update once its wait is over.

The health window and availability limit stay the same.

Related to #464

The tests cover these cases:

- An unhealthy update target waits until its health window ends, then the Parallel rollout resumes.
- Repeated checks keep the same deadline and preserve any earlier retry.
- A healthy higher target does not block a lower unhealthy target.
- Recovery clears the old timer. A new unhealthy period starts a new timer.
- An unhealthy instance at the new revision or below `partition` still counts against the budget after ten seconds. The controller must not delete a healthy target when that would exceed `maxUnavailable`.

The retry test fails without the fix.

These checks passed with Go 1.26.6:

```sh
make GOFLAGS=-mod=vendor fmt vet
go test -mod=vendor -count=1 ./api/... ./cmd/... ./internal/... ./pkg/...
go test -mod=vendor -race -count=1 ./pkg/reconciler/roleinstanceset/... ./pkg/utils/requeueduration
go build -mod=vendor -o /tmp/rbg470-rbgs ./cmd/rbgs
```

The changed package also passed golangci-lint v2.13.1:

```sh
GOFLAGS=-mod=vendor golangci-lint run ./pkg/reconciler/roleinstanceset/statefulmode/...
```

This PR covers retries for unhealthy update targets in Parallel mode. Three other cases remain open in #464:

- A NotReady instance at the new revision or below `partition` uses the budget, but is not an update target. A timer alone cannot release that budget.
- OrderedReady stops at the readiness check before it reaches this retry code. A timer alone does not get past that check.
- An early A → B → A rollback can skip the update phase because the current and desired revisions match.

The regression test sets timestamps, so it does not need a ten-second sleep. No API or configuration option is added.

## Checklist

- [ ] Update the documentation related to the change. (No API or configuration change.)


pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control.go
```diff
@@ -615,6 +615,14 @@ func (ssc *defaultStatefulInstanceSetControl) progressUpdate(
 		isFree := isSurgeSlot || isTerminating(target) || isStablyUnhealthy(target)
 
 		if !isFree && initialBaseUnavail+newlyUnavail >= effectiveBudget {
+			if observed, ok := instanceUnhealthySince.Load(target.UID); ok {
+				if first, ok := observed.(time.Time); ok {
+					// Retry when this target can pass the health window without
+					// an unrelated instance event. The window may have just expired.
+					wait := max(time.Until(first.Add(stableUnhealthyDuration)), time.Nanosecond)
```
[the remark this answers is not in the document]
JiangLLM: good call, updated!

NoobDream2568 (commented):
## Review: retry logic is correct now — but the A→B→A path of #464 remains open

**Verdict: The fix is correct and well-tested. One scope note: #464's repro can land on a second, pre-existing stall path that this PR does not fix — better track it in a follow-up issue before #464 is closed.**

### Verified on 423b4118

- `go vet` clean; full `pkg/reconciler/roleinstanceset/...` tree passes (12 `TestProgressUpdateBudget` subtests included).
- `return → continue` is behavior-preserving for all pre-existing scenarios: blocked targets are only ever followed by free ones, and free targets never consume base availability, so deferring a healthy blocked target cannot violate `maxUnavailable`.
- Retry timing is correct with multiple blocked targets: each pushes its own remaining window and `DurationStore.Update` keeps the minimum (`pkg/utils/requeueduration/duration.go:63-66`), so the wake-up lands on the earliest budget consumer's deadline. It also correctly skips non-target consumers (already at `updateRevision`, or below partition), which `progressUpdate` cannot act on anyway.
- The new budget test case and the inverted assertion in the retry test (retry tracks the consumer's ~4s window instead of `wait == 0`) cover exactly the residual scenario. Nice turnaround.

### The A→B→A variant of #464's repro is NOT covered (pre-existing, non-blocking)

The repro — add probes, then remove them before Pods are Ready — is template-wise an A→B→A rollback. Depending on rollback timing it lands on one of two distinct stall paths:

| Rollback timing | State | Path | Fixed here |
|---|---|---|---|
| Late (status already advanced to B), or template changed to C≠A | `currentRevision != updateRevision` | Budget exhaustion + missing requeue | ✅ |
| Early (status never advanced to B) | `currentRevision == updateRevision` | `inRollout` misdetermination → deadlock | ❌ |

For the second path: `computeTopology` sets `inRollout = (currentRevision != updateRevision) && !paused` = **false** (`stateful_instance_set_utils.go:541`), so `progressUpdate` — including this PR's retry logic — never runs, and no other phase recycles the stale instance. No event can unstick it.

I reproduced it on 423b4118 with a unit test mirroring the issue's steps: `current=update=revA`, one instance at revB (healthy) → 3 reconciles, zero deletions, `updatedReplicas` pinned at 1, status fully green — matching the issue's "The Pods may retain the probes".

This predates the PR and shouldn't block merging. But since `fixes #464` will auto-close the issue, please either:

1. Open a follow-up issue for the `inRollout` gap (fix direction: treat stale-revision, non-terminating instances in `[partition, replicas)` as rollout evidence, i.e. `inRollout = (currentRev != updateRev || hasStaleInRollingRange) && !paused`);
2. Downgrade the PR body to `relate to #464`;
3. Fix it in this PR.

cc @cheyang

JiangLLM:
@NoobDream2568. Thanks for pointing this out - ~~I fixed it in 2efff36 (still this PR) and added tests for the early rollback case.~~

Update: I reverted the early rollback change, so this PR now only covers the retry fix. @cheyang  thanks for testing

JiangLLM:
I reverted the early rollback change because it caused a scale-down issue. This PR now focuses on the retry fix. I also changed the body to `Related to #464`, so the early rollback case stays open.

pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_utils.go
```diff
@@ -538,21 +555,23 @@ func computeTopology(
 	}
 	t.partition = partition
 
-	t.inRollout = currentRevision != updateRevision && !set.Spec.UpdateStrategy.Paused
+	// An early rollback can leave stale base instances even when both revisions match.
+	updatePending := currentRevision != updateRevision || hasStaleBaseInstance(instances, partition, t.replicas, updateRevision)
+	existingSurge := min(countExistingValidSurge(instances, t.replicas, maxSurge, updateRevision), maxSurge)
+	baseComplete := allBaseAtUpdateRevHealthy(instances, partition, t.replicas, updateRevision)
+	// Keep surge until the replacement base is ready, even after its revision changes.
+	updatePending = updatePending || (existingSurge > 0 && !baseComplete)
```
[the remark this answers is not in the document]
JiangLLM: Reverted in b0931fa. The surge logic is unchanged in the current diff.

JiangLLM:
Thanks, @NoobDream2568, for the review, and thanks, @cheyang, for the test results.

I reverted the separate rollback change, so this PR now contains only the retry fix from 423b411. All CI checks pass. The early A → B → A rollback case remains in #464.

Could you please take one more look and let me know if this is ready to merge or if you want any further changes?

pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control.go
```diff
@@ -615,11 +615,20 @@ func (ssc *defaultStatefulInstanceSetControl) progressUpdate(
 		isFree := isSurgeSlot || isTerminating(target) || isStablyUnhealthy(target)
 
 		if !isFree && initialBaseUnavail+newlyUnavail >= effectiveBudget {
+			if observed, ok := instanceUnhealthySince.Load(target.UID); ok {
```
diw-zw, line 618: **Budget can be exhausted by an instance that is not a target — and then no retry is scheduled.**

The timer read here is the *target's own* timer (`instanceUnhealthySince.Load(target.UID)`). The budget, however, is consumed by any not-yet-available base in `[startOrdinal, replicas)` (see `collectBaseUnavailable`). When the budget consumer is not an update target, every target is healthy, so each target's entry has already been cleared by `observeInstanceHealth`, this `Load` misses, and no `durationStore.Push` happens — the rollout stalls on an unrelated event.

Two readily reachable shapes:

- An instance already at `updateRevision` that is NotReady (e.g. it crashes after an in-place update). It consumes budget but is excluded from `buildUpdateTargets` (`getInstanceRevision(...) == updateRev` → skipped), so it is never inspected here.
- A base below `partition` that is NotReady. It consumes budget (`collectBaseUnavailable` scans `[startOrdinal, replicas)`) but is below `topo.partition`, so it is not a target either.

Reproduced locally against `b0931fa3` with `replicas=2, maxSurge=0, maxUnavailable=1`: ord0 old-rev healthy (the only target), ord1 already at `updateRev` but NotReady with its timer aged past `stableUnhealthyDuration`. Three consecutive reconciles give `deleted=[] requeue=0s`. The `partition=1` variant (unhealthy base below partition) behaves the same.

Suggestion: drive the retry from the actual budget consumers (the set returned by `collectBaseUnavailable`) instead of targets only, or additionally push a retry for any *non-target* stably-unhealthy base. A regression case with a non-target budget consumer plus a healthy target would cover this.

pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control.go
```diff
@@ -615,11 +615,20 @@ func (ssc *defaultStatefulInstanceSetControl) progressUpdate(
...
+				if first, ok := observed.(time.Time); ok {
+					// Retry when this target can pass the health window without
+					// an unrelated instance event. The window may have just expired.
+					wait := max(time.Until(first.Add(stableUnhealthyDuration)), time.Nanosecond)
+					durationStore.Push(getInstanceSetKey(set), wait)
```
diw-zw, line 623: **This retry is unreachable when `podManagementPolicy: OrderedReady`.**

`progressUpdate` is only called from Phase C (`if topo.inRollout {` at line 487 of this file). Phase B reaches it via `runForAllWithBreak(replicas, processReplicaFn, monotonic)` at line 474, which is called with `monotonic := !allowsBurst(set)` (line 409 → `stateful_instance_set_helper.go:264`, i.e. `true` unless the policy is Parallel). Under `monotonic`, `processReplica` returns `true` for any instance that is not healthy (line 922: `return monotonic, false, nil`), and `runForAllWithBreak` immediately `return`s (helper.go:331-336).

The precondition for this fix — an unhealthy instance holding budget — is exactly what makes Phase B return early, so this `Push` never executes and no timed retry is scheduled. The rollout still depends on an unrelated event, which is the bug this PR sets out to fix.

Reproduced against `b0931fa3` with the same fixture as `TestUpdateStatefulInstanceSetRetriesUnhealthyRollout` but `PodManagementPolicy: OrderedReady` (replicas=2, maxSurge=0, maxUnavailable=2, both instances NotReady):

```
Parallel,    both NotReady -> requeue=9.99s   (fix works)
OrderedReady, both NotReady -> requeue=0s      (fix does not run)
```

OrderedReady is not the CRD default (that is Parallel), but it is a supported, documented policy, so the PR's "the rollout resumes on its own" claim does not hold there. Consider scheduling the retry where the unhealthy evidence is first observed (e.g. alongside `observeInstanceHealth`, or on the Phase B early-return path) so it applies regardless of policy, and add an OrderedReady regression case.

pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control.go
```diff
@@ -615,11 +615,20 @@ func (ssc *defaultStatefulInstanceSetControl) progressUpdate(
 		isFree := isSurgeSlot || isTerminating(target) || isStablyUnhealthy(target)
 
 		if !isFree && initialBaseUnavail+newlyUnavail >= effectiveBudget {
+			if observed, ok := instanceUnhealthySince.Load(target.UID); ok {
+				if first, ok := observed.(time.Time); ok {
+					// Retry when this target can pass the health window without
+					// an unrelated instance event. The window may have just expired.
+					wait := max(time.Until(first.Add(stableUnhealthyDuration)), time.Nanosecond)
```
diw-zw, line 622: Minor / optional: the `time.Nanosecond` floor means that when the window has already expired at this point, we push a 1ns delay, which turns into a near-immediate re-reconcile that only then performs the free-path deletion. It is correct (just one extra pass), but a small floor such as `time.Second` — or simply skipping the push when `time.Until(...) <= 0` and letting the current pass re-evaluate — would avoid the busy reconcile. Given `DurationStore` keeps the minimum across pushes, this also sets the shortest possible requeue for the whole set if any other waiter is present.

diw-zw (commented):
## Review: the `return` → `continue` fix is correct; two residual retry gaps remain

Reviewed at `b0931fa3` (diff is 2 files, +136/−1). The change itself is sound and I agree with the existing threads.

**Verified**
- `return` → `continue` is behavior-preserving: a blocked target can only be followed by free targets, and free updates never consume base availability, so deferring the blocked (healthy) target cannot violate `maxUnavailable`.
- Multi-target timing is right: each push is the remaining window and `DurationStore.Update` keeps the minimum (`pkg/utils/requeueduration/duration.go:63-66`).
- The PR-body claim holds: reverting this file to `main` makes both new tests fail (`TestProgressUpdateBudget/healthy_higher_ordinal_does_not_block_a_stably_unhealthy_target` and `TestUpdateStatefulInstanceSetRetriesUnhealthyRollout`).
- `go vet`, `go build ./cmd/rbgs`, `go test ./api/... ./internal/... ./pkg/reconciler/roleinstanceset/...`, `-race -count=5`, and `-shuffle=on` are all clean; `golangci-lint` reports nothing on the changed package.
- The reverted `utils.go` surge change is genuinely gone: `utils.go` is identical to `main`, so the scale-down misclassification raised by Copilot no longer applies.

**Remaining issues (two new inline comments)**

1. `control.go:618` — the retry is keyed off the *target's* timer, but budget is consumed by any not-yet-available base. When the consumer is not a target (already at `updateRevision` but NotReady, or a base below `partition`), no target has a timer, so nothing is scheduled and the rollout still stalls. Reproduced with `replicas=2, maxSurge=0, maxUnavailable=1` (ord0 healthy target, ord1 at updateRev but NotReady): three reconciles → `deleted=[] requeue=0s`.

2. `control.go:623` — the retry lives in Phase C, which OrderedReady never reaches. With `monotonic=true` (`allowsBurst` is false unless Parallel), Phase B returns early at line 474/922 as soon as an instance is unhealthy — exactly this fix's precondition. Reproduced: `OrderedReady, both NotReady -> requeue=0s` vs `Parallel -> 9.99s`.

Both are reachable through the same failure mode described in #464, just via consumers/policies the new tests do not exercise. They also are not variants of the early A→B→A `inRollout` gap already noted by @NoobDream2568 — that is a third, separate entry point.

**Suggestion:** if neither gap is fixed here, please narrow the PR description from "the rollout resumes" to "Parallel-mode target-side stall" and record gaps 1 and 2 in #464 explicitly, so `Related to #464` is not read as "addressed". Otherwise this is good to merge as-is.

pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control.go
jiangwu25: Thanks for pointing this out. A retry alone cannot free that budget. I added tests in aaecd679 to protect the availability limit and noted both cases in #464.

pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control.go
jiangwu25: That makes sense. A timer would still hit the same readiness check. The description now says Parallel mode, and I noted OrderedReady in #464.

pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control.go
jiangwu25: I kept the 1ns floor here. This edge case only needs one extra pass, so I'd rather avoid another second of delay.

pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control.go
diw-zw: **Correction / retraction of this comment.** After implementing and testing it, my proposed fix here is wrong on two counts, and the underlying "issue" is not a defect.

**1. The two shapes I described are not bugs — blocking there is correct.**

The budget consumers I named (an instance already at `updateRevision` but NotReady, and a base below `partition`) are deliberately outside `buildUpdateTargets`, so the rollout is not supposed to touch them. Treating their unavailability as consuming `maxUnavailable` is the same accounting k8s Deployment's `cleanupUnhealthyReplicas` uses: with 1 already-unavailable replica and `maxUnavailable: 1`, the second (healthy) replica must wait. I re-ran both shapes on `aaecd679` across 10 consecutive reconciles with the window expired: `deletions=0` every time, i.e. no progress is being withheld, and the new `TestProgressUpdateBudget` subtests (`unhealthy updated base still consumes budget`, `unhealthy base below partition still consumes budget`) pin exactly this and pass. That is the intended behavior.

**2. The fix as written introduces a real regression: a 1ns busy-loop.**

Copying the `max(time.Until(...), time.Nanosecond)` pattern over *every* budget consumer makes the floor reachable — the reason it is currently dead code is that the branch requires `!isFree`, and `isFree` includes `isStablyUnhealthy`, so a target that reaches it always has `time.Until(...) > 0`. A non-target consumer can have a long-expired window, so `max(..., 1ns)` selects `1ns`.

Measured on `aaecd679` with the naive change applied (healthy target at ord0, updated-but-crashing consumer at ord1, `maxUnavailable: 1`, window expired):

```
30 reconciles -> pushes=30 (every one 1ns), deletions=0
```

So each reconcile schedules a ~1ns requeue and makes no progress: an unbounded hot loop driven by the returned `RequeueAfter`. Note the existing test suite still passes with this change, so CI would not catch it.

A narrower variant (push only while `time.Until(...) > 0`, no floor) avoids the hot loop but is still a no-op for these shapes — the expired-window reconciles produce `deletions=0` — so it adds scheduling churn without unblocking anything.

**Conclusion:** please disregard the suggestion to drive the retry from `baseUnavail` consumers. The right action for this thread is to keep the current target-only behavior; no code change is needed here. Sorry for the noise.

pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control.go
diw-zw: **Correction: this nit is wrong — the floor is unreachable, so there is nothing to change here.**

`max(time.Until(...), time.Nanosecond)` never selects `time.Nanosecond`. This branch is only entered when `!isFree`, and `isFree` includes `isStablyUnhealthy(target)` (which is `time.Since(first) >= stableUnhealthyDuration`). Therefore every target that reaches this line has `time.Until(first.Add(stableUnhealthyDuration)) > 0`, so the `max` always picks the positive `time.Until` value.

Verified two ways: (a) an exhaustive check over ages `[0, 3*window]` shows no age can both enter the branch and produce `time.Until(...) <= 0`; (b) instrumenting this exact line and running the full package suite (plus the new subtests on `aaecd679`) logs zero hits. Contrast with the non-target variant discussed at line 618, where the floor *does* become reachable and does cause a 1ns hot loop — the floor is safe here precisely because it stays dead code.

Please ignore this comment; no change is warranted.

pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_control.go
diw-zw: **This one still stands on `aaecd679` — and I have now tested the candidate fixes, which is worth recording here.**

Re-verified on the new head (control.go is byte-identical to `b0931fa3`, so the line refs hold): with `podManagementPolicy: OrderedReady`, replicas=2, maxSurge=0, maxUnavailable=2, both instances NotReady and the window expired:

```
OrderedReady: retriesScheduled=0  deletions=0   (5 reconciles; "Progress rolling update" logged 0 times)
Parallel:     retriesScheduled=5  deletions=10  (same fixture)
```

Phase B returns at `control.go:474` before Phase C (`control.go:487`), so the new `Push` is dead code under OrderedReady. This is broader than the retry: the `isStablyUnhealthy` free path is in Phase C too, so under OrderedReady a permanently broken base can never be cleanup-deleted — it can only wait for an external event. That asymmetry predates this PR.

What I measured for the possible fixes (all on `aaecd679`):

| Candidate | Unsticks OrderedReady? | Side effect |
|---|---|---|
| Push `stableUnhealthyDuration` at observation time | no (nothing reaches Phase C) | **breaks `TestUpdateStatefulInstanceSetRetriesUnhealthyRollout`** (`expired health windows requested another retry: 10s`) — unbounded 10s requeue |
| Same, but only while `remaining > 0` | no (schedules 0 retries) | none, but a no-op |
| Run Phase C even when Phase B exits early | **yes** (`deletions=10`; full suite still green) | **breaks OrderedReady ordering**: with ord0 healthy and ord1 still unsettled, baseline deletes `[]` but this deletes `[ord1, ord0]` — it tears down a healthy lower replica |

So "just schedule a retry" cannot fix this, and the change that does fix it collides with the OrderedReady contract of waiting for each instance in order. That makes it a design decision for #464 rather than a patch to fold into this PR.

**Suggestion:** keep the code as-is here and narrow the PR description from "the rollout resumes on its own" to Parallel-only, then track the OrderedReady stall (including the unreachable `isStablyUnhealthy` cleanup path) separately in #464. If you'd rather fix it here, the early-return at `control.go:474` is the place to reason about, but it needs an explicit decision on whether OrderedReady may tear down a healthy lower ordinal.

diw-zw (commented):
## Updated review — after re-testing my earlier comments on `aaecd679`

I implemented and ran each suggestion instead of leaving them as reasoning, and two of my three inline comments do not survive that. Corrections posted in-thread; summary here so the record is unambiguous.

**Retracted — my `control.go:618` comment (non-target budget consumers).**
The two shapes I flagged (an instance already at `updateRevision` that is NotReady, and a base below `partition`) are intentionally not update targets, and counting their unavailability against `maxUnavailable` is correct Deployment-style accounting. I re-measured: with the window expired, 10 consecutive reconciles give `deletions=0` — no progress is being withheld. Your new subtests cover exactly this and pass. Additionally, the fix I proposed (fan the retry out over all `baseUnavail` consumers) makes the `1ns` floor reachable and produces a busy-loop: applying it gives `30 reconciles -> pushes=30 (all 1ns), deletions=0`. Recommend ignoring that comment; the current target-only behavior should stay.

**Retracted — my `control.go:622` nit (`time.Nanosecond` floor).**
The floor is provably unreachable here: entering the branch requires `!isFree`, and `isFree` includes `isStablyUnhealthy`, so `time.Until(first.Add(stableUnhealthyDuration))` is always positive at this line. Confirmed by exhaustive age sweep plus an instrumented run of the full suite (zero hits). No change needed.

**Stands — `control.go:623` (OrderedReady).**
Re-verified on this head: OrderedReady schedules 0 retries and deletes nothing, because Phase B returns at `control.go:474` before Phase C ever runs. It also means the `isStablyUnhealthy` cleanup path is unreachable under OrderedReady. I tested the fixes and documented the table in-thread: retry-at-observation either does nothing or breaks the PR's own retry test; un-gating Phase C fixes the stall but tears down a healthy lower ordinal, violating OrderedReady ordering. This is a design decision, not a safe drive-by patch.

**Net: the PR is in good shape.** `control.go` is unchanged since my first pass, `return` → `continue` is behavior-preserving, multi-target timing keeps the earliest deadline, and `aaecd679`'s tests are correct and pass (`go vet`, `gofmt`, package + `./pkg/reconciler/roleinstanceset/...` all clean). The only open item is the OrderedReady scope note above.

**One concrete ask:** narrow the PR description's claim from "the rollout resumes" to Parallel-mode, since under OrderedReady this PR schedules no retry. Track the OrderedReady stall separately in #464.

diw-zw (approved):

diw-zw:
/lgtm
Thanks for the contribution! @jiangwu25

PR #472, opened 2026-09-16
chore(api): mark InPlaceOnly update strategy as deprecated

NoobDream2568:
### Ⅰ. Motivation
The `InPlaceOnly` update strategy is not separately implemented and behaves identically to `InPlaceIfPossible`; deprecate it to steer users toward `InPlaceIfPossible`.

### Ⅱ. Modifications
Added a `Deprecated` marker on the `InPlaceOnlyUpdateStrategyType` constant and deprecation notes on the related field comments in the API (the enum value is retained for backwards compatibility), regenerated CRDs/kubectl manifests, and updated the best-practice docs (en/zh), the update-strategy doc, and the rolling-update example.

NONE.

No new tests needed: comment/doc-only changes with no behavior change.

No behavior change.

## Checklist


api/workloads/v1alpha2/rolebasedgroup_types.go
```diff
@@ -111,6 +111,9 @@ const (
 	InPlaceIfPossibleUpdateStrategyType UpdateStrategyType = "InPlaceIfPossible"
 
 	// InPlaceOnlyUpdateStrategyType - Only use in-place update.
+	//
+	// Deprecated: InPlaceOnly is not separately implemented and behaves identically
```
diw-zw, line 115: **Non-blocking: no regression anchor for the "behaves identically" claim.**

This deprecation rests entirely on `InPlaceOnly` being equivalent to `InPlaceIfPossible`, but nothing pins that equivalence. The updated envtest cases only assert the value *passes through admission*; none asserts the two types produce the same rollout behavior. If `InPlaceOnly` ever gains distinct semantics, these docs silently become wrong and no test catches the drift.

Suggestion: add one envtest that creates an `InPlaceOnly` RBG, then (a) patches the container image and asserts in-place (same Pod UID/IP), and (b) patches a non-image field such as `resources` and asserts recreate -- matching the documented `InPlaceIfPossible` outcomes.

api/workloads/v1alpha2/roleinstanceset_types.go
```diff
@@ -119,6 +119,7 @@ type RoleInstanceSetScaleStrategy struct {
 type RoleInstanceSetUpdateStrategy struct {
 	// Type indicates the type of the RoleInstanceSetUpdateStrategy.
 	// Default is InPlaceIfPossible.
+	// Note: InPlaceOnly is deprecated; use InPlaceIfPossible.
```
diw-zw, line 122: **Minor semantic note (not a blocker).**

`pkg/reconciler/roleinstanceset/statelessmode/core/implement.go:163` gates the in-place template-diff validation on `!= InPlaceIfPossible`, i.e. it does **not** include `InPlaceOnly`. Strictly speaking the two types are not identical on that path. It is currently dead code (declared in `core/api.go`, implemented, but never invoked), so there is no runtime impact today -- just a trap for whoever eventually implements `InPlaceOnly`.

examples/basic/rbg/update-strategy/rolling-update.yaml
```diff
@@ -33,13 +33,13 @@ spec:
                 ports:
                   - containerPort: 80
 
-    # InPlaceOnly: strictly in-place update only
+    # RecreatePod: delete and recreate pods (default Kubernetes behavior)
```
diw-zw, line 36: **Scope nit.** This hunk repurposes `decode` to `RecreatePod` *and* drops the third `router` role. Two side effects: (1) the file header still says "Demonstrates RollingUpdate with different update types" but now only two of them are shown; (2) the only role carrying `maxSurge: 1` is gone (the remaining roles use `maxSurge: 0`), so the example no longer demonstrates surge behavior. Consider keeping a third role, or trimming the header claim.

test/envtest/testcase/rbg/basic_test.go
```diff
@@ -298,7 +298,8 @@ var _ = Describe("RoleBasedGroup Controller", func() {
 			for _, updateType := range []workloadsv1alpha2.UpdateStrategyType{
 				workloadsv1alpha2.RecreatePodUpdateStrategyType,
 				workloadsv1alpha2.InPlaceIfPossibleUpdateStrategyType,
-				workloadsv1alpha2.InPlaceOnlyUpdateStrategyType,
+				// Deprecated but still a valid enum value.
```
diw-zw, line 301: **Correct and necessary change.** The new `// Deprecated:` marker trips staticcheck SA1019 for *cross-package* references, so these string-literal swaps (here and in `rolebasedgroupset_controller_test.go` / `webhook/default_test.go`) are required to keep `make lint` green. Verified locally: restoring `workloadsv1alpha2.InPlaceOnlyUpdateStrategyType` at `rolebasedgroupset_controller_test.go:889/910` yields 2 SA1019 findings. The same-package references in `api/workloads/v1alpha2/helper_test.go` / `rolebasedgroup_default_test.go` are correctly left untouched (same-package uses are not flagged).

diw-zw (commented):
## Review: ACK cluster end-to-end verification

Reviewed `29185a03` and validated it on a real ACK Pro cluster (Chorus-built image `29185a`, `docs/` + CRD state deployed). **Overall: looks good / approve.** The change is a clean, backwards-compatible deprecation; the test-file edits are a *required* consequence of the new marker, not gratuitous churn.

### What I verified

**Static / unit**
- `make manifests` + `make generate` -> zero diff (generated CRDs & kubectl manifests are in sync).
- `golangci-lint run` -> only 1 pre-existing finding on `main` (`rolebasedgroup_controller_test.go:2522`, goconst); nothing introduced here.
- `go test ./api/... ./internal/... ./pkg/...` and `make test-envtest` (rbg / webhook / restart_policy) -> all green.
- `yamllint` / `helm lint` / markdownlint -> no new issues.

**Behavior on a live cluster (PR's stated no-behavior-change claim)**

| Case | Expected | Result |
|---|---|---|
| CRD `description` shows the deprecation note (`kubectl explain`) | note present | PASS (3 CRDs + manifests) |
| `InPlaceOnly` + image-only change | in-place (same Pod UID/IP) | PASS |
| `InPlaceIfPossible` + `resources` change | recreate (new UID/IP) | PASS |
| **`InPlaceOnly` + `resources` change** | **recreate, identical to `InPlaceIfPossible`** | PASS |
| status convergence | readyReplicas == replicas | PASS |
| invalid enum value (webhook and `--validate=false`) | rejected | PASS |
| legacy `Recreate` / empty type | healed to `RecreatePod` / defaulted to `InPlaceIfPossible` | PASS |
| concurrency x10 + rapid type flips | no deadlock/panic (0 controller restarts) | PASS |
| PR-modified example YAML applies | 4/4 Ready | PASS |

The deprecation claim ("behaves identically to `InPlaceIfPossible`") holds in practice.

Three non-blocking suggestions are inlined below.

cheyang (approved):
/lgtm
/approve

PR #473, opened 2026-09-16
Kep-473: topology-aware scheduling

diw-zw:
### Ⅰ. Motivation

Two inference-serving scenarios need network-topology-aware placement that RBG
cannot express today:

1. **Multi-node instance packing.** Models too large for a single node are served
   with multi-node TP/PP: one serving instance spans several machines (e.g., a
   prefill instance of 4 nodes) whose pods exchange heavy RDMA traffic. All pods of
   one instance must pack into a single high-performance network domain, so
   cross-machine traffic never crosses oversubscribed spine links.
2. **Prefill–decode co-location.** In PD-disaggregated serving, prefill and decode
   instances of the same group constantly transfer KV cache; they must stay within
   the same network block to keep transfer latency (TTFT) deterministic — gathered
   tightly when capacity allows, but never across the block boundary.

Every AI scheduler already models this, but each speaks a different dialect —
Volcano `HyperNode` tiers/tierNames, Koordinator `topologyLayer` gather strategies,
KAI level aliases. As the workload API, RBG should let users state topology intent
once and have it honored on the configured scheduler.

### Ⅱ. Modifications

Adds `keps/473-topology-aware-scheduling/` (`README.md` + `kep.yaml`, status:
provisional, stage: alpha, milestone: v0.9.0):

- **API**: `TopologyConstraint` (`pack.required` / `pack.preferred`) attached at
  `spec.topologyConstraint` (group = whole RBG) and
  `role.leaderWorkerPattern.topologyConstraint` (group = one RoleInstance).
- **Pass-through level identifiers**: values are the configured scheduler's own level
  names (Volcano `tierName`, Koordinator `topologyLayer`, KAI alias) — no abstract
  vocabulary, no cluster-side mapping resource; the controller is bound to one
  scheduler at startup, so one string only ever needs one dialect.
- **Translation layer** riding the KEP-430 gang channel: renders Volcano
  `networkTopology` / `subGroupPolicy[].networkTopology`, Koordinator `gatherStrategy`,
  and KAI topology constraints; two-level (required + preferred) declarations where
  the dialect supports them.
- **Explicit failure semantics**: reconcile-time validation of level existence and
  ordering against the scheduler's own topology objects; failures surface as
  `TopologyTranslated=False` conditions (`LevelUnresolved`, `InvalidLevelOrder`,
  `SchedulerUnsupported`) plus edge-triggered events on the declaring object —
  never silent degradation (Volcano silently drops unknown `highestTierName`).
- **Gather-only scope**: "spread" is deliberately excluded (no scheduler dialect
  supports it as a first-class citizen).
- Documented fidelity gaps: Volcano requires `tierName` (integer
  `highestTierAllowed` is never rendered); Volcano soft semantics absorbed by
  scoring; Koordinator cannot express instance-level constraints (one uniform
  topology spec per GangGroup).
- Rejected alternatives recorded: TopologyProfile CRD vocabulary indirection,
  node label keys as a uniform vocabulary, structured per-scheduler fields.

ref #473 (KEP tracking issue — this PR adds the proposal document and does not
close it; implementation PRs follow per the phase plan)

NONE — documentation-only KEP. A test plan for the implementation phases
(unit/integration/e2e) is included in the KEP itself.

Review the rendered `keps/473-topology-aware-scheduling/README.md` and the metadata
in `kep.yaml`. Design decisions most worth scrutiny:

1. Pass-through level identifiers instead of a vocabulary indirection (rationale in
   "Level Identifiers" and "Alternatives").
2. Volcano requires `tierName` — no integer-tier fallback, one string carries one
   semantic.
3. Topology constraints ride the gang channel: an RBG/role with constraints gets a
   PodGroup even without gang scheduling configured.
4. Koordinator instance-level constraints reported unsupported rather than faked.

- Follows up on the dialect survey and design iterations done with KEP-430 (gang
  scheduling); the gang channel and CRD-schema-detection pattern are reused.
- The Chinese companion translation is maintained out-of-tree and can be added in a
  follow-up if the community wants localized KEPs.

## Checklist


keps/473-topology-aware-scheduling/README.md
```diff
@@ -0,0 +1,456 @@
...
+
+### User Stories
+
+#### Story 1: Multi-Node Instance Packing
+
+As an inference platform engineer deploying very large models with multi-node TP/PP, I need every pod of one serving instance (e.g., 4 machines of a prefill instance) to be packed inside a single high-performance network domain, so that cross-machine RDMA traffic never crosses oversubscribed spine links. I declare a hard instance-level constraint (`pack.required: rack`) once on the role; every instance created by scale-out, rolling update, or recovery keeps honoring it.
+
+#### Story 2: Prefill–Decode Co-Location
+
+As an inference application developer running PD-disaggregated serving, I need prefill and decode pods of the same group to land in the same network block so that KV-cache transfer latency (TTFT) stays deterministic. I declare `pack.required: block` as the hard upper bound and `pack.preferred: rack` as the soft target in a CoordinatedPolicy rule over `roles: [prefill, decode]`: the scheduler gathers the prefill+decode pods into one rack when capacity allows, and may fall back to block otherwise — but never crosses the block boundary.
```
cheyang, line 90: This scope is not yet the "same group" described by the story: the rule contains every current and future instance of both roles, so one hard `block` constraint pins the entire deployment—including rollout surge and scale-out—to a single block. A paired/segmented PD deployment usually needs one P/D segment per domain, rather than one domain for all replicas. Please either define the segment identity and its lifecycle here, or narrow this story and document the resulting hard capacity ceiling. RBGSet cannot currently recover this by templating one policy per child RBG: `pkg/scheduler/common/gang_strategy.go` looks up CoordinatedPolicy by the exact RBG name/namespace, while `internal/controller/workloads/rolebasedgroupset_controller.go` does not create those policies.

keps/473-topology-aware-scheduling/README.md
```diff
@@ -0,0 +1,456 @@
...
+- **Cross-role level** (`spec.policies[].strategy.scheduling.topology` on CoordinatedPolicy): constraint group = all pods of the roles named in the rule's `roles` list (cross-role co-location). Roles not listed stay outside the cross-role domain (their own role-level constraints still apply). Listing every role of the RBG recovers whole-group co-location, so no dedicated RBG-level field exists.
+
+**Translation layer.** Per scheduler, a translator compiles (constraint group, level identifier, hard/soft) into the scheduler dialect — mostly verbatim pass-through — riding the gang-scheduling channel (KEP-430): the Volcano `PodGroup`/`subGroupPolicy` is reused as the carrier, so topology composes with gang semantics on the same objects. Levels are validated at reconcile against the scheduler's own topology objects (HyperNode tierName set, ClusterNetworkTopology layers, Topology CR aliases); failures (unknown level, incapable scheduler) are explicit: status conditions plus events, never silent degradation.
+
+### Notes/Constraints/Caveats
```
cheyang, line 107: Please add one scheduler-independent placement-planning step before translation. KEP-430 collapses gang rules into one coverage set/PodGroup, while this KEP lets each topology rule determine a PodGroup partition; allowing the gang and topology translators to choose membership independently makes composition order-dependent, and a pod can carry only one `group-name`. A concrete model would canonicalize logical groups by role set / instance partition: equal scopes merge gang and topology attributes, disjoint or contained scopes form a tree, partially overlapping scopes are rejected, and the scheduler backend alone compiles that tree into physical PodGroups or reports it unsupported. This also gives future segment/spread strategies one owner for membership instead of another independent layout algorithm.

keps/473-topology-aware-scheduling/README.md
```diff
@@ -0,0 +1,456 @@
...
+- **Gather-only semantics.** The API models "gather into a common domain" (聚). "Spread across domains" (散) is deliberately excluded.
+- **The constraint group is exactly the rule's role set.** Referenced roles join one common domain; unreferenced roles stay outside it (their own role-level constraints still apply). A role may appear in at most one topology-bearing rule — PodGroup membership partitions pods — and overlap is rejected by the CoordinatedPolicy validating webhook, following the `gang.minReplicas` validation pattern.
+- **Soft semantics are approximate on Volcano.** Volcano cannot anchor a soft target at a middle tier; in hard mode its scoring naturally prefers lower tiers, so "tighten within the required bound" is absorbed by scoring rather than declared.
+- **Volcano requires tier names.** Constraints reference `HyperNode.spec.tierName` only; integer `highestTierAllowed` is never rendered. Clusters whose HyperNodes predate `tierName` must adopt it (otherwise they are unsupported).
+
+### Risks and Mitigations
+
+| Risk | Mitigation |
+|---|---|
```
cheyang, line 118: `preferred` is not actually translated on Volcano here: `preferred: rack` produces no field anchoring the preference at `rack`, and generic lower-tier scoring is not equivalent to honoring that requested target. Please treat this as an explicit fidelity loss—for example, an admission warning plus a `PreferredAbsorbed`/degraded condition—define both preferred-only and required+preferred behavior, and add a scheduler-level test for the scoring claim. Otherwise the same public field silently carries materially different semantics across backends.

keps/473-topology-aware-scheduling/README.md
```diff
@@ -0,0 +1,456 @@
...
+
+### Attachment Points
+
+```yaml
+# RoleBasedGroup spec
+spec:
+  roles:
+  - name: router
+    replicas: 1
+  - name: prefill
```
cheyang, line 216: The KAI entry appears factually incorrect for KAI v0.16.9: [`TopologyLevel`](https://github.com/kai-scheduler/KAI-scheduler/blob/v0.16.9/pkg/apis/kai/v1alpha1/topology_types.go#L65-L81) has only `nodeLabel`, not an alias, and [`requiredTopologyLevel` / `preferredTopologyLevel`](https://github.com/kai-scheduler/KAI-scheduler/blob/v0.16.9/pkg/apis/scheduling/v2alpha2/podgroup_types.go#L381-L395) therefore carry that label key. The same KAI constraint also requires the `topology` CR name. Please correct the table and the string-uniformity rationale: pass-through on KAI means an infrastructure-specific node-label key, unlike Volcano `tierName` or Koordinator `topologyLayer`, and the API still needs a deterministic source for the Topology resource name.

keps/473-topology-aware-scheduling/README.md
```diff
@@ -0,0 +1,456 @@
...
+- **Cross-role co-location** — one constraint group per topology-bearing rule, rendered at the group level: Volcano uses the rule PodGroup's `spec.networkTopology`; Koordinator scopes the GangGroup to the member PodGroups of the rule's roles; KAI renders a subGroup that nests the rule's roles (uncovered roles stay sibling subGroups, unconstrained) and carries its own `topologyConstraint`. The levels nest naturally — a rule's domain contains the instance domains of its member roles (e.g., block ⊃ rack) — so any level combination is satisfiable and no cross-level ordering validation is needed.
+
+#### Translation Matrix
+
+| `pack` API (role-level or cross-role) | Volcano | Koordinator gatherStrategy | KAI |
+|---|---|---|---|
```
cheyang, line 257: Being satisfiable is not sufficient reason to omit this validation. If a parent rule requires `rack` while a contained RoleInstance requires `block`, the child declaration is reversed/redundant and almost certainly signals a configuration mistake; it also makes the hierarchy harder to evolve consistently. Since reconcile already obtains an ordered level source, require each child constraint to be equal to or narrower than its parent (or at least emit a warning), and apply the analogous check to preferred levels. Grove enforces this invariant for PCS → PCSG → PCLQ rather than accepting every satisfiable combination.

cheyang (commented):
Design review focused on placement-group composition, workload granularity, and scheduler-dialect fidelity. The inline comments below intentionally avoid repeating the existing review findings.

PR #474, opened 2026-09-17
feat: support RoleBasedGroupSet rolling update

NoobDream2568:
### Ⅰ. Motivation

A `RoleBasedGroupSet` propagates its `spec.groupTemplate` to a set of child `RoleBasedGroup`s. Today that propagation is un-paced: when the template changes, every outdated child is updated within a single reconcile, with no ordering and no availability gating. For a set of N groups that means all N groups can restart at once — availability might drop to zero for the whole set.

This PR adds an opt-in rolling update for `RoleBasedGroupSet`, so a template change is propagated one group at a time under an unavailability budget, while scaling stays immediate. It is the group-set analogue of the role-level rollout the `RoleBasedGroup` already offers.

#### API (`api/workloads/v1alpha2/rolebasedgroupset_types.go`)

A new optional `spec.rolloutStrategy`. The strategy is intentionally flat — its five fields sit directly under `spec.rolloutStrategy`, with no intermediate `rollingUpdate` wrapper, because a set always rolls its child groups and there is no alternative mechanism to select.

**Three propagation modes** The field being unset, a replicas-only diff, and a real template change are routed to three different code paths that never converge:

| Situation | Path | Semantics |
|---|---|---|
| `rolloutStrategy` **unset** | `reconcileStatic` | Legacy behavior, unchanged: every outdated child `RoleBasedGroup` is **updated in place** within a single reconcile — the child's `spec.roles` is overwritten, no child object is deleted, and each downstream workload rolls its own Pods according to its role-level `rolloutStrategy`. Whole-set, immediate, un-gated, exactly as before this PR. |
| `rolloutStrategy` **set** + diff is replicas-only | in-place scale | See "A replicas-only diff never recreates" below. |
| `rolloutStrategy` **set** + any other diff | paced recreate rollout | Outdated children are deleted and rebuilt one at a time under the budgets. |

**`type` selects how ONE child group moves onto the new template, and currently offers two semantics:**

- **`Recreate`** (default, implemented): the child `RoleBasedGroup` is deleted and rebuilt from `groupTemplate` under the same name and ordinal. Every downstream workload and Pod is replaced, so the group restarts as a whole and gang consistency is preserved. The role-level `rolloutStrategy` of the `RoleBasedGroup` has no effect in this mode.
- **`InPlaceUpdate`** (reserved, rejected by the CRD enum today): would update the child `RoleBasedGroup` object in place, keeping its name and UID, and let each downstream workload decide how its Pods move to the new template according to its own role-level `rolloutStrategy`. The controller does not implement this yet.

| Field | Type | Default | Meaning |
|---|---|---|---|
| `type` | enum `{Recreate}` | `Recreate` | As above. `InPlaceUpdate` is reserved and rejected by the CRD enum. |
| `partition` | IntOrString | `0` | Number of lowest ordinals held back on the previous template. Only ordinals in `[partition, replicas)` roll. May equal `replicas` (freezes everything, so `maxSurge` can hold a standing canary). |
| `maxUnavailable` | IntOrString | `1` | Max serving groups the rollout may take down at once. Rounded **up** when `maxSurge == 0` and **down** otherwise, so a resolved `0` is reachable only together with a surge budget. |
| `maxSurge` | IntOrString | `0` | Max groups created above `replicas`, on ordinals `[replicas, replicas+maxSurge)`. Surge groups take real traffic exactly like base groups. |
| `paused` | bool | `false` | Freezes template propagation. Scaling still happens while paused, and existing surge groups are kept. |

Status grows four counters — `currentReplicas`, `updatedReplicas`, `updatedReadyReplicas`, `expectedUpdatedReplicas` — and two conditions, `Rolling` and `Paused`. All counters are computed over base ordinals only; surge groups are never counted as replicas.

#### Controller (`internal/controller/workloads/rolebasedgroupset_controller.go`)

Children are named deterministically `<rbgset>-<index>` and carry a `rbg.workloads.x-k8s.io/groupset-index` label. `classifyChildren` splits them into `base` (`[0, replicas)`), `surge` (`>= replicas`) and `invalid` (missing/broken label). `Reconcile` branches on `spec.rolloutStrategy`: `nil` → `reconcileStatic` (the legacy whole-set in-place update, unchanged), set → `reconcileRolling`.

**`reconcileRolling` classifies every base child into exactly one of three buckets:**

- **up-to-date** — `rolesEqual(child, template)` is true. Only template label/annotation drift is synced in place.
- **scale-only** — `onlyReplicasChanged(child, template)` is true: the role set is identical except for `replicas`.
- **outdated** — everything else. Recreated under the budget.

**A replicas-only diff never recreates.** Scale-only children are applied **in place** by `updateExistingRBGs` (the child keeps its name and UID; the RoleInstanceSet underneath scales the pods). This path is not a fallback that can degrade to recreate: a replicas-only diff never reaches `recreateOutdatedGroups`, in any reconcile, under any budget state. It is also not paced by the budget and still happens while `paused`, because scaling capacity and rolling a template are different operations.

`rolesEqual` compares two role slices by copying them, sorting by role name, then `reflect.DeepEqual`, so the comparison is order-insensitive and independent of API-server defaulting. Because versioning is pure content comparison with no revision history, a template flip-flop (A → B → A) is idempotently convergent: a group still on A is up-to-date again the moment the template returns to A, and a group that moved to B is recreated back to A.

`recreateOutdatedGroups` implements the pacing:

1. Compute `unavailableBase` = missing base ordinals (in-flight capacity) + base groups that are not `isServing`.
2. `budget = maxUnavailable + readySurge` — ready surge groups widen the budget, which is what makes `maxUnavailable: 0` possible without stalling.
3. Sort outdated groups by **descending** ordinal; skip any below `partition`; delete with **foreground** propagation and a **UID precondition**. Deletion only — the next reconcile's scale-up step refills the missing ordinal.
4. **The budget gates serving groups only.** A serving group may be deleted only while `unavailableBase < budget`; a group that is already not serving consumes no budget and is replaced immediately. `maxUnavailable` limits how much *working capacity* the rollout takes down — a broken group is already down, and holding it back would slow a rollback to a healthy template for no gain.

Foreground propagation is what makes the budget hold: the group stays visible and terminating until its whole `RoleInstanceSet` → `RoleInstance` → Pod chain is gone, so it keeps counting as unavailable instead of letting a premature replacement free the budget. Descending ordinal + UID precondition together make acting on a possibly stale informer snapshot safe.

`isServing` = Ready condition true **and** `DeletionTimestamp` zero. A group being deleted keeps `Ready=True` while its Pods run preStop hooks, so every readiness decision (budget, surge lifetime, status) goes through `isServing` to avoid overcounting availability.

**Partition scopes both the rollout and its completion.** Ordinals below `partition` are held back by design, so `rolloutComplete` requires a template match only from in-scope ordinals (`>= partition`) — while still requiring every base group to be serving — and the `currentReplicas` status counter likewise ignores held-back ordinals. A rollout with `partition > 0` therefore reaches `RolloutComplete` and releases its surge groups instead of reporting "still on the previous template" forever.

Surge groups are created only while the rollout is in progress and not paused, and are reclaimed once the rollout is complete or they exceed the budget. `rolloutComplete` requires every base ordinal to be both on the template **and** serving, so surge capacity is not withdrawn while a just-recreated group is still coming up.

#### Webhook (`api/workloads/v1alpha2/rolebasedgroupset_admission.go`)

`validateSpec` is now shared by create and update and additionally runs `validateGroupSetRollout`, which rejects `maxSurge` and `maxUnavailable` both resolving to `0` (no way to make progress) and `partition > replicas`.

Besides, a `validateNoRoleScalingAdapter` rule is added. A RoleBasedGroupSet owns `spec.roles` of every child RBG and keeps it consistent with `spec.groupTemplate`.

NONE.

**Unit tests** (`rolebasedgroupset_controller_test.go`, fake client):

- `TestRollingUpdate_PacedRecreate` — one group recreated per reconcile under `maxUnavailable`.
- `TestRollingUpdate_TemplateFlipFlopConverges` — A→B→A template flip converges: the not-updated group is deleted immediately and both groups return to A.
- `TestRollingUpdate_NotServingOutdatedBypassesBudgetOnly` — the budget gates serving groups only: a not-serving outdated group is replaced even with the budget exhausted, a serving one still waits.
- `TestRecreateDeleteCarriesUIDPrecondition` — the delete carries the target UID.
- `TestRollingUpdate_ReplicasOnlyChangeIsScaledInPlace` — a replicas-only diff scales in place and does **not** recreate.
- `TestOnlyReplicasChanged` — table test for the scale-only detector.
- `TestRollingUpdate_PartitionHoldsBackLowerOrdinals` — ordinals below `partition` stay on the old template.
- `TestRollingUpdate_PartitionRolloutReportsComplete` — held-back ordinals do not keep the rollout from reporting complete or releasing surge.
- `TestRollingUpdate_SurgeGroupsCreated` / `TestRollingUpdate_SurgeKeptUntilRecreatedGroupIsReady` — surge lifecycle, including `maxUnavailable: 0`.
- `TestRollingUpdate_RolloutCompleteRequiresReady` — completion needs serving, not just template equality.
- `TestRoleBasedGroupSetReconciler_TerminatingChildIsNotCountedReady` — `isServing` excludes terminating groups.
- `TestRollingUpdate_PausedFreezesRollout` — paused freezes propagation but not scaling.
- `TestStaticUpdate_AllGroupsUpdatedInOnePass` — the unset-strategy path is unchanged.
- `TestResolveRollingParams`, `TestClassifyChildren`, `TestSyncRBGMetadataKeepsSystemKeys` — helpers.

**E2E tests** (`test/e2e/testcase/v1alpha2/rbgs.go`, real kind cluster):

- `role replicas change scales groups in place without recreating them`
- `recreate rolling update replaces one group at a time`
- `surge backs a zero maxUnavailable so ready groups never drop below replicas`
- `surge group survives until the rollout completes`
- `partition rolls the upper ordinals first and releases the rest when lowered`
- `paused holds the rollout still until it is lifted`

**Webhook E2E** (`test/e2e/testcase/v1alpha2/webhook_validation.go`): create/update validation for the new rules.

```bash
# Unit + generated-code drift check
go build ./...
go test ./internal/controller/workloads/... ./api/... -count=1

# E2E against a kind cluster
make test-e2e
```

### Ⅵ. Special notes for reviews

- **Backward compatibility.** With `spec.rolloutStrategy` unset the controller runs `reconcileStatic`, which is the pre-existing behavior. The rolling path is fully opt-in.
- **Scale vs rollout are deliberately separated.** `onlyReplicasChanged` routes a replicas-only diff to in-place scaling, bypassing the budget and the pause gate, and there is no code path from scale-only to recreate. Please check the three-way classification in `reconcileRolling` (up-to-date / scale-only / outdated).
- **The budget gates serving groups only** (`recreateOutdatedGroups`): a not-serving outdated group is replaced immediately. This is deliberate — it keeps a rollback to a healthy template fast — and is what `TestRollingUpdate_NotServingOutdatedBypassesBudgetOnly` pins down.
- **`rolesEqual` is order-insensitive** (sort-by-name then DeepEqual). Content comparison also makes A→B→A template flips idempotently convergent without any revision bookkeeping.
- **Why foreground delete.** The long comment on `recreateOutdatedGroups` explains the interaction with background GC: without foreground, a replacement can adopt the still-terminating chain and report Ready early, freeing the budget for a second group. The UID precondition + descending-ordinal order are what make a stale snapshot safe.
- **`isServing` centralizes readiness.** Every availability decision goes through it so the budget, surge lifetime and reported status cannot drift apart.
- **Surge groups are real capacity, not labeled canaries.** They occupy ordinals `>= replicas` and take traffic; there is no dedicated surge label. Distinguish them by `rbg.workloads.x-k8s.io/groupset-index >= spec.replicas`.
- **Partition scopes completion as well as the rollout.** Held-back ordinals count neither in `rolloutComplete` nor in `status.currentReplicas`, so a partial rollout can report `RolloutComplete` and release surge.
- **`maxUnavailable: 0`** is only meaningful with `maxSurge > 0`; the webhook rejects both being `0`.
- **`InPlaceUpdate` is reserved**, not implemented — the CRD enum accepts only `Recreate` today.

## Checklist


internal/controller/workloads/rolebasedgroupset_controller.go
```diff
@@ -94,115 +97,517 @@ func (r *RoleBasedGroupSetReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+// reclaimSurgeGroups deletes the surge groups that exceed the surge budget or outlived the
+// rollout, and returns the ones still serving rollout capacity.
+func (r *RoleBasedGroupSetReconciler) reclaimSurgeGroups(
+	ctx context.Context, replicas int, children classifiedChildren, params rollingParams, complete bool,
+) (map[int]*workloadsv1alpha2.RoleBasedGroup, error) {
+	logger := log.FromContext(ctx)
+	keptSurge := make(map[int]*workloadsv1alpha2.RoleBasedGroup, len(children.surge))
+	var surgeToDelete []*workloadsv1alpha2.RoleBasedGroup
+	for ordinal, rbg := range children.surge {
+		if ordinal >= replicas+params.maxSurge || complete {
```
diw-zw, line 394: **Blocker — `partition > 0` combined with `maxSurge > 0` makes the surge group churn forever.**

`complete` comes from `rolloutComplete` (L553), which deliberately exempts ordinals below `partition`. So as soon as every in-scope ordinal (`>= partition`) is on the template and serving, `complete == true` and every surge group is reclaimed here. But `outdated` still contains the held-back ordinals below `partition` (they are intentionally left on the old template), so the very next step at L326:

```go
if !params.paused && len(outdated) > 0 && len(keptSurge) < params.maxSurge {
```

sees `len(outdated) > 0` and immediately recreates the surge group. The next reconcile repeats it: delete surge -> create surge -> delete surge -> ...

Reproduced with the fake client (`replicas=2, partition=1, maxSurge=1`, ordinal 1 already rolled and Ready):

```
Reclaiming surge RoleBasedGroups count:1  -> delete s-2
Creating surge RoleBasedGroups  count:1  -> create s-2
... repeated every reconcile; children never settle back to 2
```

The same churn hits the documented "standing canary" (`partition == replicas`), which makes that feature unusable.

Fix: use one consistent notion of "done". Either reclaim surge based on "everything matches the template" (including held-back ordinals) rather than the partition-scoped `complete`, or gate the L326 warm-up on `!complete`. Please also add a unit test for the `partition>0 && maxSurge>0` steady state — none exists today (`TestResolveRollingParams` only covers parameter parsing).

internal/controller/workloads/rolebasedgroupset_controller.go
```diff
@@ -94,115 +97,517 @@ func (r *RoleBasedGroupSetReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		byName[child[i].Name] = &child[i]
+	}
+	for i := range template {
+		c, ok := byName[template[i].Name]
+		if !ok {
+			return false
+		}
+		// Compare with Replicas masked out; sharing the pointer makes the field equal.
+		cc := *c
+		cc.Replicas = template[i].Replicas
```
diw-zw, line 605: **Blocker — `onlyReplicasChanged` skips the legacy-strategy normalization that `rolesEqual` does, so a replicas-only diff can be misrouted to recreate.**

`rolesEqual` (L861-869) deep-copies and calls `normalizeRolloutUpdateTypes` before comparing, precisely because a stored template can keep the legacy `Recreate` spelling while children carry the webhook-normalized `RecreatePod`. This function does a raw `reflect.DeepEqual`, so that spelling-only delta is read as a real change.

Reproduced:

```
child    rollingUpdate.type = "RecreatePod"
template rollingUpdate.type = "Recreate"   (legacy spelling); replicas differs
rolesEqual          = false   (correct)
onlyReplicasChanged = false   (wrong)
=> classified as `outdated` and deleted+recreated instead of scaled in place
```

That contradicts this PR's stated contract ("A replicas-only diff never recreates") on any cluster still holding a legacy spelling, turning an intended scale-up into a full group restart.

Fix: mirror `rolesEqual` — `deepCopyRoles` both slices, `normalizeRolloutUpdateTypes`, then compare with `Replicas` masked out.

internal/controller/workloads/rolebasedgroupset_controller.go
```diff
@@ -474,13 +969,24 @@ func (r *RoleBasedGroupSetReconciler) needsTemplateAnnotationUpdate(
...
 	}
 	return false
 }
 
+// isSystemManagedMetadataKey reports whether a label or annotation key belongs to this project's
+// controllers rather than to the user. Template syncing carries those keys over from the child
+// instead of rebuilding them from the template, which is what keeps two controllers from
+// fighting over the same object.
+func isSystemManagedMetadataKey(key string) bool {
+	return strings.HasPrefix(key, constants.RBGPrefix)
```
diw-zw, line 987: **Major — the prefix match swallows user-facing `rbg.*` annotations, silently disabling their removal and update.**

Treating the whole `constants.RBGPrefix` (`rbg.workloads.x-k8s.io/`) as controller-owned means `needsTemplateAnnotationUpdate` (L972) never reports drift, and `syncRBGMetadata` (L1084/L1099) always keeps the child's value for these keys.

But `rbg.workloads.x-k8s.io/group-exclusive-topology` is a user-facing annotation — `doc/features/exclusive-topology.md` and `examples/basic/rbgs/rbgs-exclusive-topology.yaml` both instruct users to set it inside `groupTemplate.annotations`.

Reproduced:

```
template annotations: (empty)
child annotation: rbg.workloads.x-k8s.io/group-exclusive-topology=kubernetes.io/hostname
needsTemplateAnnotationUpdate = false            (should be true)
after syncRBGMetadata the annotation is still present   (should be cleared)
```

So removing or changing exclusive-topology in the template no longer propagates. `TestSyncRBGMetadataKeepsSystemKeys` only pins `DiscoveryConfigModeAnnotationKey`, which is the one key that genuinely needs protecting.

Fix: enumerate the specific controller-owned keys (e.g. `DiscoveryConfigModeAnnotationKey`) instead of matching on the shared prefix.

diw-zw (commented):
Thanks for the PR — the opt-in design, the UID-precondition + foreground-delete reasoning, the `isServing` consolidation, and the `Consistently`/`Eventually` E2E invariants are all solid.

I reviewed with a local build and `go test ./internal/controller/workloads/... ./api/...` (both green), then wrote targeted probes against the fake client. Two **blockers** reproduce deterministically:

1. `partition > 0` combined with `maxSurge > 0` never converges — the surge group is deleted and recreated on every reconcile.
2. `onlyReplicasChanged` skips the legacy-strategy normalization that `rolesEqual` performs, so a replicas-only diff can be routed to recreate instead of in-place scaling.

Two **major** items follow (system-prefix metadata swallowing user-facing `rbg.*` annotations; the scaling-adapter validation locking pre-existing sets), plus a few minor notes — all inline below.

Test-coverage gap worth closing with the fix: there is no reconcile-level test for `partition>0 && maxSurge>0`, and the E2E cases also avoid the combination (`partition=2,maxSurge=0` and `partition=0,maxSurge=1`). That combination is exactly where blocker #1 lives.

api/workloads/v1alpha2/rolebasedgroupset_admission.go
```diff
@@ -68,23 +59,120 @@ func (v *RoleBasedGroupSetValidator) ValidateUpdate(_ context.Context, _ runtime
...
+const scalingAdapterInGroupSetHint = "a RoleBasedGroupSet owns spec.roles of every child RoleBasedGroup and keeps it " +
+	"consistent with spec.groupTemplate; suggest deploying a standalone RoleBasedGroup that needs role-level autoscaling"
+
+// validateNoRoleScalingAdapter rejects roles that enable a ScalingAdapter inside a group
+// template. The RoleBasedGroupSet controller propagates the whole role list to its children,
+// which conflicts with the ScalingAdapter controller writing role replicas back into a child
+// RoleBasedGroup: the two keep overwriting each other and the RoleBasedGroup validator rejects
+// the propagation, so the set never converges. Autoscaling a role therefore requires a
+// standalone RoleBasedGroup.
+func validateNoRoleScalingAdapter(fieldPath string, roles []RoleSpec) error {
```
diw-zw, line 99: **Major (upgrade compatibility) — this runs on every update, and with `failurePolicy=fail` it locks pre-existing sets.**

The ownership/conflict analysis is sound, but because `validateSpec` also runs on update, any `RoleBasedGroupSet` that already has `scalingAdapter.enable: true` in its `groupTemplate` (created before this change) can no longer be updated at all — even an unrelated `spec.replicas` bump or a label change is rejected until the user removes the adapter.

Please either:
- run this only on create, or only when `groupTemplate.spec.roles` actually changed; and/or
- flag it explicitly as a breaking change in the PR description / release notes.

api/workloads/v1alpha2/rolebasedgroupset_types.go
```diff
@@ -43,12 +44,88 @@ type RoleBasedGroupSetSpec struct {
...
+	// GroupTemplate under the same name and ordinal. Every downstream workload and Pod is
+	// rebuilt, so the group restarts as a whole and gang consistency is preserved.
+	// The role level rolloutStrategy of the RoleBasedGroup has no effect in this mode.
+	RecreateStrategyType GroupUpdateStrategyType = "Recreate"
+
+	// InPlaceUpdateStrategyType - Update the RoleBasedGroup object in place,
+	// keeping its name and UID, and let each downstream workload decide how its Pods move
+	// to the new template according to its own role level rolloutStrategy.
+	// the current controller only implements Recreate.
+	InPlaceUpdateStrategyType GroupUpdateStrategyType = "InPlaceUpdate"
```
diw-zw, line 69: Minor — `InPlaceUpdateStrategyType` is defined and exported, but the CRD enum only accepts `Recreate` and the controller does not implement it, so any user who writes it is rejected by the CRD. Consider not exporting the constant yet, or marking it clearly as reserved/not-yet-accepted, to avoid a misleading public API surface.

internal/controller/workloads/rolebasedgroupset_controller.go
```diff
@@ -246,69 +651,155 @@ func (r *RoleBasedGroupSetReconciler) scaleUp(
...
+				updatedReady++
+			}
+		} else if ordinal >= partition {
+			current++
 		}
+	}
+
+	newStatus := *rbgset.Status.DeepCopy()
+	newStatus.ObservedGeneration = rbgset.Generation
+	newStatus.Replicas = baseTotal
```
diw-zw, line 724: Minor — this changes the meaning of a released field: previously `status.replicas == len(children)`, now it counts base ordinals only, so during a surge window `status.replicas` is smaller than the number of child RoleBasedGroups. Please confirm this is intended for the `scale` subresource and note the behavior change in the PR description.

api/workloads/v1alpha2/rolebasedgroupset_admission.go
NoobDream2568: Agree that locking pre-existing sets on every update is too aggressive.

I have updated the validation logic to only run on `CREATE` operations. This ensures backward compatibility for existing RoleBasedGroupSet resources, allowing users to perform unrelated updates (e.g., label changes or replica bumps) without being blocked.

**However, I want to highlight two important context points regarding this configuration:**

- Post-creation modification: Users can still set `scalingAdapter.enable: true` after creation via an update. Since we are not blocking this on update, they might inadvertently enter an unsupported state.
- Functional incompatibility: It is worth noting that scalingAdapter has never worked correctly with RoleBasedGroupSet (RBGS), even before this PR. Enabling it would cause the RBGS controller to fight with the adapter over roles.replicas. With the new `recreate` strategy introduced in this PR, enabling this adapter is even more dangerous as it could trigger unintended RBG recreations.

internal/controller/workloads/rolebasedgroupset_controller.go
NoobDream2568: I’ve adjusted the semantics of `replicas` and several other status fields to properly align with the changes introduced in this PR.

Specifically, both `status.replicas` and `status.readyReplicas` now include surge instances. Therefore, during a surge window, `status.replicas` will correctly reflect the total number of child RoleBasedGroups (base ordinals + surge), rather than just the base ordinals. This ensures the meaning remains consistent with the scale subresource expectations.

api/workloads/v1alpha2/rolebasedgroupset_types.go
NoobDream2568: This is intentionally kept as reserved for future implementation.

internal/controller/workloads/rolebasedgroupset_controller.go
NoobDream2568: Fixed. A corresponding e2e test case has been added to cover this scenario.

PR #476, opened 2026-09-20
test(upgrade): add the v0.8.0 -> current hop to the upgrade e2e suite

diw-zw:
### Ⅰ. Motivation

The upgrade-compatibility e2e suite (`test/e2e/upgrade`) only covered the **v0.7.0 → current** hop. With v0.8.0 published, the **v0.8.0 → current** hop needs equal coverage: it is the upgrade path users on the latest release will actually take, and it carries its own risks (e.g. the new update-strategy type enum on the RIS CRD, the CoordinatedPolicy `gang.minReplicas`/`OrderScheduled` defaults) that the v0.7.0 leg cannot see.

The suite's invariant — *upgrading must not disturb already-running workloads* — is hop-independent, so rather than duplicating the suite, this PR parameterizes the four things that genuinely differ per from-release and runs the same experiment from both starting points.

### Ⅱ. Modifications

- **`test/e2e/upgrade/profile.go` (new)**: a `releaseProfile` resolved from `RBGS_FROM_GIT_TAG`, carrying everything that differs per hop:
  - `installValues` — v0.7.0 uses the old top-level values paths (`image.*`, `crdUpgrade.tag`, `portAllocator.enabled`); v0.8.0 already uses the current `controller.*` layout. (The chart silently ignores keys from the wrong layout, so this must be exact.)
  - `newBundleMarker` — evidence the upgrade's CRD bundle landed: the warmup CRD for v0.7.0; the `updateStrategy.type` enum on the RoleInstanceSet CRD for v0.8.0 (the warmup CRD already exists in v0.8.0, so it cannot mark that hop).
  - `prunesRestartPolicyConfig` — gates the phase-1 prune probe: v0.7.0's CRDs predate the field, v0.8.0's already carry it (probe skips with a named reason).
  - `rewrites` — the recorded, bidirectionally-asserted changes per hop. v0.8.0 records only the legacy strategy-type heal; the KEP-260 `sharedServiceSelection` selector narrowing is v0.7.0-only, since v0.8.0 already resolves unset to `LeaderOnly`.
- **Suite parameterization** (`preflight.go`, `install.go`, `upgrade.go`, `specs.go`, `snapshot.go`, `suite_test.go`): install/verify/marker checks and spec titles are driven by the profile; `upgradeRewrites` moved into the profiles.
- **`fromTag()` derivation**: the from-release is now selected with `RBGS_FROM_GIT_TAG` alone — the suite derives the matching published image tag (`v0.8.0 → v0.8.0-16705159`). `RBGS_FROM_TAG` remains as an override (CI passes it explicitly). Local usage: `RBGS_FROM_GIT_TAG=v0.8.0 make test-e2e-upgrade`.
- **Makefile**: single `test-e2e-upgrade` target (the separate v080 target is gone; it only injected the two env vars).
- **`.github/workflows/release-test.yml`**: `e2e-test-upgrade` becomes a two-leg matrix (`v0.7.0`, `v0.8.0`) with `fail-fast: false`, per-leg job names and per-leg log artifacts. The single `Release Test / e2e-test-upgrade` commit-status context now aggregates both legs via `needs`. A future hop is a one-line matrix entry.

NONE

This PR *is* the test addition: the upgrade e2e suite now runs twice per release gate, once per supported from-release. The v0.8.0 leg runs 19 of 20 specs (the prune probe skips by design, with the reason logged); the v0.7.0 leg runs all 20 unchanged.

- `RBGS_FROM_GIT_TAG=v0.8.0 make test-e2e-upgrade` against a clean kind cluster with the four images pre-loaded (suite prints the exact `docker pull` list on precondition failure).
- Verified locally on kind (K8s v1.31.12, same dependency versions as CI): **v0.8.0 → current 19 Passed / 0 Failed / 1 Skipped**; **v0.7.0 → current regression 20 Passed / 0 Failed**. Also verified on a real 3-node ACK cluster (v1.36.2): same 19/0/1 result, with teardown leaving the cluster clean.

- The v0.8.0 leg records *no* new rewrites beyond the shared strategy heal. This was verified by diffing v0.8.0 → HEAD: the gang-annotation injection is a no-op without a CoordinatedPolicy (fixtures have none, as documented), the conversion diff is only the strategy-type mapping (the v1alpha1 fixtures' `RollingUpdate.Type` is empty), and `sharedServiceSelection` already resolves to `LeaderOnly` in v0.8.0.
- Both legs intentionally run the *same* specs: the suite asserts a hop-independent invariant on an identical fixture world, and the per-hop differences are asserted bidirectionally through the profile's `recordedRewrites` (a recorded change must happen, and nothing unrecorded may).


cheyang (approved):
/approve
/lgtm

PR #478, opened 2026-09-20
fix: make warmup target waits recoverable and drop the CRD rule that …

diw-zw:
### Ⅰ. Motivation

Follow-up to #466 (`fix: validate warmup actions and target RBG failures`), addressing the two problems left open when it merged.

1. **A missing target `RoleBasedGroup` is an unrecoverable one-shot failure.** #466 made a missing target terminal (`Failed/InvalidTarget`). But the target is often created moments *after* the Warmup — a single `kubectl apply -f` of a combined manifest, or any GitOps ordering — so the Warmup fails permanently even once the target exists. #466's `apiReader` fallback only covers informer lag, not "created a moment later".

2. **An existing target RBG whose Pods are not scheduled yet reports `Completed/NoNodesMatched`.** `getDesiredNodesToWarmup` skips Pods with `spec.nodeName == ""`, so a Warmup that reconciles before its target's Pods are bound sees zero nodes and completes successfully. Nothing is pre-warmed, yet `Completed` is exactly the signal the KEP workflow tells users to wait for before triggering the RBG upgrade — a silent false success that a one-shot resource can never correct.

3. **#466's customized-container-image CEL rule freezes pre-existing objects.** The rule is re-evaluated on *every* write against the whole object. On Kubernetes <= 1.32 the status subresource strategy runs the CEL validator **without ratcheting** — `CRDValidationRatcheting` was only wired into the status path in 1.33 — so an already-stored object with a missing/blank image can no longer have its `status` written at all. The controller cannot record `phase`/`conditions`, and `kubectl wait` never returns. This repo's CI targets k8s 1.31 (`ENVTEST_K8S_VERSION: 1.31.0`, `KUBERNETES_VERSION: v1.31.12`).

### Ⅱ. Modifications

**1. Recoverable target waits (`internal/controller/workloads/rolebasedgroupwarmup_controller.go`)**

- A missing target RBG enters a recoverable waiting state instead of a terminal failure: `phase: Running`, `TargetReady=False` with reason `RoleBasedGroupNotFound`, no Pods created, and a 10s requeue. The Warmup proceeds as soon as the target appears.
- An existing target RBG with unscheduled Pods waits with `TargetReady=False` / `TargetPodsNotScheduled` rather than reporting `Completed/NoNodesMatched`. `NoNodesMatched` is now reserved for configurations with genuinely nothing to do (an empty node selector, or a target RBG whose listed roles have no Pods at all).
- `TargetReady=True` (`RoleBasedGroupReady`) is recorded once the target is usable.
- Waiting is bounded by `globalTimeoutSeconds` when set, measured from the Warmup's **creation time** while no Pod exists yet (`status.startTime` is only set once a Pod is created, so it cannot bound the wait itself). On expiry the Warmup fails terminally with `InvalidTarget` (target still absent) or `GlobalTimeoutExceeded` (Pods still unscheduled). Without `globalTimeoutSeconds`, waiting is unbounded.

**2. Remove the customized-container-image CEL rule; validate in the controller only**

- The `+kubebuilder:validation:items:XValidation` marker added by #466 is removed, and the generated CRD/manifests are regenerated. The CRD diff is a pure deletion of the rule.
- An action whose container `image` is missing or blank now fails the Warmup with reason `InvalidWarmupSpec` and creates no Pods — the behavior #466 already added, unchanged.
- This matches every other pod template in the project: the RoleBasedGroup, RoleBasedGroupSet, Instance and InstanceSet CRDs do not validate container images either. A customized action container is a bare `corev1.Container` embedded in the Warmup spec, so the Pod API's own image validation never runs on it; the controller is the enforcement point.
- The `minLength` marker on `imagePreload.images` from #466 **stays**: it is ratcheting-safe (a pre-existing object with `images: [""]` keeps a writable status), which I verified separately.

**3. Docs and a regression guard**

- KEP: documents the `TargetReady` condition, the bounded-wait contract, the narrowed `NoNodesMatched` meaning, and why customized container images are validated by the controller rather than the CRD.
- `TestGeneratedWarmupCRDDoesNotValidateContainerImages` pins the invariant so the CEL rule is not reintroduced, with the reasoning in a comment. If admission-level validation is ever wanted here, it should be a validating webhook rather than a CRD rule.

NONE

Follow-up to #466, which is already merged.

Unit tests (`internal/controller/workloads/rolebasedgroupwarmup_controller_test.go`):

| Test | What it covers |
|---|---|
| `TestReconcile_MissingTargetRBGRecoversWhenCreatedLater` | Missing target waits (`Running` + `TargetReady=False/RoleBasedGroupNotFound` + requeue), then proceeds and creates the Pod once the target appears. |
| `TestReconcile_MissingTargetRBGFailsAfterGlobalTimeout` | A lingering missing target still fails terminally (`Failed/InvalidTarget`) once `globalTimeoutSeconds` lapses. |
| `TestReconcile_UnscheduledTargetPodsWaitInsteadOfCompleting` | Unscheduled target Pods yield `TargetReady=False/TargetPodsNotScheduled` instead of `Completed`; warmup proceeds after the Pod is scheduled. |
| `TestReconcile_TargetRBGWithNoPodsCompletesNoNodesMatched` | A target RBG with no Pods at all still completes with `Complete/NoNodesMatched`. |
| `TestReconcile_InvalidCustomizedContainerGapIsClosed` | An empty/whitespace `customizedAction` image fails with `InvalidWarmupSpec`, creates no Pods, and never reaches Pod creation. |
| `TestReconcile_MissingTargetRBGDoesNotFailImmediately` | Replaces `TestReconcile_MissingTargetRBGFailsWithoutRequeue`, whose assertion encoded the old terminal-failure semantics. |

CRD guard (`api/workloads/v1alpha2/rolebasedgroupwarmup_crd_test.go`):

| Test | What it covers |
|---|---|
| `TestGeneratedWarmupCRDDoesNotValidateContainerImages` | The generated CRD carries neither the #466 CEL rule nor its message, while the ratcheting-safe `images` `minLength` marker is retained. |

Existing Warmup controller, Pod construction, timeout, backoff and status tests pass unchanged; the envtest suite (`rbg`, `restart_policy`, `webhook`) passes.

```bash
# Generated artifacts must be stable, code formatted, lint clean
make manifests generate fmt
git status --porcelain          # expect: empty
./bin/golangci-lint run         # expect: 0 issues

# Unit + envtest
go test ./internal/... ./api/... ./pkg/... ./hack/... -count=1
KUBEBUILDER_ASSETS="$(setup-envtest use 1.31.0 -p path)" go test ./test/envtest/... -count=1 -timeout 25m
```

Verified end-to-end on a **kind v1.31.12** cluster with this build deployed:

| Scenario | Result |
|---|---|
| Warmup created before its target RBG | `TargetReady=False/RoleBasedGroupNotFound`; on target creation -> `TargetReady=True`, warmup Pod created |
| Target RBG with an unschedulable Pod | `TargetReady=False/TargetPodsNotScheduled` (previously `Completed/NoNodesMatched`) |
| Create with a missing / blank container image | accepted by admission, then `Failed/InvalidWarmupSpec`, **0 Pods created** |
| Pre-existing object with a missing image, after CRD upgrade | controller writes status normally; `Failed/InvalidWarmupSpec` recorded, manual status patch accepted |

The A/B that motivates fix 2, on the same pre-existing object (k8s 1.31.12):

| CRD | `PATCH .../status` |
|---|---|
| #466 as merged (item-level CEL) | rejected: `invalid: customized action container image must not be empty` |
| This PR (rule removed) | accepted |


internal/controller/workloads/rolebasedgroupwarmup_controller.go
```diff
@@ -184,13 +192,50 @@ func (r *RoleBasedGroupWarmupReconciler) reconcileUnfinished(ctx context.Context
...
+			}
+			return r.markTargetNotReady(ctx, warmup, "RoleBasedGroupNotFound", err.Error())
 		}
 		logger.Error(err, "failed to get desired nodes to warm up")
 		return ctrl.Result{}, err
 	}
 
+	// The target RoleBasedGroup exists. If it yielded no nodes because some of its Pods
+	// are not scheduled yet, do not report success: wait until they are scheduled.
+	if warmup.Spec.TargetRoleBasedGroup != nil && len(desiredNodes) == 0 {
```
cheyang: This check only runs while `desiredNodes` is empty. If one selected `worker` Pod is bound and another is still pending, `desiredNodes` is non-empty, so the controller sets `TargetReady=True` and creates a warmup Pod for only the bound node. The job can complete before the second node is discovered. Could we check every Pod in the selected roles before starting any warmup Pods?

internal/controller/workloads/rolebasedgroupwarmup_controller.go
```diff
@@ -413,6 +467,108 @@ func (r *RoleBasedGroupWarmupReconciler) requeueForTimeout(warmup *workloadsv1al
...
+	pods, err := r.listTargetRBGPods(ctx, warmup)
+	if err != nil {
+		return 0, err
+	}
+	pending := 0
+	for _, pod := range pods {
+		if pod.DeletionTimestamp != nil {
+			continue
+		}
+		if pod.Spec.NodeName == "" {
```
cheyang: This counts every unscheduled Pod in the RBG, including roles absent from `targetRoleBasedGroup.roles`. In the reproduced case, a pending `router` Pod keeps a Warmup that selects only a zero-replica `worker` in `TargetPodsNotScheduled` instead of `Completed/NoNodesMatched`. Please filter by the selected role before incrementing `pending`.

internal/controller/workloads/rolebasedgroupwarmup_controller_test.go
```diff
@@ -1174,3 +1178,269 @@ func TestGetDesiredNodes_UsesAPIReaderAfterCachedRBGMiss(t *testing.T) {
...
+		t.Fatalf("expected Failed after global timeout, got %q", updated.Status.Phase)
+	}
+	if cond := apimeta.FindStatusCondition(updated.Status.Conditions, "Failed"); cond == nil || cond.Reason != "InvalidTarget" {
+		t.Fatalf("expected Failed/InvalidTarget, got %#v", updated.Status.Conditions)
+	}
+}
+
+// An existing target RBG whose Pods are not scheduled yet must not be reported as a
+// successful (but empty) warmup.
+func TestReconcile_UnscheduledTargetPodsWaitInsteadOfCompleting(t *testing.T) {
```
cheyang, line 1417: This test covers only the all-unscheduled case, so `desiredNodes` stays empty and the new guard runs. It misses the boundary where one selected Pod is bound and another is pending, and it does not cover a pending Pod in an unselected role. Could we add those two cases? They reproduce the partial-warmup and unrelated-role blocking paths.

cheyang (changes requested):
I found two readiness gaps in the new state machine. A mix of scheduled and pending Pods in a selected role can start a partial warmup, and a pending Pod from an unselected role can block the job. Both reproduce in focused contract tests and need coverage before merge.

PR #479, opened 2026-09-21
fix: isolate unhealthy timers by RoleInstanceSet

cyf-2002:
### Ⅰ. Motivation

Rolling updates can remain stuck on the old template when two stateful RoleInstanceSets repeatedly reconcile in turn.

For example, `prefill` and `decode` each have one old, NotReady RoleInstance and a new template to roll out. With `podManagementPolicy: Parallel`, `maxUnavailable: 1`, and `maxSurge: 0`, the old instance already consumes the unavailable budget. The controller can update it after ten seconds without an intervening Ready observation, because doing so does not make an additional instance unavailable.

That wait can keep restarting because health timestamps are shared across sets, but cleanup only considers the instances of the set currently being reconciled:

1. Reconciling `prefill` records when its instance was first observed NotReady.
2. Reconciling `decode` deletes that timestamp because the prefill instance is absent from decode's instance list, even though it still exists.
3. Reconciling `prefill` again starts a new ten-second wait and deletes decode's timestamp in the same way.

If this sequence repeats, both rollouts can stay blocked and their RoleInstances continue using the old template. The timed retries in #470 bring the controller back, but cannot help if another set keeps deleting the start time.

### Ⅱ. Modifications

Preserve each set's own unhealthy start times so that reconciling `decode` no longer resets the wait for `prefill`, or vice versa. Once an instance's wait expires, the existing retry can resume its update, subject to the other rollout checks.

Store timestamps using the set's namespace/name, set UID, and RoleInstance UID. Only remove the current set's recovered or missing instances during health observation, and read the same key when checking the ten-second wait or scheduling a retry. Reclaim records for deleted or recreated sets using the existing informer cache during later stateful reconciles; retain records if the cache lookup fails for a reason other than NotFound.

NONE. Follow-up to #470 for cross-set timer isolation.

Extend the existing `stateful_instance_set_control_test.go`; no new test files are added.

- Extend the existing retry test with another set's health observations; verify that retry deadlines are preserved and updates resume after the unhealthy window expires.
- Healthy, unhealthy, and empty sets cannot reset another set's timer, including across namespaces.
- Cover set and instance recreation, plus cleanup of live, deleted, and recreated sets and cache lookup errors.

```sh
go test -mod=vendor -race ./pkg/reconciler/roleinstanceset/statefulmode -count=1
go vet -mod=vendor ./pkg/reconciler/roleinstanceset/statefulmode
```

Both commands passed locally. Removing the set-scope guard makes the regressions fail: the remaining four-second retry resets to ten seconds, and another namespace loses its timer. They pass with the guard in place.

Preserve the existing ten-second health window and availability budget, along with the timed retries added in #470. Cleanup uses the local informer cache without extra API requests. Cluster E2E tests were not run.

## Checklist

- [x] Format changed Go files with `gofmt`.
- [x] Add regression tests to the existing test file.
- [x] Documentation: not required for this internal controller fix.


pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_controller.go
```diff
@@ -170,6 +170,8 @@ func (ssc *ReconcileStatefulInstanceSet) Reconcile(ctx context.Context, request
 		}
 	}()
 
+	ssc.pruneInstanceHealth()
```
cheyang, line 173: Minor (efficiency): `pruneInstanceHealth()` is called at the top of every `Reconcile` and `Range`s the entire global `instanceUnhealthySince` map, doing a lister `Get` per entry — O(N) per reconcile, including reconciles where the map only holds the current set's own entries.

The lister `Get` is an O(1) cache lookup, so this is cheap in practice and I would not block on it. But it is unconditional full-scan overhead on the hot reconcile path. A lazier cadence — prune periodically, or only when the map grows past a threshold — would carry no correctness cost and avoid the per-reconcile scan. Worth a follow-up rather than a gate.

pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_controller.go
```diff
@@ -201,6 +203,21 @@ func (ssc *ReconcileStatefulInstanceSet) Reconcile(ctx context.Context, request
...
 
+// pruneInstanceHealth removes observations for deleted or replaced sets using
+// the informer cache. The outer controller returns early for deleted sets, so
+// reclaim their records when another set reconciles instead of relying on a
+// NotFound request reaching this reconciler.
+func (ssc *ReconcileStatefulInstanceSet) pruneInstanceHealth() {
+	instanceUnhealthySince.Range(func(key, _ interface{}) bool {
+		healthKey := key.(instanceHealthKey)
+		set, err := ssc.roleInstanceSetLister.RoleInstanceSets(healthKey.instanceSet.Namespace).Get(healthKey.instanceSet.Name)
+		if errors.IsNotFound(err) || (err == nil && set.UID != healthKey.instanceSetUID) {
```
cheyang: Minor (correctness/edge): in `pruneInstanceHealth`, informer-cache `IsNotFound` is treated as definitive and the entry is deleted — including entries for *other* sets:

```go
set, err := ssc.roleInstanceSetLister.RoleInstanceSets(...).Get(...)
if errors.IsNotFound(err) || (err == nil && set.UID != healthKey.instanceSetUID) {
    instanceUnhealthySince.Delete(key)
}
```

During a brief cache desync — controller restart before the informer has fully synced, or a watch reconnect — an existing set can be momentarily absent from the cache and read as `IsNotFound`. A reconcile of set B would then wipe set A's timer. That reintroduces the *same cross-set reset class* this PR fixes, just via cache inconsistency rather than the `observeInstanceHealth` cleanup loop.

The PR comment says "retain records if the cache lookup fails for a reason other than NotFound," but a transient cache inconsistency that *looks like* NotFound isn't covered. Low probability in steady state, but it is the same failure class the PR is closing. One option: gate the delete on the informer having synced (`HasSynced`), or skip pruning during the initial-sync window.

cheyang (commented):
## Verdict: COMMENT

Verified this PR with a layered harness on top of the provided ACK cluster (full evidence: `docs/verification/pr479-isolate-unhealthy-timers/` on the `verify/pr479-isolate-unhealthy-timers` branch). The premise **reproduces** on the base (`1a9653ad`, #470) and the fix **is confirmed** on the head (`5a322f23`) — both at the unit level and live A/B against the real cluster. No blockers or majors; one minor, non-blocking observation below.

### Premise (P0) — confirmed

`instanceUnhealthySince` is a global `sync.Map` keyed only by RoleInstance UID. `observeInstanceHealth` runs at the top of every reconcile with just the *current* set's instances, and its cleanup deletes any UID absent from that slice. So reconciling set B wipes set A's unhealthy start time (and vice versa). Under interleaving, neither set's 10 s `stableUnhealthyDuration` window ever matures, `isStablyUnhealthy` stays false, the unavailable budget stays exhausted, and both rollouts remain stuck on the old template. #470's timed retries keep re-enqueuing the controller but cannot help while the other set's reconcile keeps deleting the start time.

Reproduced deterministically on base: `TestPremiseCrossSetTimerWipe`.

### Fix — confirmed

Keying by `instanceHealthKey{instanceSet, instanceSetUID, instanceUID}`, scoping `observeInstanceHealth` cleanup to the current set, and adding `pruneInstanceHealth()` (informer-cache-backed reclamation for deleted/replaced sets) eliminates the cross-set wipe. Each set's timer is untouched by the other's reconcile, matures after ~10 s, and the rollout proceeds.

- Head unit harness: `TestPremiseCrossSetTimerIsolated_HEAD` passes; existing tests green under `-race`; `go vet` clean.
- Harness-bites: removing the set-scope guard makes the cross-set tests fail (and only those) — the tests genuinely cover the fix.

### Live A/B on the provided ACK cluster

Dual standalone `RoleInstanceSet` roles (`prefill` + `decode`), 1 replica each, Parallel, `maxUnavailable=1 maxSurge=0`, failing readiness probe so each instance consumes the unavailable budget. Trigger a rollout (label `rollout-trigger` v2→v3), restart the controller for fresh in-process timers, then poke both RoleInstanceSets' annotations every 3 s to simulate the interleaved reconcile from the premise.

| Controller | "budget exhausted" events | Pod template after ~33 s | Result |
|---|---|---|---|
| base (#470) | **41** | stuck at v2 | premise reproduced live |
| head (#479) | **2**, then `targets=[]` | progressed to v3 | fix confirmed live |

### One minor, non-blocking observation

It is about `pruneInstanceHealth`, not the core fix:

It `Range`s the whole global map and does a lister `Get` per entry on **every** `Reconcile` — O(N) full-scan overhead even when only this set's entries exist. Cheap in practice (lister `Get` is an O(1) cache lookup), but a lazier cadence would avoid it.

It does not block merge. Net: the PR correctly isolates per-RoleInstanceSet unhealthy timers; the note is a minor hardening suggestion.

---

Cluster state restored after the run (webhook back to `Fail`, deployed controller back to 2/2 Ready, `pr479-verify` namespace deleted).

pkg/reconciler/roleinstanceset/statefulmode/stateful_instance_set_controller.go
diw-zw: Thanks @cheyang. Tracked as #481. Let's merge #479 first and address the prune frequency as a follow-up optimization PR.

diw-zw (approved):

